PR #105, opened 2020-10-19
Initial setup of cucumber (#43)

kryswisnaskas:
**Description of change**
This PR covers the initial setup of cucumberJS. Before getting to deep into writing a lot of tests, we want to make sure the approach here is approved.

**How to test**
More information on how to test can be found here
https://github.com/adhocteam/Head-Start-TTADP/pull/43

**Issue(s)**
* https://github.com/HHS/Head-Start-TTADP/issues/43

**Checklist**

- [x] Meets issue criteria
- [x] Code tested
- [ ] Documentation updated
    - API methods

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,7 +147,6 @@ jobs:
       - run:  # Fail audit only for moderate and higher vulnerabilities
           name: Audit vulnerability of backend node_modules
           command: |
-            cd frontend
             chmod 744 ./run-yarn-audit.sh
             ./run-yarn-audit.sh; \
             EXIT_CODE=$?; \
@@ -155,7 +154,10 @@ jobs:
       - run:  # Fail audit only for moderate and higher vulnerabilities
           name: Audit vulnerability of frontend node_modules
           command: |
-            yarn --cwd frontend audit --level low; EXIT_CODE=$?; \
+            cd frontend
+            chmod 744 ./run-yarn-audit.sh
+            ./run-yarn-audit.sh; \
+            EXIT_CODE=$?; \
               (($EXIT_CODE >= 4)) && exit $EXIT_CODE || exit 0
       - store_artifacts:
           path: reports
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,6 +173,37 @@ jobs:
           path: reports/
       - store_artifacts:
           path: coverage/
+  cucumber_test:
+    executor: docker-executor
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: Start server
+          command: |
+            yarn deps
+            yarn build
+            yarn start:ci
+          background: true
+      - run:
+          name: Run migrations ci
+          command: yarn run db:migrate:ci
+      - run:
+          name: Install Headless Chrome dependencies
+          command: |
+            sudo apt-get install -yq \
+            gconf-service libasound2 libatk1.0-0 libatk-bridge2.0-0 libc6 libcairo2 libcups2 libdbus-1-3 \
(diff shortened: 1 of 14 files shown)


cucumber/features/devLoginPup.feature
```diff
@@ -0,0 +1,10 @@
+Feature: TTA Smarthub HSES login
+ 
+    Scenario: Welcome page is displayed
+        # Given https://tta-smarthub-dev.app.cloud.gov
+        Given the home page of tta-smarthub
+        Then we should see "Welcome to the TTA Smart Hub!" message
+    Scenario: Login is redirected to HSES
+        Given the home page of tta-smarthub
+        When pressing login
+        Then we should see "Head Start Enterprise System" page
```
rahearn, line 10: This is good, but ideally I'd like the scenarios to be a bit more like a first-person narrative. Something like:

```suggestion
        Given I'm on the Smart Hub home page
        When I press login
        Then I should see the HSES login page
```

cucumber/features/steps/devLoginPupSteps.js
```diff
@@ -0,0 +1,35 @@
...
+  await scope.context.currentPage.click('a[href$="api/login"]');
+  await scope.context.currentPage.waitForSelector('.hses');
+});
+
+Then('we should see "Head Start Enterprise System" page', async () => {
+  const result = await scope.context.currentPage.$('.hses');
+  const value = await scope.context.currentPage.evaluate((el) => el.textContent, result);
+
+  assertTrue(value.includes('Head Start Enterprise System'));
+});
```
rahearn, line 35: One pattern I've seen work well, is to use `""` in step definitions to mean specific things to look for, so this could be generalized to:

```suggestion
Then('I should see {string} on the page', async (searchString) => {
  // set value to be all the text on the page
  const value = ...
  assertTrue(value.includes(searchString));
});
```

More info on that `{}` format: https://cucumber.io/docs/cucumber/cucumber-expressions/

Similarly, you could create a generic "press button" step and use it for the login step. `When('I press {string}', async (buttonLabel) => {`

package.json
```diff
@@ -107,9 +109,10 @@
     "supertest": "^4.0.2"
   },
   "dependencies": {
-    "@babel/runtime": "^7.11.2",
```
rahearn, line 110: Losing babel/runtime is surprising to me, was that intended? It looks likely to be the cause of the deploy step not working.

rahearn (commented):
Great start to this, and I really like the ability to view the test results in the artifacts listing.

cucumber/features/steps/devLoginPupSteps.js
kryswisnaskas: Yes, most of the future tests include this pattern. Thanks!

package.json
kryswisnaskas: Will fix.

.circleci/config.yml
```diff
@@ -147,7 +147,10 @@ jobs:
       - run:  # Fail audit only for moderate and higher vulnerabilities
           name: Audit vulnerability of backend node_modules
           command: |
-            yarn audit --level low; EXIT_CODE=$?; \
+            cd frontend
```
SarahJaine, line 150: This is supposed to be the audit for the BE, you're now auditing the FE twice. Wouldn't we want to run that script both for the FE and the BE deps audit?
```suggestion
```

rahearn (changes requested):
This resulted in running `yarn audit` twice against the frontend and not at all against the backend.

SarahJaine (dismissed):
Sorry, I missed my window on the AH review. I think we need to change the yarn audit in two place.

.circleci/config.yml
SarahJaine: Aaaand, apparently @rahearn and I were writing the same thing at the same time. Whoops.

rahearn (approved):

PR #107, opened 2020-10-21
Update react scripts

jasalisbury:
New release of `react-scripts` updates a dependency that previously had a vulnerability. From `create-react-app` release notes:

> v3.4.4 release bumps resolve-url-loader to a version for which npm audit does not report a vulnerability. Note that this vulnerability did not affect Create React App projects, so this change is only necessary to satisfy auditing tools.


jasalisbury:
Made against wrong repo

Issue #79, opened 2020-10-07
As a user who has left my computer, I should be logged out after a period of inactivity, so my account remains secure (AC-2(5), AC-12)


rahearn:
**User story:**
As a user who has left my computer for 25 minutes or longer, I want the system to log me out of Smart Hub so the site remains secure.

**Acceptance Criteria:**
1. After 25 minutes of inactivity, the user is given a warning and option to continue session.
2. If the user clicks anywhere on the screen, the message is dismissed and the session is restarted.
3. If the user doesn't click anywhere on the screen, then at the 30 minute mark, the system logs the user out of Smart Hub and redirects them to the Welcome page.
4. The length of the inactivity timeout is easily configurable (a config file is fine).

**Considerations:**
-  HSES's timeout will not affect our users. 

**Exclusions:**
- design by Aricka - she'll review what the developers build and if we need changes, they'll go into another story.
- autosaving of work before timeout - #98 
- locking of screen (rather than logout) - not needed

rahearn:
Seems like there's supposed to be a *15* minute session lock, and a *30* minute session termination timeout. Asking for clarity from IPT on the practical differences between lock and termination for a web app.

rahearn:
We can ignore the 15 minute session lock and only implement the 30 minute session termination

pamlo412:
Added screen lock to list of exclusions. Assigned 8 points based on discussion with @kryswisnaskas and @jasalisbury

rahearn:
Definition of done:

- [x] The business and/or user value of a story has been met
- [x] CI pipeline green
- [x] Code is meaningfully tested
- [x] No linting errors or warnings
- [x] UI meets WCAG 2.1 - Levels A and AA
- [x] Code has deployed to a staging environment
- [x] Security scans have passed
- [x] [Architectural Decision Records](https://adr.github.io/) are written for major infrastructure decisions with [the Nygard template](https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_by_michael_nygard.md).
- [x] Boundary and Data Flow Diagrams are updated when a change invalidates the current version.
- [x] Logical Data Model Diagram is updated on DB schema changes.
- [x] Public API methods are documented when added or changed.

pamlo412:
Screen shot of timeout warning:
![image](https://user-images.githubusercontent.com/59671463/101673016-98b27d00-3a24-11eb-8847-cd44bd65009b.png)

PR #133, opened 2020-11-05
Logout when inactive

jasalisbury:
**Description of change**
The idle modal shows after 25 minutes of inactivity. After 30 minutes of inactivity the user is logged out of the TTA smart hub. If the user becomes active while the modal is displayed it is hidden and the counters are reset.

Currently when the modal is shown the message says how long the user has until logout when the modal was first shown. I see two ways to improve the user experience:

 1. We have a running countdown that updates until the user is logged out (e.g. T-minus 1 minute 2 seconds until logout)
 2. We display the actual time at which the user is logged out (e.g. 10:55 am)

**How to test**

I've set sandbox to show the logout modal at 1 minute of inactivity and to logout users after an additional minute (2 minutes total). I'll need to remember to update the environment variables to 25 minutes (modal) and 30 minutes (logout) before merging.

1. Browse to sandbox https://tta-smarthub-sandbox.app.cloud.gov/
2. Login
3. Become inactive for a minute, note the modal
4. Wait an additional minute to get logged out

**Notes**

* We cannot set frontend environment variables per environment since they have to be set at build time. We will need to figure out how to do so soon. I can see us wanting to set feature flags for one environment but not another.
* I added a timeout to the session cookie sent from the server. The timeout is currently set at 30 minutes. This timeout is updated every time the user hits the server. There is a potential issue if a user is active on a page but not making API calls for 30 minutes. In that case the user's session will expire but they won't get explicitly logged out on the frontend. If this happens the next call to the API will give a 401, which we will need to handle. I gave some thought to having a 'heartbeat' API endpoint that gets hit periodically, but things get complicated when combined with the inactivity timeout. Let me know if you have any ideas!

**Issue(s)**
* closes https://github.com/HHS/Head-Start-TTADP/issues/79

**Checklist**

- [x] Meets issue criteria
- [x] Code tested
- [n/a] ~Documentation updated~
    - API methods


src/app.js
```diff
@@ -28,7 +28,9 @@ app.use(session({
   cookie: {
     httpOnly: true,
     sameSite: 'strict',
+    maxAge: Number(process.env.SESSION_TIMEOUT),
   },
+  rolling: true,
```
rahearn, line 33: What does `rolling` do?

frontend/src/components/IdleModal.js
```diff
@@ -0,0 +1,104 @@
...
+          Stay logged in
+        </Button>
+      )}
+    >
+      <Alert role="alert" type="warning">
+        You will be automatically logged out due to inactivity in
+        {' '}
+        { timeToLogoutMsg }
+        {' '}
+        unless you become active again.
```
rahearn, line 58: I think in voice-over situations this might be a bit confusing. There is the button, but you can't keyboard nav to it because hitting tab dismisses the whole modal. Maybe add `Press any key to continue your session` in a screen-reader only span?

frontend/src/components/IdleModal.js
```diff
@@ -0,0 +1,104 @@
...
+  const onIdle = () => {
+    const timer = setTimeout(() => {
+      closeModal();
+      logoutUser(true);
+    }, modalVisibleTime);
+    openModal();
+    updateInactiveTimeout(timer);
+  };
+
+  const onActive = () => {
```
rahearn, line 80: Does `onActive` fire only when the user is active after the modal is presented, or on any activity?

rahearn (changes requested):
Great work! I'd like to see the accessibility change before merging this in.

Also, can we set the timeout via a CircleCI variable? I don't want a 2 minute timeout in staging/prod, but maybe it's a decent workaround to have one set of flags for staging/production and one for dev/sandbox, at least until something more involved is needed and can be developed.

src/app.js
jasalisbury: Every call to the API resets the timeout on the cookie. https://www.npmjs.com/package/express-session#rolling

frontend/src/components/IdleModal.js
jasalisbury: `onActive` fires when a user moves from being idle to active, so only after the modal is presented. There is a callback we could use if we wanted to perform an action on any activity, `onAction`.

rahearn (approved):

Issue #132, opened 2020-11-05
Handle cookie timeouts in the frontend


jasalisbury:
**User Story**

As a TTA Smart Hub user, I want my session on the server to remain active as long as I am active on the page, so I don't have to login when sending/receiving data from the API

---

We added a idle timeout in the frontend in https://github.com/HHS/Head-Start-TTADP/issues/79. This logs out a user who has been inactive for a number of minutes (currently 30). However there is a potential issue if a user stays active on a page without hitting the API. In this case the user would never explicitly be logged out but their session cookie would timeout after 30 minutes of not making API calls. We need to ensure the user has an unexpired session as long as they are active on the frontend.

**Context**

Comments from this [PR](https://github.com/adhocteam/Head-Start-TTADP/pull/58)

**Acceptance Criteria**

 - [ ] A user that stays active on a single page does not have their session timeout

rahearn:
In addition to the above, We should rethink using the cookie's `maxAge`, as opposed to keeping it as a session cookie (and thus automatically logging out on browser quit). Maybe the `maxAge` is instead implemented as a data point in the session, and we automatically log people out if it's too old?

jasalisbury:
> In addition to the above, We should rethink using the cookie's `maxAge`, as opposed to keeping it as a session cookie (and thus automatically logging out on browser quit). Maybe the `maxAge` is instead implemented as a data point in the session, and we automatically log people out if it's too old?

Yeah coming back to this it seems like the best solution is to just drop the `maxAge` on the server side. I was worried about stale sessions in the session store but it seems most session stores have a mechanism for clearing out old session cookies.

rahearn:
Definition of Done:

- [ ] The business and/or user value of a story has been met
- [ ] CI pipeline green
- [ ] Code is meaningfully tested
- [ ] No linting errors or warnings
- [ ] UI meets WCAG 2.1 - Levels A and AA
- [ ] Code has deployed to a staging environment
- [ ] Security scans have passed
- [ ] [Architectural Decision Records](https://adr.github.io/) are written for major infrastructure decisions with [the Nygard template](https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_by_michael_nygard.md).
- [ ] Boundary and Data Flow Diagrams are updated when a change invalidates the current version.
- [ ] Logical Data Model Diagram is updated on DB schema changes.
- [ ] Public API methods are documented when added or changed.
[...]

PR #189, opened 2020-12-07
Remove server side session cookie timeout

jasalisbury:
**Description of change**

Having no cookie timeout defined causes the cookie to expire only when the user closes their browser, instead of their cookie expiring after a set amount of time. This prevents an edge case where the user is active on the site without hitting the server and having their session expire on the backend but not the frontend.

**How to test**

1) Pull down, and login.
2) Look at the cookie sent from the server, it should be a listed as a `session` cookie in the `expires at` field.

**Issue(s)**
* https://github.com/HHS/Head-Start-TTADP/issues/132

**Checklist**

- [x] Meets issue criteria
- [n/a] Code tested
- [n/a] Documentation updated
    - API methods

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,8 +13,6 @@ POSTGRES_HOST=localhost
 AUTH_CLIENT_ID=clientId
 AUTH_CLIENT_SECRET=clientSecret
 SESSION_SECRET=secret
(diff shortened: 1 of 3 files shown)


jasalisbury:
Wrong repo.

Issue #140, opened 2020-11-09
Activity Report navigation (stepper replacement)


pamlo412:
**User story:**
As a TTA specialist working on an activity report, I want to be able to enter sections of the report non-linearly so that I can complete the report one section at a time, as I have the information ready.

**Acceptance criteria:**
- [x] A section menu is added to the activity report
- [x] a status tag displays for each section (Not Started, In Progress, Complete)
- [x] The already-implemented fields are broken out into their appropriate sections/pages
- [x] the user can navigate to any section and complete fields in any order (aka the "anti-stepper")

**Design:**
https://preview.uxpin.com/ad995fd722bd26c0f9a1c8a115f1b41833637308#/pages/133952430/comment/no-panels

**Exclusions:**
- full population of each section will be handled via #99, #100, #101,  and #135. This story is just to get the menu in place and the already-implemented fields moved.
- handling of browser navigation, # TBD

rahearn:
Definition of Done

- [x] The business and/or user value of a story has been met
- [x] CI pipeline green
- [ ] Code is meaningfully tested
- [x] No linting errors or warnings
- [ ] UI meets WCAG 2.1 - Levels A and AA
- [x] Code has deployed to a staging environment
- [x] Security scans have passed
- [ ] [Architectural Decision Records](https://adr.github.io/) are written for major infrastructure decisions with [the Nygard template](https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_by_michael_nygard.md).
- [ ] Boundary and Data Flow Diagrams are updated when a change invalidates the current version.
- [ ] Logical Data Model Diagram is updated on DB schema changes.
- [ ] Public API methods are documented when added or changed.
- [x] OSCAL templates are completed when security controls are implemented or modified.

PR #190, opened 2020-12-07
Add navigator

jasalisbury:
**Description of change**

The navigator allows us to split up large forms into multiple pages. It has a side navigation that allows any page to be selected. The side navigation displays the state of each page and is stickied to the top of the page. Some notes:

 * Adds a new library `react-helmet` to help update page titles. This library helps keep the page title updated with the page the user is viewing in the activity report. Note this library will be replaced with `react-helmet-async`(https://github.com/staylor/react-helmet-async) that removes some react deprecation warnings but has the same usage.
 * Reduced the use of `act(...)` in tests. Previously we would get warnings that looked like `Warning: An update to *COMPONENT* inside a test was not wrapped in act(...).`. This indicates updates to the DOM were happening outside of a test. I've updated tests to not need to be wrapped in act. Some tests still use act but they deal with third party libraries or using mock timers.
 * Started replacing `fireEvent` with `userEvent` per this [suggestion](https://testing-library.com/docs/dom-testing-library/api-events/). Will make a separate PR to switch everything over
 * The activity summary page is pretty long, it would not surprise me if it gets broken down into multiple pages at some point.

**How to test**

1) Browse to https://tta-smarthub-sandbox.app.cloud.gov/activity-reports and fill out the form

OR

1) Pull down branch
2) Install deps
3) Start dev environment, login and browse to `localhost:3000/activity-report`
4) Fill out the form

**Issue(s)**
* closes https://github.com/HHS/Head-Start-TTADP/issues/140

**Checklist**

- [x] Meets issue criteria
- [x] Code tested
- [n/a] Documentation updated
    - API methods


frontend/src/pages/ActivityReport/index.js
```diff
@@ -1,80 +1,129 @@
...
-    handleSubmit,
-    getValues,
-  } = hookForm;
+  const resetForm = () => {
+    updateSubmitted(false);
+  };
 
   return (
     <>
+      <Helmet titleTemplate="TTA Smart Hub - Activity Report - %s" defaultTitle="TTA Smart Hub - Activity Report" />
```
rahearn, line 109: Would it be better to use a template of "%s - Activity Report - TTA Smart Hub"? At least in my browser, the actual page you're on ends up not fitting in the visible part of the tab

frontend/src/components/Navigator/__tests__/index.js
```diff
@@ -0,0 +1,64 @@
...
+  it('sets dirty forms as "in progress"', async () => {
+    renderNavigator();
+    const firstInput = screen.getByTestId('first');
+    userEvent.click(firstInput);
+    const second = await waitFor(() => screen.getByText('second page'));
+    userEvent.click(second);
+    const first = screen.getByText('first page');
+    await waitFor(() => expect(within(first.nextSibling).getByText('In progress')).toBeVisible());
+  });
+
```
rahearn, line 55: Should we have a test for `sets complete form as "complete"`?

rahearn (changes requested):
Great work, overall. I think there are a couple things that could be slightly improved, but the big change I'd like to see is that the "Not started" tags fail accessibility for contrast: https://app.circleci.com/pipelines/github/adhocteam/Head-Start-TTADP/684/workflows/7799b4b8-0574-4042-ad1b-460d705cdf10/jobs/3019

frontend/src/pages/ActivityReport/index.js
jasalisbury: Yeah seems like most of the time you want the title to be most specific to lest specific (github titles this page as `Add navigator by jasalisbury - Pull Request #190 - HHS/Head-Start-TTADP`). I'll update.

rahearn (approved):

Issue #208, opened 2020-12-11
As a system administrator, I want to handle multiple topics formats and regions


rahearn:
Builds on https://github.com/HHS/Head-Start-TTADP/pull/205. We need to support:

1) Topics column may be of form `Topic name | RoleAcronym[, RoleAcronym]` or just `Topic name`
2) Multiple regions need to be imported, each from their own smartsheet, so the script should handle being called multiple times (once for each region)

rahearn:
Region 9 has changed "Goal..." to be "Anticipated Outcome...". We need to support both of those column formats.

@PatricePascual-ACF we also need to check the rest of the regions for any deviations, and ideally ask the regions to not make any (more) changes.

PatricePascual-ACF:
@rahearn  I want to require regions to change any column names back before we export data.  
In other words, R9  will go back to Goal.  Confirming w/ Damon now.

PR #77, opened 2020-10-05
Add api documentation

jasalisbury:
**Description of change**

Add [Open api 3](https://swagger.io/specification/) api docs in `docs/openapi`. Api documentation uses [Redoc](https://github.com/Redocly/redoc) to serve API documentation locally

**How to test**

You'll need to update dependencies (`yarn deps` or `yarn docker:deps`). Run `yarn docs:serve` (or `yarn docker:start`) and browse to `localhost:5000` to view the API documentation.

**Checklist**

~- [n/a] Meets issue criteria~
- [x] Code tested
~- [n/a] Meets accessibility standards (WCAG 2.1 Levels A, AA)~
- [x] Documentation updated
    - API methods
    - Boundary and Data Flow Diagrams
    - [Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions with the [Nygard template](https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_by_michael_nygard.md)
    - OSCAL templates completed when security controls are implemented or modified

PR #239, opened 2020-12-23
Support TTA plan goals import from another region and dynamic ids

kryswisnaskas:
**Description of change**
Refactors change from #77 to support dynamic ids as well as import from another region. Included "Grantee Name" as the expected column name in addition to "Grantee (distinct by row number)". Added test for a sample data from region 9.

**How to test**
Running locally:

- yarn deps
- yarn db:migrate
- ./node_modules/.bin/babel-node ./tools/importTTAPlanGoals.js -f R9GranteeTTAPlanTest.csv
- verify that the data has been imported

To test using the original smartsheet file:

- Navigate to https://app.smartsheetgov.com/sheets/fjfHfQCg6mmvRP5ghV77wP9wQ5xg2P2H37QhF7Q1?view=grid
- Export the file to MS Excel or Google Sheets
- Replace smart quotes with non-smart quotes
- Replace "Anticipated Outcome" with "Goal" in the column names
- Save as comma delimited CSV, e.g. R9GranteeTTAPlan.csv
- Follow the above steps substituting the file name with R9GranteeTTAPlan.csv

**Issue(s)**
* closes https://github.com/HHS/Head-Start-TTADP/issues/208

**Checklist**

- [x] Meets issue criteria
- [x] Code tested
- [n/a] Documentation updated
    - API methods


src/migrations/20201120182130-create-role.js
```diff
@@ -3,14 +3,16 @@ module.exports = {
     await queryInterface.createTable('Roles', {
       id: {
         allowNull: false,
-        autoIncrement: false,
+        autoIncrement: true,
```
rahearn, line 6: Do we need to wipe dbs that have already had the previous create-role migration run to pick up this change?

rahearn (commented):
Just concerned about the migration question.

At this point in development, I think it's fine if we do have to blow away dbs and recreate, just want to make sure of the plan.

kryswisnaskas:
> At this point in development, I think it's fine if we do have to blow away dbs and recreate, just want to make sure of the plan.

Yes, that was the approach. To re-recreate the dbs from scratch until we need to preserve the schema and data.

src/migrations/20201120182130-create-role.js
kryswisnaskas: Yes. We need to undo all migrations and run them again.

rahearn: is `terraform destroy && terraform apply` the best way to accomplish that at this point? cc/ @SarahJaine

SarahJaine: What environments are we talking about doing this in? We just need to do this the **_first_** time this migration is run, correct?
--
To answer your question, to blow away and restore dev rds instance you'd need to do
1. from `terraform/dev` run `terraform -destroy -target=cloudfoundry_service_instance.database` (target the db, there's no need to destroy the s3 bucket too)
2. respond to confirmation prompt
3. `terraform apply`
4. respond to confirmation prompt
5. Then I think you might need to rerun circleci `build_release_deploy` workflow on the `main` branch in circle ci to deploy the app and reinstate the "binding" between the rds instance and the application. You can first check if it's still there by looking at the services in cloud.gov.

But I think you're right to blow away the whole rds instance. I tried to just run `db:drop` and `db:create` directly against a our sandbox db last week and I couldn't get `db:create` to run. Then again, I tried for like 5 minutes before moving on and just destroying/reapplying the whole rds instance.

rahearn: It'll need to happen in any environments that this migration had run against already before being edited, so that's `staging` and `dev` at least. It won't need to happen on `prod` because prod will only see this updated version the first time it's run. Once we have data that needs to be preserved the process will be to create new migrations for any changes, rather than editing old migraitons.

And yeah - a new deploy to redo the binding (and also to re-run the migrations) would be needed.

SarahJaine: Oh okay, we're changing a migration that's already run- sorry, I'm all caught up now. In sandbox and sandbox only, I added a `migrate:undo` step before migrations run (you'll see it in a forthcoming PR). I was pretty much banking on dev/staging/prod not having edited migrations, so I didn't add this to any other environment. Since we don't have any data in these systems yet blowing them away is probably the most straightforward approach.

rahearn (approved):

Issue #52, opened 2020-09-21
User story: Enable admin to assign permissions to Smart Hub users


pamlo412:
This user story will tie the spike and backend work from #39, #50, and #51 together so that we have a fully functioning admin tool in which to set or update user permissions for the Smart Hub.

**User story:** As a Smart Hub admin, I want to be able to specify the regions and read/write/admin access permissions of Smart Hub users so that the TTA team members can access what they need to complete their work in Smart Hub.

**Acceptance criteria:**

1. Launch the admin tool by entering this url: TBD
2. Search for or select a user for whom to set permissions.
3. Set permissions for key permutations as defined in #51 :
     - user A can access region x and has read/write permissions
     - user B can access region y and has read only permissions
     - etcetera - fill in the rest of these permutations once #51 is done

**Question for grooming:** can we test these permissions before login is complete?

pamlo412:
Agreed at grooming: we'll push this story back to the Backlog column for now. When #51 gets done, we will fill in the permission permutations and regroom this story.

pamlo412:
Revised job title requirements to change label and add list of roles provided by @PatricePascual-ACF. And here is the [source document for this list](https://docs.google.com/document/d/1XBsyrGWvKkpCbROvflOAsANibHX4y1B3ZlMSMuVv7Fw/edit#heading=h.cu097nvfyf7w).

PR #242, opened 2021-01-04
Admin UI talks to the API/DB

jasalisbury:
**Description of change**

 * Admin page now saves user changes up to the database
 * Add user seeder
 * add test users as a seeder
 * created at and updated at for users and permissions have default values of "now"
 * users "title" column renamed to "role"
 * GET `/user` returns the currently logged in user from the DB

**Notes**

 * Would be nice to have a validation on permissions that asserts any "global" scope id (the admin scope, "2") has to have a region id of 14 (no region). I tried to get a [model validation](https://sequelize.org/v5/manual/models-definition.html#model-wide-validations) working but something wasn't working correctly. Either need a ticket to write the validation or need to pair with someone to figure it out.

**How to test**

1) Go to https://tta-smarthub-dev.app.cloud.gov/admin
2) Select a user, update a field and save
3) Reload the page and note the updated field has the proper value

**Issue(s)**
* https://github.com/HHS/Head-Start-TTADP/issues/52

**Checklist**

- [x] Meets issue criteria
- [x] Code tested
- [x] Documentation updated
    - API methods


frontend/src/fetchers/index.js
```diff
@@ -0,0 +1,24 @@
+export const get = async (url) => {
+  const res = await fetch(url, {
```
rahearn, line 2: This should be using axios for IE11, right?

rahearn (changes requested):
I'm concerned at this point that we're still hardcoding the session user id instead of utilizing the `findOrCreateUser` function from `accessValidation.js`. If that isn't already scheduled work I think we need to just roll it into this PR.

frontend/src/fetchers/index.js
jasalisbury: IE11 support for fetch (among other things) is handled by the [polyfill](https://github.com/HHS/Head-Start-TTADP/blob/ff221a79a71ad1f1a09f8bac1f145088db77d8e3/frontend/src/index.js#L1) (https://www.npmjs.com/package/react-app-polyfill)

jasalisbury:
> I'm concerned at this point that we're still hardcoding the session user id instead of utilizing the `findOrCreateUser` function from `accessValidation.js`. If that isn't already scheduled work I think we need to just roll it into this PR.

One consequence of this change that I just realized... We will need to come up with some way of setting which users are Admins. Two ways I can think of off the top of my head:
1) We get a list of the `hsesUserId` `email` pairs and write a seed/migration adding the users and permissions
2) We write a stand alone script (run from the cli) that can elevate a normal user to Admin

1 seems like a good solution for prod/staging where hses ids should stay static. Not sure what happens to hses ids in the test environment when hses resets all accounts.

rahearn:
> 2\. We write a stand alone script (run from the cli) that can elevate a normal user to Admin

I think I prefer this one because it will work no matter the environment. Something that can be run with `cf run-task`. Can you create a card for that?

jasalisbury:
> > 2. We write a stand alone script (run from the cli) that can elevate a normal user to Admin
> 
> I think I prefer this one because it will work no matter the environment. Something that can be run with `cf run-task`. Can you create a card for that?

https://github.com/HHS/Head-Start-TTADP/issues/243

rahearn (approved):

Issue #88, opened 2020-10-09
Create job to pull Grantee Name and Grant ID  data from HSES into Smart Hub database


pamlo412:
**User story 1:**
As a TTAS creating an Activity Report, I want to be able to select (rather than type in) the grantees and associated grants my activity addressed so that I can avoid typos or mistakes that would make my report less accurate and searchable. 

**User story 2 (a specific case of User Story 1):**
As a specialist working with a grantee who has been transitioned to Region 13/CDI and given a new grant number, I want to be able to select the name of the grantee with the new CDI grant number so that I'm accurately referencing the grantee and grant.

**Acceptance criteria:**
- [ ] The user can select a grantee name and grant number from a list based on currently active grants
- [ ] The Hub recognizes when a grantee's grant has transitioned to CDI, flagging the first grant as inactive (and thus not available for selection in the Activity Report) and the CDI grant as the now active one available for selection in the report.
- [ ] The Hub recognizes when a grantee's grant number has changed because a new, continuation grant has been granted, and handles it the same as the CDI grant.
- [ ] The user can select multiple grantee name + grant number combinations (1 grantee + multiple grants and also multiple grantees and multiple grants)  

**Notes:**
- Dev plan is to create a job that pulls grantee and grant data from HSES into the Smart Hub database at some regular interval so that it is readily available for us to use in our UI.
- CDI grant numbers are all prefixed with "90CH"
- Glossary entry on CDI grants: https://github.com/HHS/Head-Start-TTADP/wiki/Region-13
- CDI employees will not be TTA Smart Hub users.
- Non-CDI specialists who were working with a grantee before they entered CDI retain access to the grantee's information while the grantee is in Region 13. 
- grantees moved to CDI will have new grant numbers so we can mark the previous grant numbers inactive
- it's fine, per Damon, if the user continues to select the old grant number until we have the updated data with the CDI grant number

**Exclusions:**
- help identifying the correct grant numbers. A grantee may receive multiple grants at the same time; for example, a grant for Head Start programming and a grant for recovery from a disaster. We may try to address this in the future. For the MVP, the specialists need to know which grant numbers to select.
- ability to enter the CDI grant number if it isn't yet in HSES but the specialist knows it. In this case, the specialist will select the pre-CDI grant number

pamlo412:
Per discussion at grooming today, I've updated this story with the requirements for handling Region 13/CDI grants and closed #96, the story which would have handled those requirements separately.

[...]

PR #248, opened 2021-01-08
Pull grant and grantee data from HSES on a schedule

kryswisnaskas:
**Description of change**
Adding a cron job, which is part of the server and runs everyday at midnights.
The job downloads data from HSES (in production env), extracts the zip archive into "temp" directory. Parses the xml data and populates or updates the SmartHub db with downloaded data.

**How to test**
Make sure the tests pass.
The imported data can be seen on sandbox in the database using the following:

- open SSH tunnel: cf connect-to-service -no-client tta-smarthub-sandbox ttahub-sandbox
- use your favorite postgres client with the parameters listed (if can't connect with localhost, use 127.0.0.1)
- verify that there are grantees and grants in tables

(bonus test) additionally, to verify the download:

- modify one of the HSES data download environment variables:
- HSES_DATA_FILE_URL, HSES_DATA_USERNAME, HSES_DATA_PASSWORD
- modify the cron's schedule to '*/5 * * * *' in the app.js (run every 5 minutes instead of at midnight)
- deploy a branch to sandbox
- observe that there will be an error message

**Issue(s)**
* closes https://github.com/HHS/Head-Start-TTADP/issues/88

**Checklist**

- [x] Meets issue criteria
- [x] Code tested
- [n/a] Documentation updated
    - API methods


src/lib/updateGrantsGrantees.js
```diff
@@ -0,0 +1,115 @@
...
+    // process grantees aka agencies that are non-delegates
+    const agencyData = await fs.readFile('./temp/agency.xml');
+    const agency = JSON.parse(toJson(agencyData));
+
+    // filter out delegates by matching to the non-delegates
+    // eslint-disable-next-line max-len
+    const grantees = agency.agencies.agency.filter((a) => grantGrantees.some((gg) => gg.agency_id === a.agency_id));
+
+    grantees.forEach((g) => granteesForDb.push({
+      id: parseInt(g.agency_id, 10),
```
rahearn, line 48: I think we need a new `externalId` or `hsesId` column for both `Grants` and `Grantees` to use for the IDs we get from this data, and allow the `id` column to be auto-incremented by postgresql. Explicitly assigning to the auto-incremented column will cause problems down the road.

How are we merging (or matching) the grantees from this data source and those pulled in via the tta goals?

src/lib/updateGrantsGrantees.js
```diff
@@ -0,0 +1,115 @@
...
+        updateOnDuplicate: ['number', 'regionId', 'granteeId', 'status', 'startDate', 'endDate', 'updatedAt'],
+      });
+  } catch (error) {
+    logger.error(`Error reading or updating database on HSES data import: ${error.message}`);
+    throw error;
+  }
+}
+
+// reading archives
+const zip = new AdmZip('./hses.zip');
```
rahearn, line 85: Should we remove the individual `temp/*.xml` files from git? I can see this being confusing if someone updates that source data thinking they're updating a test, only for this line to overwrite their changes.

Either that or leave them, remove the base `hses.zip` file, and update the test to skip over the zip extraction part.

tools/importPlanGoals.test.js
```diff
@@ -7,6 +8,39 @@ describe('Import TTA plan goals', () => {
   afterAll(() => {
     db.sequelize.close();
   });
+  describe('Update grants and grantees', () => {
```
rahearn, line 11: I like these tests, but I'm confused why they're in this file rather than one that matches the filename where the source is.

tools/importPlanGoals.test.js
```diff
@@ -7,6 +8,39 @@ describe('Import TTA plan goals', () => {
...
+      expect(granteesBefore.length).toBe(0);
+      await updateGrantsGrantees();
+
+      const grantee = await Grantee.findOne({ where: { id: 1335 } });
+      expect(grantee).toBeDefined();
+      expect(grantee.name).toBe('Agency 1, Inc.');
+    });
+
+    it('should import or update grants', async () => {
+      const grantsBefore = await Grantee.findAll();
```
rahearn, line 32: I think you meant to check the count of `Grant` before the test?

```suggestion
      const grantsBefore = await Grant.findAll();
```

rahearn (changes requested):
Hi @kryswisnaskas, good stuff. My biggest concern is the mixing of explicit IDs with auto-increment columns. As we saw with the user seeders in dev and sandbox doing that will likely result in duplicate IDs where we didn't mean to have them.

src/lib/updateGrantsGrantees.js
rahearn: I could also see removing the auto-increment and relying on HSES to provide the IDs for these two classes, though I'm a little squeamish on trusting any canonical IDs in our system to any outside system.

.circleci/config.yml
```diff
@@ -86,7 +95,10 @@ commands:
               --var AUTH_CLIENT_ID=${<< parameters.auth_client_id >>} \
               --var AUTH_CLIENT_SECRET=${<< parameters.auth_client_secret >>} \
               --var NEW_RELIC_LICENSE_KEY=${<< parameters.new_relic_license >>} \
-              --var SESSION_SECRET=${<< parameters.session_secret >>}
+              --var SESSION_SECRET=${<< parameters.session_secret >>} \
+              --var HSES_DATA_FILE_URL=${<< parameters.hses_data_file_url >>} \
```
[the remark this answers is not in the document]
kryswisnaskas: I was under impression that the file URL was also confidential to further hide the information in order to reduce the attack surface. If we are confident that it is not, I can move it.

src/app.js
```diff
@@ -77,4 +79,23 @@ if (process.env.NODE_ENV === 'production') {
   });
 }
 
+// Set timing parameters.
+// Run at midnight
+// const schedule = '0 0 * * *';
+const timezone = 'America/New_York';
+
+// tmp schedule for testing
+const schedule = '*/25 * * * *';
```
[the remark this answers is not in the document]
kryswisnaskas: Ah, yes, this slipped by. I'll take a look at the link. Thanks.

.circleci/config.yml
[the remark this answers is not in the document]
rahearn: Well, I don't think it should be, but upon further reflection it is not our URL to make that choice, so please do leave this as-is. Better safe than sorry.

src/lib/updateGrantsGrantees.js
```diff
@@ -0,0 +1,115 @@
+import AdmZip from 'adm-zip';
+import { toJson } from 'xml2json';
+import {} from 'dotenv/config';
```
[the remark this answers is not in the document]
kryswisnaskas: This is incorporates using the dotenv module. More info on the syntax:
https://github.com/motdotla/dotenv/issues/89

src/lib/updateGrantsGrantees.js
kryswisnaskas: > How are we merging (or matching) the grantees from this data source and those pulled in via the tta goals?

That is on my list of modifications to the goals import script. The main idea was to change the code to find grants and grantees in the database and use the ids for the GrantGoals table - if running in production or other deployed environments. For local development, I'll have to think what would be the best solution.

kryswisnaskas: I could also see removing the auto-increment and relying on HSES to provide the IDs for these two classes, though I'm a little squeamish on trusting any canonical IDs in our system to any outside system.

If we could re-use HSES ids which are already present in 3 tables, it would give us "free" ids. I was envisioning that all grants and grantees would come from HSES, so we wouldn't need to insert any real grantees and grants. If that's a safe assumption, then we could adjust how we are entering test data.

The alternative would be to rely on auto increment for the HSES pull, but then we couldn't use the bulk operations and rely on something like "findOrCreate". This is fine, but not as fast as the bulk inserts and updates. The only advantage of the autoincrement would be an easier handling of the test data (unless I am missing something)

src/lib/updateGrantsGrantees.js
kryswisnaskas: I can refactor this.

tools/importPlanGoals.test.js
kryswisnaskas: Good observation. Originally, I had them in a separate file, but due to `jest `running tests in parallel, I was getting inconsistent results. Perhaps we can postpone separating these for later to investigate exactly why the discrepancies. I didn't think spending more time on this now was worth it.

tools/importPlanGoals.test.js
kryswisnaskas: Good catch!

tools/importPlanGoals.test.js
rahearn: My guess is that it's an impact of the mix of explicit ids and auto increment.

src/lib/updateGrantsGrantees.js
rahearn: > I was envisioning that all grants and grantees would come from HSES, so we wouldn't need to insert any real grantees and grants

this is almost certainly true.

My biggest worry is _if_ we find out that a grantee for whatever reason gets assigned a new ID in hses, how do we handle it on our side. My impression had been that there would be a new agency object on each new 5 year grant cycle. But I think we've seen in the data that that would not be the case. There would be one agency with the old inactive grant and a new active grant, right?

> For local development, I'll have to think what would be the best solution.

if we go the explicit ID route, we could use the test data in /temp/*.xml as our test/dev data across the whole system. Add them to seeders like the users and permissions used in those environments.

tools/importPlanGoals.js
```diff
@@ -163,15 +157,16 @@ export default async function importGoals(file, region) {
...
+            const fullGrant = { number: grant.trim(), regionId };
+            const dbGrant = await Grant.findOne({ where: { ...fullGrant }, attributes: ['id', 'granteeId'] });
+            if (!dbGrant) {
+              console.log(`Couldn't find grant: ${fullGrant.number}. Exiting...`);
+              process.exit(1);
             }
-            const [dbGrant] = await Grant.findOrCreate({ where: { ...fullGrant, regionId } });
             grantId = dbGrant.id;
+            currentGranteeId = dbGrant.granteeId;
+            // }
```
rahearn, line 169:
```suggestion
```

rahearn (changes requested):
Great stuff, Krys. There are just two minor things I'd like to change before merge, both with suggestions below.

.circleci/config.yml
```diff
@@ -363,6 +387,9 @@ jobs:
                 deploy_config_file: deployment_config/prod_vars.yml
                 new_relic_license: PROD_NEW_RELIC_LICENSE_KEY
                 session_secret: PROD_SESSION_SECRET
+                hses_data_file_url: HSES_DATA_FILE_URL
+                hses_data_username: HSES_DATA_USERNAME
+                hses_data_password: HSES_DATA_PASSWORD
```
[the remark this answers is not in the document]
kryswisnaskas: Ah, good catch. Cut and paste omission.

rahearn (approved):

Issue #98, opened 2020-10-15
Saving an Activity Report


pamlo412:
**User story:**
As a Regional TTA Specialist drafting a report on one of my TTA Activities, I want to be able to save my work in progress, without having to submit it for approval, so that I can draft reports on my various activities with grantees piecemeal when I have windows of time. 

**Acceptance Criteria:**
**Backend**
- [x] the activity reports table(s) is/are created
- [x] the API endpoint and documentation is available for the front end to save (GET, POST, PUT)
- [x] users can't save until they at least enter the grantee name 
- [x] the system captures the author and a datetime stamp on saving (so that we can display "last modified by" (name) and "last modified" (date))

**Frontend**
- [x] the system autosaves to the server every 2 minutes 
- [x] a message displays to users when an autosave or manual save happens:
     - [x] Message content: This report...[wording in design]
- [x] on reopening a saved report, users can edit the fields and re-save the report

**Exclusions:**
- document upload API endpoint https://github.com/HHS/Head-Start-TTADP/issues/150
- submitting the report for review/approval - #135
- collaborating with other specialists - #109 
- display of saved report on Home page - #181 
- local saving in the case of Internet connection going down - post MVP
- ability to draft and save offline and pull it into Smart Hub when you're back online - post MVP #144 
- assignment of report ID - #145

SarahJaine:
Assigning this ticket to myself for Backend work only

SarahJaine:
Per conversation in planning today, I'm excluding all cols that are displayed as multiselect FE components from this ticket:
- targetPopulation https://github.com/HHS/Head-Start-TTADP/issues/173
- headStartProgramType https://github.com/HHS/Head-Start-TTADP/issues/170
- granteeParticipantRole and otherParticipantRole https://github.com/HHS/Head-Start-TTADP/issues/172
- reason (example: 'Change in Scope',)
- topic (example: 'Behavioral / Mental Health',)

PR #254, opened 2021-01-15
Saving an activity report

### LINKED ISSUE #73: Stop using memory session store for `express-session`
We use [express session](https://www.npmjs.com/package/express-session) to manage user sessions. The default session store is the memory session store. This store is not appropriate for production and will leak memory. We need to figure out which session store we should use for our deployed environments.

This story could easily be split into two, one being a Spike to determine which session store to use and a second to implement the decision made in the spike.

**Acceptance Criteria**

 - [ ] A new session store is picked from [this list](https://www.npmjs.com/package/express-session#compatible-session-stores) Note: session store needs to be created via terraform as a cloud foundry/cloud.gov artifact, so we're probably limited to redis or rds postgres
 - [ ] Any new infrastructure that is needed is added to terraform
 - [ ] The backend uses the new session store

### LINKED ISSUE #255: Sandbox and dev db migration failing
Seeing this error on the db migration step for sandbox and dev whenever there is a seed for a new table

```
14:14:50.638: [APP/TASK/18d59cfb.0] == 20210107170250-nonGrantees: reverting =======
14:14:50.650: [APP/TASK/18d59cfb.0] ERROR: relation "NonGrantees" does not exist
14:14:50.666: [APP/TASK/18d59cfb.0] error Command failed with exit code 1.
```

jasalisbury:
**Description of change**

Add the ability to save, retrieve and update an Activity Report. Also provides an endpoint the frontend will need to complete an Activity Report (`activity-report/participants`). Activity Reports use the Activity Participants table to join to many grants or non-grantees. This PR only contains backend changes. Frontend changes are coming in a separate PR.

 * DB tables added for `ActivityReports`, `ActivityParticipants` and `NonGrantees`
 * API documentation updated
 * DB diagram updated
 * Adds two helpers to the `services` folder so controllers don't have to directly talk to the database layer
 * Uses already written seeders for grantees/grants but adds a seeder for non-grantees
 * Add virtual column to grants making `{grantee.name} - {grant.number}` available as a column

Jest tests are run in parallel by default. This was causing issues for tests that hit the database, as one test would remove items a different test relied on. We are now passing the `--runInBand` flag to jest so tests run sequentially. After the MVP we may want to look at allowing tests to run in parallel again.

**How to test**

1) Pull down changes
2) Install dependencies
3) Run migrations and seeders
4) Run the following against the API

```
POST http://localhost:3000/api/activity-reports
{
    "activityParticipants": [
        {
            "participantId": 1
        }
    ],
    "participantType": "grantee"
}
```
Note the ID returned (will probably be 1)

```
GET http://localhost:3000/api/activity-reports/{id}
```
```
PUT http://localhost:3000/api/activity-reports/{id}
{
    "additionalNotes": "notes"
}
```
```
GET http://localhost:3000/api/activity-reports/participants
```
**Issue(s)**
* https://github.com/HHS/Head-Start-TTADP/issues/98
* closes #255 
* closes #73 

**Checklist**

- [ ] Meets issue criteria
- [x] Code tested
- [x] Documentation updated
    - API methods


rahearn:
@jasalisbury I'll take a look at this soon, or first thing Monday. Does this complete 98 or is there something else to do on it before we can close it out?

jasalisbury:
> @jasalisbury I'll take a look at this soon, or first thing Monday. Does this complete 98 or is there something else to do on it before we can close it out?

#98 will still need https://github.com/adhocteam/Head-Start-TTADP/pull/118 before it can be closed

src/migrations/20210106160931-create-activity-reports.js
```diff
@@ -0,0 +1,114 @@
...
+      },
+      endDate: {
+        allowNull: true,
+        type: Sequelize.DATEONLY,
+      },
+      startDate: {
+        allowNull: true,
+        type: Sequelize.DATEONLY,
+      },
+      participantType: {
```
rahearn, line 38: Is this for Grantees vs. NonGrantees? I'm feeling some naming confusion over this field vs. `participants` and `numberOfParticipants`.

I wonder if we even need to store this vs figuring it out from the ActivityParticipants relationships

src/migrations/20210106160953-create-activity-participants.js
```diff
@@ -0,0 +1,53 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    queryInterface.createTable('ActivityParticipants', {
```
rahearn, line 3: Kinda similar naming confusion here. I was thinking of Participants as individual people, or roles at the grantee that were at the activity. The PR is using participants for both that as well as the Grantee(s) or NonGrantee(s)

src/routes/activityReports/handlers.js
```diff
@@ -1,9 +1,9 @@
-import { Op } from 'sequelize';
-import {
-  User, Permission,
-} from '../../models';
 import handleErrors from '../../lib/apiErrorHandler';
 import SCOPES from '../../middleware/scopeConstants';
+import {
+  reportParticipants, activityReportById, createOrUpdate, reportExists,
+} from '../../services/activityReports';
+import { userById, usersWithPermissions } from '../../services/users';
```
rahearn, line 6: ❤️ pulling these into services.

rahearn (commented):
I'm a bit concerned about the overloaded naming around participants. Did I interpret what's going on correctly?

src/migrations/20210106160931-create-activity-reports.js
jasalisbury: Yes it is for Grant vs NonGrantee. We probably could get away with not storing it but it feels easier to store it here instead of looking at the `ActivityParticipants` relationship. Although the only time we'll need to look it up is when retrieving activity reports. Creates and updates would stay pretty much the same.

I agree with the naming confusion but couldn't come up with a bettername for `grant` or `nonGrantee`. Best I could think of (and this isn't a good name at all!) is `subjects`, which also seems a bit overloaded.

src/migrations/20210106160953-create-activity-participants.js
jasalisbury: Yeah I agree it is confusing, I just lived with the confusion enough that I forgot about it! I'll ping the other engineers and try to come up with a better name.

package.json
```diff
@@ -18,7 +18,7 @@
     "server:debug": "nodemon src/index.js --exec babel-node --inspect",
     "client": "yarn --cwd frontend start",
     "test": "jest src",
-    "test:ci": "cross-env JEST_JUNIT_OUTPUT_DIR=reports JEST_JUNIT_OUTPUT_NAME=unit.xml POSTGRES_USERNAME=postgres POSTGRES_DB=ttasmarthub CURRENT_USER_ID=5 CI=true jest src tools --coverage --reporters=default --reporters=jest-junit",
+    "test:ci": "cross-env JEST_JUNIT_OUTPUT_DIR=reports JEST_JUNIT_OUTPUT_NAME=unit.xml POSTGRES_USERNAME=postgres POSTGRES_DB=ttasmarthub CURRENT_USER_ID=5 CI=true jest src tools --runInBand --coverage --reporters=default --reporters=jest-junit",
```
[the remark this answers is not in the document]
jasalisbury: The main issue is when tests run in parallel they are independently inserting and removing items from the DB, which causes races within tests. I can spend some time trying to see if there are any obvious offenders but it may be a task better suited to after the MVP.

src/migrations/20210106160953-create-activity-participants.js
jasalisbury: What do you think about renaming `ActivityParticipants` to `ActivityReceivers` or `ActivityRecipients`?

jasalisbury:
This PR now also includes work for [this ticket](https://github.com/HHS/Head-Start-TTADP/issues/73)

rahearn (approved):
🎉

Issue #271, opened 2021-01-28
As an account administrator, I want to be able to lock a user who has been inactive for a long time, to keep my system secure (IA-04)


rahearn:
Add the ability to lock a user account from login, without removing any other permissions, to make re-activating easier.

Issue #109, opened 2020-10-21
Add collaborators to draft Activity Reports


pamlo412:
**User story:**
As a specialist who is drafting an activity report, I want to select another collaborator from my region so they can review or edit the draft report.

**Acceptance Criteria:**
- [ ] author of a report that is still in draft can select one or more TTA Hub users from their region with whom they want to collaborate on the report
- [ ] TTA Hub users in the current user's region should be selectable 
- [ ] type-ahead is available to help narrow the list
- [ ] the selected user(s) are granted editing and submitting permissions for the draft report
- [ ] changes that collaborating users make in the report are not tracked (i.e. no version history or comments or ability to accept/reject)
- [ ] Post-MVP: Nice to have: the user has a way to find and select someone outside of their region (for Regions 11 and 12, espcially)

**Exclusions:**
- any sort of visual indicator (like a comment box) for collaborating specialists to note for each each other what changes they made or edits they recommend. This kind of indicator is only available in the manager's workflow for requesting revisions - see #138
- any sort of visual indicator allowing a collaborator to alert other collaborators when they're done they're part of the report work - the report will simply sit in the collaborators' queues on their Home page until one of them submits it and all listed managers approve it - see #
- any sort of tracking of changes/version/comments  
- submitting the report for approval, #135 
- actual push notifications - this work will be wanted in the future
- ability to see the report in the Home page list, # TBD

Issue #265, opened 2021-01-22
Missing authorization checks on activity report editing


rahearn:
The activity reports handler should be checking permissions before allowing creating and updating activity reports. Right now it is only validating that the user is logged in.

PR #272, opened 2021-01-28
Save collaborators to report

jasalisbury:
**Description of change**

Includes a few additional improvements, got bigger because I refactored some tests to hopefully make them easier to deal with in the future.

 * Collaborators join table added.
 * Collaborators hooked up on the frontend. Available collaborators depends on the report's region (hardcoded to 1 for now)
 * Activity report checks permissions
 * Added the first "policies" which can be used for authorization. Should be easier to test separated from controllers
 * Route tests that hit the DB moved to the services folder, only deal with the DB interactions
 * Route tests mock out DB and authorization calls focusing on constructing correct responses based on (mocked) calls to the DB/policies
 * Multiselect sets null values to an empty array, allowing collaborators/recipients to be "zeroed" in the report
 * Approvers are scoped to region
 * The activity report now makes all API calls up front in parallel. Should keep other components simpler and cuts down on calls to the API (they can be made once instead on every form page load)
 * Added Krys and I to seeders as Admins so we should be admins in sandbox/dev until HSES resets again. (can add others to the seeders but we were the only two who currently have accounts on sandbox, so the only two I know HSES ids)
 * Updated swagger/API docs

**How to test**

1) Browse to [dev](https://tta-smarthub-dev.app.cloud.gov/activity-reports/new)
2) Fill out a report, add Chuck or Josh as a collaborator (If you aren't on the list I can add you, just ping me. Permissions should be setup correctly on Dev unless another deploy happens)
3) Verify collaborators persist across reloads
4) Send the ID of the report to someone who isn't a collaborator and verify they can't access the report

**Note**
I still need to update API docs/DB docs

**Issue(s)**
* closes https://github.com/HHS/Head-Start-TTADP/issues/109
* https://github.com/HHS/Head-Start-TTADP/issues/265
* closes https://github.com/HHS/Head-Start-TTADP/issues/271

**Checklist**

- [x] Meets issue criteria
- [x] Code tested
- [x] Documentation updated
    - API methods


src/migrations/20210120220221-add-collaborators-to-activity-reports.js
```diff
@@ -0,0 +1,50 @@
...
+        type: Sequelize.INTEGER,
+        allowNull: false,
+        references: {
+          model: {
+            tableName: 'ActivityReports',
+          },
+          key: 'id',
+        },
+        onUpdate: 'CASCADE',
+        onDelete: 'CASCADE',
```
rahearn, line 20: I always mess these up. Does this delete the ActivityReport when a collaborator is removed, or delete the collaborator automatically when an ActivityReport is deleted? (Same question for user)

src/models/activityReport.js
```diff
@@ -16,6 +16,16 @@ export default (sequelize, DataTypes) => {
       ActivityReport.belongsTo(models.User, { foreignKey: 'lastUpdatedById', as: 'lastUpdatedBy' });
       ActivityReport.belongsTo(models.User, { foreignKey: 'approvingManagerId', as: 'approvingManager' });
       ActivityReport.hasMany(models.ActivityRecipient, { foreignKey: 'activityReportId', as: 'activityRecipients' });
+      ActivityReport.belongsToMany(models.User, {
+        through: models.ActivityReportCollaborator,
+        // The key in the join table that points to the model defined in this file
+        foreignKey: 'activityReportId',
+        // The key in the join table that points to the "target" of the belongs to many (Users in
+        // this case)
```
rahearn, line 24: A+ comment to keep the two keys straight.

src/routes/activityReports/handlers.js
```diff
@@ -21,13 +22,9 @@ const logContext = {
  * @param {*} res - response
  */
 export async function getApprovers(req, res) {
-  const user = await userById(req.session.userId);
-  const reportRegions = user.permissions
-    .filter((p) => p.scopeId === READ_WRITE_REPORTS)
-    .map((p) => p.regionId);
-
+  const { region } = req.query;
```
rahearn, line 25: Should this method should be using the Users policy? I think there are a couple of methods still with no authorization, so I'm leaning towards not making any changes here beyond the good work you've already done, and just keeping #265 open for the other routes.

src/routes/activityReports/handlers.test.js
```diff
@@ -35,171 +24,170 @@ const mockResponse = {
...
-      permissions: [
-        {
-          userId: 51,
-          regionId: 6,
-          scopeId: SCOPES.APPROVE_REPORTS,
-        },
-      ],
+
+    it('returns the report', async () => {
+      ActivityReport.prototype.canUpdate = jest.fn().mockReturnValue(true);
```
rahearn, line 52: This is some of my javascript ignorance showing, but, how does this implementation impact the rest of the tests? Does this pollute everything? Or is it somehow scoped to this test?

rahearn (changes requested):
Awesome work. I really like the policies classes.

src/migrations/20210120220221-add-collaborators-to-activity-reports.js
jasalisbury: I wasn't 100% sure either:

> CASCADE specifies that when a referenced row is deleted, row(s) referencing it should be automatically deleted as well.

from https://www.postgresql.org/docs/9.5/ddl-constraints.html

src/routes/activityReports/handlers.js
jasalisbury: :+1: that sounds good to me. There may be a few other routes to update that have been added in-between starting this work and getting it merged in to main. We'll probably just want to audit all our routes on main and make sure every route added after is using a policy (or explicitly doesn't)

src/routes/activityReports/handlers.test.js
jasalisbury: That is a good question, I'll look into it.

jasalisbury:
> Awesome work. I really like the policies classes.

Stole the pattern from a great ruby library https://github.com/varvet/pundit

rahearn:
> Stole the pattern from a great ruby library https://github.com/varvet/pundit

Probably why I like it, I use pundit on all of my ruby apps.

rahearn (approved):

Issue #274, opened 2021-02-01
Allow searching of users be case-insensitive


gopar:
*User Story*
As a user, I want to be able filter users without needing to be case sensitive aware

Issue #263, opened 2021-01-22
Backend tests too brittle.


rahearn:
Looks like we're running into a lot of issues with tests clobbering the DB for other tests.

adding `--runInBand` might be a bandaid for this issue, but I'd like to solve the root issue for the long term.

https://app.circleci.com/pipelines/github/HHS/Head-Start-TTADP/165/workflows/bae0eaf5-5821-4682-940b-837dc796c434/jobs/922

dcloud:
Just started taking a look at this. It sounds like people are seeing errors in tests that rely on the database, and the thinking is that one tests's db transactions aren't completing before another db-dependent test runs, or something to that effect.

The evidence in CircleCI, based on the test run linked above, is errors like:

```
SequelizeForeignKeyConstraintError: insert or update on table "Permissions" violates foreign key constraint "Permissions_userId_fkey"
```

It sounds like in that case the errors are related to creation of users and permissions, and [another test run (jobs/923)](https://app.circleci.com/pipelines/github/HHS/Head-Start-TTADP/165/workflows/a0068554-d8dc-4bdd-a978-3d4ae9aeff69/jobs/923) shows those same errors. An [earlier test run (jobs/920)](https://app.circleci.com/pipelines/github/HHS/Head-Start-TTADP/165/workflows/0ecc8659-8c6b-4fff-9377-a6f8987ee109/jobs/920) also shows db errors, but for foreign key constraints on ActivityReports.

So we need to work on guaranteeing that db is in a "clean and ready" state at the beginning of each tests. It also sounds like this may only be happening in CircleCI? Need to investigate how to reproduce, and how to make sequelize and Jest play nice with each other (and CircleCI, probably)

rahearn:
I think it's happening locally too, though perhaps less often? It's definitely a bit of a race condition though so it's difficult to get a consistent reproduction.

I think there must be some best-practices around test setup that we're not doing that are likely all that are needed here. In a rails app this would be handled with wrapping tests in transactions to rollback any changes after the test finishes.

dcloud:
Yeah, it looks like this does happen locally, and that might be more complicated, since the `dbdata` volume persists between runs. Also noticed that `yarn run docker:test:be` doesn't use the same docker-compose-test.yml that circleci uses.

I agree that using transactions might be part of the path forward. I'll be looking at that in a bit.

rahearn:
> `dbdata` volume persists

[...]

Issue #236, opened 2020-12-22
Manager update of AR status


marriej:
As a manager I want to be able to update the AR status.

**Acceptance criteria:**
- [x] while reviewing a submitted Activity Report, a manager can 
     - [x] select  "Needs Action"
     - [x] select "Approved"
**Design:**
https://preview.uxpin.com/ad995fd722bd26c0f9a1c8a115f1b41833637308#/pages/134880478/simulate/sitemap

Issue #231, opened 2020-12-17
Approving manager adding note to activity report


pamlo412:
**User story:**
As an approving manager I want to add notes in a notes box to request changes, therefore I need permissions to to edit the notes box. I don't need to edit the rest of the report. 
There is only one approving manager.

Acceptance Criteria:

- [x] Only the approving manager as designated by the report submitter can add a note.
- [x] This is the only field an approving manager can edit.

**Design Link:**https://preview.uxpin.com/ad995fd722bd26c0f9a1c8a115f1b41833637308#/pages/134880478/simulate/sitemap

**Notes:** (history behind this issue)
  - this story was broken out from #138 because we weren't certain whether we should let managers edit whatever they want or constrain them to requesting changes (unless they are a collaborator). Both options were deemed technically feasible, so the user need is what's paramount here. Patrice will get with Damon and discuss this issue with some TTACs and other managers for their input.
  - the biggest concern we had is that letting managers edit during the approval phase could create confusing content,  if they change something in one place that the author references in another place in the report. The comparison was made to Twitter: if you comment on a tweet and the author then changes their tweet, your comment may look irrelevant or mistaken or otherwise confusing.

#update 1/25/21 
Note  - only one approving manager for MVP

PR #281, opened 2021-02-04
Case insensitive admin search, testing updates and manager setting of notes and report status

jasalisbury:
## Allow search to be case-insensitive
**Description of change**
User search is now case-insensitive in the admin page

**How to test**
1. Pull down changes
2. Go to admin page
3. Type a users name in a mixture of upper/lower and you should still get the desired user (if exists)

**Issue(s)**
* closes https://github.com/HHS/Head-Start-TTADP/issues/274

## Create separate docker-compose file for running tests, add bin/run-tests command for easy testing

**Description of change**

While looking at issues with the backend tests, I saw some room for improvement with how docker tests run locally. The `bin/run-tests` file is the culmination of my attempts to sandbox docker-tests from docker-dev environment. 

I tried [multiple docker-compose.yml](https://docs.docker.com/compose/extends/) files, but that had many issues. For one, overrides don't override multivalve options. Another issues was Docker not treating the different configurations as different services, so it wasn't using the test-db when expected. 

I looked at the `circleci` CLI, but that only runs jobs, not workflows, and the circleci needed to run tests is two jobs. There may be something useful here, but also introduces another tool.

If you look in the commit history, you can also see that I wrote a much longer bash script initially, in part to help me experiment with multiple docker-compose.yml files and in part to compose a number of smallish docker commands into larger actions. When I realized multiple docker-compose.yml files wasn't going to work, this tool became unnecessary. You could retrieve it via `git checkout 8b0beced865b3a9733bb5c61123dcc9d33451981^ -- bin/docker-run` if you're interested.

**How to test**

Run `./bin/run-tests` then start your regular Docker environment. Running tests should not overwrite/modify your development database.

**Issue(s)**
* closes https://github.com/HHS/Head-Start-TTADP/issues/263

## Managers can add notes and set status of reports
**Description of change**
 * Review page is now different if the user looking at the page is the approving manager
 * Manager's review page includes the user's additional note and allows the manager to leave a manager note
 * Manager note added to DB
 * Report state is shown in the side nav's "review" section once it leaves "draft" mode
 * Policy, service and route added to update status of a report

**How to test**
1) Pull down, run migrations/seeders
2) Make sure the `CURRENT_USER_ID` is set to "1"
3) Fill out a complete report. Add "Hermione" as the approving manager
4) Refresh the report, since Hermione is user with the id 1 you should see the approving manager version of the review page.
5) Add a manager note and set the status. Note once you submit the status on the side nav `Review and submit` section updates
6) Refresh the page to verify the manager notes persists

**Issue(s)**
* closes https://github.com/HHS/Head-Start-TTADP/issues/138
* closes https://github.com/HHS/Head-Start-TTADP/issues/231
* closes https://github.com/HHS/Head-Start-TTADP/issues/236
* closes https://github.com/HHS/Head-Start-TTADP/issues/137

**Checklist**

- [x] Meets issue criteria
- [x] Code tested
- [x] Documentation updated
    - API methods


rahearn (changes requested):
Code looks good, but in my testing there are a couple of issues with the activity report that I'd like to see cleaned up.

1) When the approving manager is a different user than the report creator, the auto-save fails (good! they shouldn't be able to save the report details) and shows a failure message. Please disable autosave when the approving manager is reviewing.

2) I don't think this is related to these changes, but I'm often getting a failure where the save status is not displayed. Looks like it is related to this console error: 
Screen Shot 2021-02-04 at 12 16 57 PM

bin/run-tests
```diff
@@ -16,7 +16,7 @@ docker exec test-backend bash -c "yarn db:seed;"
 docker exec test-backend bash -c "yarn test:ci"
 
 # Test frontend
-docker exec test-backend bash -c "yarn --cwd frontend run test:ci"
+docker exec test-frontend bash -c "yarn --cwd frontend run test:ci"
```
rahearn, line 19: This doesn't work. Looks like a missing `container_name` for the frontend service. `--cwd frontend` is also unnecessary (though doesn't break anything) in the frontend container.

Screen Shot 2021-02-05 at 9 35 36 AM

That also exposes that this script papers over any failures if you don't read the output carefully. Since this is not used in CI/CD I think I'm ok with that, but I think it'd be good to collect exit statuses and exit the script with non-zero if any individual step exists with non-zero.

dcloud: Hmmm, I've run this a few times and haven't seen this problem. I'll take a look again. Could you get me the output of `docker-compose -f docker-compose.test.yml ps --all` on your machine?

As for handing failures, it was intentional that we didn't error out if one of the intermediate steps threw an error. We could catch those statuses, but I'm not sure we want to prevent frontend tests from running if the backend tests fail, for example.

rahearn: > I'm not sure we want to prevent frontend tests from running if the backend tests fail, for example

Yeah, I went back and forth on that. I'm 100% fine with all the steps running as long as I get a nice clear indication that at least one step failed when the script exits.

Here's the `ps` output:

Screen Shot 2021-02-05 at 10 20 14 AM

dcloud: Just a quick check of the doc page for [`docker-compose exec`](https://docs.docker.com/compose/reference/run/), it sounds like that command takes a _service_ name rather than a _container_ name, so that might be related.

While we're discussing, I had originally considered having the frontend and backend tests run as separate scripts, since  it seems like engineers might be focused on either frontend or backend during development. Would you be interested in making this accept an optional frontend/backend param?

rahearn: > Would you be interested in making this accept an optional frontend/backend param

If it's quick, sure, but that's pretty low priority to me.

dcloud: It was quick: https://github.com/adhocteam/Head-Start-TTADP/pull/149/files

I also added some logging of errors and have the script return an exit code value equal to the number of script failures.

rahearn (approved):
🎉

PR #598, opened 2021-10-29
simplify number of widget routes

thewatermethod:
## Description of change
Condensed the Overview Widget & Dashboard Widget routes, which were very similar, into one route. Both front-end variations use the same back-end route now.

## How to test
Visit the /activity-reports page and the /regional-dashboard page. Note that both overview widgets work and that the data is correct.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-267

### Every PR

- [x] Meets issue criteria
- [x] Boundary diagram updated
- [x] [Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions

### Production Deploy

### After merge/deploy


src/widgets/overview.js
```diff
@@ -1,74 +1,76 @@
...
-  by the widget. In this case we return a single object but other widgets my require an array
-  (say for a time series). All widgets will need to honor the scopes that are passed in. All
-  that is required is that the scopes parameter is used as value for the `where` parameter (or
-  combined with [op.and] if the widget needs to add additional conditions to the query).
+export default async function dashboardOverview(scopes) {
+  const duration = await ActivityReport.findAll({
+    attributes: [
+      'duration',
+      'deliveryMethod',
+      'numberOfParticipants',
```
jasalisbury, line 13: Can we combine these two sql queries and sum/count these attributes the same way has reports and grants/non-grantees? I would think `duration` and `numberOfParticipants` just need a `SUM(..)` and `deliveryMethod` follows the pattern of the other three columns.

thewatermethod: Wouldn't inPerson require a separate or inline query given you need to add up the duration of only reports that have "in person" as their delivery method?

thewatermethod: another issue seems to be that some of the legacy data throws off the SUM
For example, 
```js
[sequelize.fn('SUM', sequelize.col('"ActivityReport".numberOfParticipants')), 'numParticipants'],
```
yields 769,802 whereas the way it is now, with the ```reduce``` is 76,944. Seems like at least some of the data is stored as a string.

A sum of the duration in the same way is also wildly inflated, although I'm not exactly sure why. Maybe it's the same issue. 

Not sure if they is a SQL forward way to write this instead of the JS solution I have.

jasalisbury: I think I remember what is going on. The `required: false` causes the joins to be left outer joins, which will give us duplicated activity reports. If a report has five recipients that report will be returned five times, which is messing with the results. There is probably a way to solve it in sql but it isn't worth it, I'd keep the two queries. Can you add a comment explaining the joining situation and why we can't have a single query? I'm almost positive this is the second time I've wonder about this, hopefully a comment will keep me from wasting everyones time a third time!

jasalisbury (dismissed):

src/widgets/overview.js
thewatermethod: Haha, will do. Not a time waster to think it through and get it right, IMO

Looks like adding the comment dismissed your review!

jasalisbury (approved):
:shipit:

jasalisbury:
Oh another thing to be aware of, we are removing a widget route. If a user has their browser open when we deploy they will have a broken frontend that asks for a widget that no longer exists. It seems like we have a built in refresh on the frontend with the inactivity timeout. They will have to re-login (and pull in new frontend code) at least every morning (I'm _sort_ of assuming that). We should probably keep the old route around for a couple of deploys so people will have time to pull in the new frontend code.

PR #601, opened 2021-11-01
Add initial ADR discussion

hardwarehuman:
This is a discussion of how an RDS Read Replica might help solve three separate issues we have. After review, only the second item seems like it could go either way.

## Issue

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-401

### Every PR

- [X] Meets issue criteria

### After merge/deploy

### DIFF
--- a/docs/adr/0016-rds-read-replica.md
+++ b/docs/adr/0016-rds-read-replica.md
@@ -39,8 +39,8 @@ We could potentially promote a read replica to its own instance and run the scri
 
 However, that’s probably complicating things. We could just as easily have a terraform script that starts up a database from the latest snapshot, runs the script against it, and produces a new snapshot from which the other phases start their databases. That would also be safer, as a script would obviate the possibility of accidentally choosing the wrong database.
 
-##
-
+### Verdict
+Of the possible benefits, RDS Read Replicas only seem like a possibly worthwhile response for the second, "DB single source of failure" challenge. Whether it's worth it in the near term is not clear.
 
 ## Consequences
 
--- a/docs/adr/0016-rds-read-replica.md
+++ b/docs/adr/0016-rds-read-replica.md
@@ -1,10 +1,3 @@
-# #. Title
-
-This is the template in [Documenting architecture decisions - Michael Nygard](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions).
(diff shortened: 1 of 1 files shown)


jasalisbury:
I tend to agree read replicas aren't worth it for the first and third challenges. For the first, I can't see our data analysts adding enough load to our DB to really impact performance, at least right now. For the third, as you said we can further automate the DB script we already have making it safer to run.

I'd argue it also probably isn't worth it for the second as well, as you said we'd have to quickly act to prevent corruption spreading to the read replica. If the extra hosting costs are low enough and the extra engineering lift is also low (along with updates to ATO documentation, not sure how much there is there) I could see us opting to use a read replica. It would protect us if the main database went down for some reason but given the history of RDS that seems like a pretty low probability event. Even then the read replica really only allows us to stay up during an AWS event, it doesn't really help prevent data loss (baring the DB going down and AWS losing all our backups!).

hardwarehuman:
I tend to agree with @jasalisbury that the read replica _probably_ doesn't help us that much with stability events given that data events seem far more likely than a particular RDS instance going down.

One possible exception is if cloud.gov comes back to @kryswisnaskas with the information that they can't do point-in-time backups or more frequent backups to reduce the impact of any data errors. If that is the case (which seems unlikely but plausible to me), then we could make backups from the read replica without impacting prod.

thewatermethod:
Is there anything to getting this infrastructure in place as a future-proofing sort of thing? Meaning, is there a point where our need for analytics & data extraction grows, will we eventually want a read replica purely for performance reasons? 

My instinct as a dilettante is that the DB wouldn't be the bottleneck, but just curious.

jasalisbury:
> Is there anything to getting this infrastructure in place as a future-proofing sort of thing? Meaning, is there a point where our need for analytics & data extraction grows, will we eventually want a read replica purely for performance reasons?
> 
> My instinct as a dilettante is that the DB wouldn't be the bottleneck, but just curious.

There may be. Depending how cheap it is to have a read replica it might be worth it to have one sitting in a different availability zone just in case. That assumes we have ec2 instances running in different availability zones, which I'm not sure of (we should if we don't) because it wouldn't do much good to have an extra db in AZ 2 if an event in AZ 1 takes out all our ec2 instances anyways.

jasalisbury:
Are we ready to merge this in soon?

jasalisbury (approved):
:+1:

PR #606, opened 2021-11-02
Add topic/reasons graph widget

jasalisbury:
## Description of change

 * pull bar graph into it's own component
 * bar graph now scrolls just the graph content, allows title/labels to still be shown when the graph is scrolled
 * add topic/reason freq widget. Can switch between display of reasons or topics
 * added backend test helpers to quickly (as in how long to code) create activity reports and destroy them for tests, along with any needed associations

## How to test

Browse to the grantee tta record page

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-394

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


src/widgets/frequencyGraph.js
```diff
@@ -0,0 +1,49 @@
+import { Op } from 'sequelize';
+import {
+  ActivityReport,
+  sequelize,
+} from '../models';
+import { REPORT_STATUSES, TOPICS, REASONS } from '../constants';
+
+async function countOccurrences(scopes, column, possibilities) {
```
thewatermethod, line 8: In this PR https://github.com/adhocteam/Head-Start-TTADP/pull/480/ I started a widgets/helper.js file. I would think it makes sense to move this function in there, and also maybe we could standardize on this and the topicFrequencyGraph endpoint.

thewatermethod (commented):
Had several comments written when I realized I'll have to restart my computer in order to actually test it out. Looks great so far though!

thewatermethod:
I actually seem to be running into the same issue as previous, where using the Plot.ly react library crashes my docker container. I'll set aside some time to try and figure out if I can fix this

Screen Shot 2021-11-03 at 10 00 49 AM
s

src/widgets/frequencyGraph.js
jasalisbury: My loose plan is to replace the `topicFrequencyGraph` completely with this new widget. Need to raise the issue with patrice/angela when demoing this widget.

jasalisbury:
> I actually seem to be running into the same issue as previous, where using the Plot.ly react library crashes my docker container. I'll set aside some time to try and figure out if I can fix this on my setup
> 
> Screen Shot 2021-11-03 at 10 00 49 AM
> 
> I apologize, I should have documented having this issue at the time
> 
> * [ JavaScript heap out of memory error with the quickstart Plot example. plotly/react-plotly.js#135](https://github.com/plotly/react-plotly.js/issues/135)

Strange. Let me know if you can't get it to work, I haven't seen anything like that locally.

frontend/src/components/ButtonSelect.js
```diff
@@ -160,7 +161,7 @@ function ButtonSelect(props) {
   const ariaLabel = `${menuIsOpen ? 'press escape to close ' : 'Open '} ${ariaName}`;
 
   return (
-    <div className="margin-left-1" onBlur={onBlur} data-testid="data-sort">
+    <div className={`margin-left-1 ${className}`} onBlur={onBlur} data-testid="data-sort">
       <button
```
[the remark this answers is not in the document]
thewatermethod: Whoops **don't use a select**, not sure why I suggested that. We need the "apply" button in there. Sorry!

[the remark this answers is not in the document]
jasalisbury: yeah you are right, didn't think it all the way through!

frontend/src/widgets/FrequencyGraph.js
```diff
@@ -0,0 +1,131 @@
...
+
+  const onSelectedGraphChange = (e) => {
+    updateSelectedGraph(e.value);
+  };
+
+  return (
+    <Container className="ttahub--frequency-graph" padding={3} loading={loading} loadingLabel={`${selectedGraph} frequency loading`}>
+      <Grid row className="position-relative margin-bottom-2">
+        <Grid className="flex-align-self-center" desktop={{ col: 'auto' }} mobileLg={{ col: 10 }}>
+          <ButtonSelect
```
thewatermethod, line 61: I can select it with the keyboard! That's great.
The reader interaction is close, but I think just needs a tuneup

Right now if I proceed through the document, when I get to the top of this widget, I am read this:
- Open change graph menu, button
- Heading, level 2 - In activity reports by frequency

I think there are two easy solutions
- Expand the button description to include the "activity reports by frequency" bit and then hide the h2 via aria-hidden=true.
- Add some visually hidden text to the heading to explain the full context

I swear I **can** do a screenreader demo if that would be helpful

frontend/src/widgets/FrequencyGraph.css
```diff
@@ -0,0 +1,52 @@
...
+}
+
+@media(min-width: 1090px){
+  .ttahub--show-accessible-data-button {
+    flex: 1;
+    text-align: right;
+  }
+}
+
+.ttahub--frequency-graph-tooltip {
```
thewatermethod, line 12: Is this something I left in there? We are using the out of the box tooltips now

frontend/src/widgets/BarGraph.js
```diff
@@ -0,0 +1,138 @@
...
+
+  return (
+    <>
+      <div className="display-flex flex-align-center">
+        <div className="margin-right-1 smart-hub--vertical-text">
+          { yAxisLabel }
+        </div>
+        {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
+        <div className="overflow-x-scroll" tabIndex={0}>
+          <caption className="sr-only">graph receives focus so keyboard users can scroll it</caption>
```
thewatermethod, line 108: This is a really smart addition. Wish I had thought of it!
I think that it might make sense to read off something like "use the arrows keys to scroll" instead of "graph receives focus...", etc

frontend/src/widgets/FrequencyGraph.js
jasalisbury: Yeah we might need to pair for that, the system screen reader I have for linux doesn't seem to function properly (I'm pretty sure it doesn't read out the h2 at all).

frontend/src/widgets/BarGraph.js
jasalisbury: Don't give the credit to me, give it to stack overflow haha

thewatermethod:
I think this is less of a point about this code and more about plot.ly in general but these bundle sizes are kind of absurd.

Before react-plot.ly (17.18 mb)
before

After react-plot.ly (26.42 mb)
after

If we are going to need to support low bandwidth users, we may need to find a new library.

jasalisbury:
> I think this is less of a point about this code and more about plot.ly in general but these bundle sizes are kind of absurd.
> 
> Before react-plot.ly (17.18 mb) before
> 
> After react-plot.ly (26.42 mb) after
> 
> If we are going to need to support low bandwidth users, we may need to find a new library.

Might be time to split up application code from our dependencies, so users only need to download the bigger dependencies bundle when dependencies are updated (which is infrequently). Hopefully `react-create-app` is easy to configure to do that.

frontend/src/widgets/BarGraph.js
GarrettEHill: I like the instructional caption more then the current one.

GarrettEHill:
> > I think this is less of a point about this code and more about plot.ly in general but these bundle sizes are kind of absurd.
> > Before react-plot.ly (17.18 mb) before
> > After react-plot.ly (26.42 mb) after
> > If we are going to need to support low bandwidth users, we may need to find a new library.
> 
> Might be time to split up application code from our dependencies, so users only need to download the bigger dependencies bundle when dependencies are updated (which is infrequently). Hopefully `react-create-app` is easy to configure to do that.

GarrettEHill:
Not sure if this should be specific to this review, but are we doing any compaction or compression on the js files are all data from the server to reduce bandwidth/download size?

jasalisbury:
> Not sure if this should be specific to this review, but are we doing any compaction or compression on the js files are all data from the server to reduce bandwidth/download size?
> 
> Would incorporating something like https://hub.docker.com/r/tdewolff/minify be a good idea?

Network tab from the prod site (without caching)

Screenshot from 2021-11-08 09-52-39

From `react-create-app` https://create-react-app.dev/docs/production-build

Looks like `react-create-app` splits out dependencies for us automagically. It also minifies our code for us. Most of the time users are going to be downloading the `main.*` files as that is where our code/css lives.

thewatermethod (approved):
Thanks for taking the time to work on the accessibility! 

Re: the bundle size issue - seems like 99% of our users would not experience much of a downside for having a giant bundle. (I would still think that low-bandwidth users might need a more engineered solution, but that's an issue for another day!)

jasalisbury:
Seems like we have a setting in github where a commit after a review resets the review

thewatermethod (approved):
🥇

PR #610, opened 2021-11-05
Move report table to own component

jasalisbury:
## Description of change

 * Report table is now it's own component
 * Activity reports table replaced with new stand alone component
 * My alerts still needs to be updated to use new component. There are a few difference between the two tables that need to be handled. Saving for a later date to keep the change smaller
 * `withWidgetData` uses deep compare to prevent multiple renders
 * Moved table related tests from landing page to new component
 * Add report table to grantee dashboard

Only added functionality is the ability to show or hide the filter menu on report tables, and dynamically setting the table caption.

## How to test

View activity reports table on landing page. Check the following:
 * filters, changing regions
 * filters are preserved through region changes and regions through filter changes
 * correct results are displayed when filtering
 * download buttons still function
 * you can select individual and all reports
 * sorting still works
 * paging still works

View reports on the grantee TTA record page. Check the same as above. There won't be filter controls and the reports should be scoped to the current grantee.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-396

## Notes

This PR will wait for global filters to show up on the grantee TTA page before merging.

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


frontend/src/components/ActivityReportsTable/ReportRow.js
```diff
@@ -0,0 +1,144 @@
...
+    id,
+    displayId,
+    activityRecipients,
+    startDate,
+    author,
+    topics,
+    collaborators,
+    lastSaved,
+    calculatedStatus,
+    legacyId,
```
GarrettEHill, line 25: Would it be better to make the API for generating the report more generic, and an instance of the API have the required specifics?

frontend/src/components/ActivityReportsTable/index.js
```diff
@@ -0,0 +1,339 @@
...
+          aria-label={`${displayName}. Activate to sort ${sortClassName === 'asc' ? 'descending' : 'ascending'
+          }`}
+        >
+          {displayName}
+        </a>
+      </th>
+    );
+  };
+
+  const displayReports = reports.length ? reports : [emptyReport];
```
thewatermethod, line 206: I'm probably not understanding something - do we want an row with an empty report rather than an empty array?

thewatermethod (dismissed):
👍 Looks good

I did notice while testing it out a bit that those ariaLiveContext.annouce functions weren't reading anything off to me, but that is a pre-existing condition, nothing to do with this.

frontend/src/components/ActivityReportsTable/ReportRow.js
jasalisbury: Yeah it would be nice to have some way of defining which fields are required for the table be passed in as props. We will need _something_ like that when we update the "My Alerts" table to use this component. However in this PR I'm trying to keep changes to a minimum, only want to move the table to a new component, not make any substantial changes.

frontend/src/components/ActivityReportsTable/index.js
jasalisbury: I'm not too sure, I was trying to move over the table to it's own component without changing how the report works in any way. Now that it is in it's own component it is easier to refactor.

thewatermethod (approved):
Still looks good!

PR #615, opened 2021-11-09
Fix plotly import

jasalisbury:
## Description of change

Plotly was causing OOMs when building.

## How to test

Build passes and deploys to dev


thewatermethod (approved):
Very cool and normal library

PR #621, opened 2021-11-12
[TTAHUB-412] More flexible date picker behavior

thewatermethod:
## Description of change

On the date range selector component, you can now choose any start date after 9/1/2020. If you choose a start date after your chosen end date, the end date will adjust and the delta between them is preserved.

I did not change the filters on the Activity Report table as its a separate component and we will soon be retiring it.

## How to test
Go to the date range selector on the Dashboard or the Grantee TTA Record. Choose a start date. Choose an end date. Change your start date to be after your end date. Note that the end date is changed and the delta between them is the same.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-412

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


jasalisbury:
If we move dates around for the start date do we want to do the same for the end date? So if you select an end date before the start date the delta is preserved?

frontend/src/components/DateRangeSelect.js
```diff
@@ -256,17 +260,32 @@ function DateRangeSelect(props) {
...
                         setStartDateFocused(focused);
                       }
                     }}
+                    isOutsideRange={(day) => isOutsideRange(day, 'start')}
                     onDateChange={(selectedDate) => {
-                      setStartDate(selectedDate);
+                      // weird that we'd have to explicitly do this
+                      // but otherwise things get all wacky when you
+                      // type in a date that's different than the end date
+                      if (!selectedDate) {
```
jasalisbury, line 273: Just a guess, `onDateChange` is called for every character that is typed and selectedDate is null when the data is invalid?

jasalisbury (approved):
Looks good to me pending the question on if we want end date to behave similarly (but in reverse!)

frontend/src/components/DateRangeSelect.js
thewatermethod: you are correct sir

thewatermethod:
> If we move dates around for the start date do we want to do the same for the end date? So if you select an end date before the start date the delta is preserved?

For some reason, this isn't as intuitive to me, but I also can't think of a good reason not to - I will bring it up at office hours.

thewatermethod:
@jasalisbury not sure if you wanted to re-validate this, I had to change a few more files since I changed the wrong thing

jasalisbury (approved):

PR #626, opened 2021-11-17
Make overview & activity reports table queries match

thewatermethod:
## Description of change

The overview widget and activity reports table would sometimes be off by one on particular regions. The culprit seems to have been a greater than or equal to 8/31/2020 on one and a greater than to 8/31/2020 on the other.

## How to test

On /activity-reports, visit the page with an account that has access to many regions. View each region individually. Note the total count on the table and the total count on the widgets are the same across all regions.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-442

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


src/services/activityReports.js
```diff
@@ -311,12 +310,10 @@ export function activityReports(
 ) {
   const scopes = filtersToScopes(filters);
 
-  const endDte = moment().format('MM/DD/yyyy');
-
   const where = {
     calculatedStatus: REPORT_STATUSES.APPROVED,
     [Op.and]: scopes,
-    startDate: { [Op.gte]: '2020-08-31', [Op.lte]: endDte },
+    startDate: { [Op.gte]: '2020-09-01' },
```
AdamAdHocTeam, line 316: I think this change was added as part of TTAHUB-321 see [files changed](https://github.com/adhocteam/Head-Start-TTADP/pull/454/files).

thewatermethod: looks like in those files changed you linked it was ```[Op.gte]: 2020-09-01``` as well, so it must have been tweaked to ```2020-08-31``` in the meantime. 

we don't need to specify not showing start dates in the future, right? since we are only fetching approved reports?

jasalisbury:
I think I mentioned this in an earlier PR but there are 4 legacy reports with a start date before September 1st 2020. I would suggest we verify the legacy reports are accessible in the previous system and delete them in our system. If they are not available in the external system (I assume HSES) we see if we can insert them there and delete them in our system

If we keep these reports in our system we will always have to remember to throw a start date condition on any query against the activity reports table.

The legacy report ids
```
"R01-AR-000002"
"R01-AR-000027"
"R01-AR-000001"
"R01-AR-000016"
```

thewatermethod:
@jasalisbury do you think that should be done instead of this code change? if we do that, we could remove the start date condition from both places

jasalisbury:
> @jasalisbury do you think that should be done instead of this code change? if we do that, we could remove the start date condition from both places

Yeah if we can delete the reports that seems like the way to go to me

jasalisbury:
See https://ocio-jira.acf.hhs.gov/browse/TTAHUB-468

GarrettEHill:
Is this still in progress or abandoned?

thewatermethod:
> Is this still in progress or abandoned?

Blocked by https://ocio-jira.acf.hhs.gov/browse/TTAHUB-468

jasalisbury (approved):
:tada:

PR #639, opened 2021-11-22
[TTAHUB-417] Keyboard user improvements on Activity Reports table

thewatermethod:
## Description of change
Previously, if a keyboard user wanted to export selected reports, it would require **a lot** of tabbing if one of those reports were in the middle of the table.

Hidden tab stops were created directly after the selection checkbox. These are "aria-hidden" since a bunch of buttons with identical text would be confusing to screen-reader-only users, and they have additional navigational options.

## How to test
Select a report and tab through the document. Observe that you can tab to focus one of the hidden buttons (directly after the report checkbox) and hit enter to initiate a download of selected reports.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-417

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -127,10 +127,10 @@
   },
   "devDependencies": {
     "@sheerun/mutationobserver-shim": "^0.3.3",
-    "@testing-library/dom": "^7.28.0",
+    "@testing-library/dom": "^8.11.1",
     "@testing-library/jest-dom": "^5.11.9",
     "@testing-library/react": "^11.2.5",
-    "@testing-library/user-event": "^12.2.2",
+    "@testing-library/user-event": "^13.5.0",
     "cross-env": "^7.0.2",
     "eslint": "^7.20.0",
     "eslint-config-airbnb": "^18.2.0",
--- a/frontend/src/components/ActivityReportsTable/ReportRow.css
+++ b/frontend/src/components/ActivityReportsTable/ReportRow.css
@@ -1,0 +1,22 @@
+.tta-smarthub--report-row td:first-child {
+    position: relative;
+}
+
+.tta-smarthub--report-row .ttahub-export-reports:not(:focus) {
+	border: 0;
+	clip: rect(1px, 1px, 1px, 1px);
+	clip-path: inset(50%);
+	height: 1px;
+	margin: -1px;
+	overflow: hidden;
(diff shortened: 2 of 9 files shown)


jasalisbury (commented):
Looks and works well! Have one question...

> Hidden tab stops were created directly after the selection checkbox. These are "aria-hidden" since a bunch of buttons with identical text would be confusing to screen-reader-only users, and they have additional navigational options.

Seems like having aria-hidden on a focus-able element might be an accessibility issue

> According to the fourth rule of ARIA, aria-hidden="true" should not be used on a focusable element. Additionally, since this attribute is inherited by an element's children, it should not be added onto the parent or ancestor of a focusable element. 

(from https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute)

If we want to go this route we will probably need to document why we think we can ignore the rule.

thewatermethod:
Yeah, I'd be glad to (and definitely should) reach out to a specialist to confirm my thinking on that before we move forward. I could very well be wrong/there may be a better way. 

I just know that a common screen-reader usage pattern is using a shortcuts to list off all the buttons on a page (same for links, headings, etc). In that case, I would think that we wouldn't want 12 buttons that do the same thing with the same name being read off? But I wouldn't want to violate a core behavior either. Tough one.

> Adding aria-hidden="true" to an element removes that element and all of its children from the accessibility tree. This can improve the experience for assistive technology users by hiding ...duplicated content, such as repeated text

thewatermethod:
Figured out an alternate way to that same solution that didn't involve aria-hidden.

As a sidenote, I think I might start a jira just to list off all the things we can change or update when we finally get to abandon IE11.

jasalisbury:
> Figured out an alternate way to that same solution that didn't involve aria-hidden.

Cool I'll take a look!

> As a sidenote, I think I might start a jira just to list off all the things we can change or update when we finally get to abandon IE11.

Yeah we need to bring up IE11 end of life at our next standup, because it is coming up next year (thankfully!). We might get lucky and be able to drop support sooner. If I forget try to remember to add a sticky for it!

jasalisbury (approved):
:tada:

PR #645, opened 2021-11-23
css fixes for grantee record

thewatermethod:
## Description of change
There are currently 4 different filters variations on the Recipient TTA record page
- Specialist filters
- Date range within
- Date before or after
- Either topic or condition not selected
In all those scenarios, the select dropdowns should align vertically.

This PR also modifies the activity report menu buttons so that they aren't outliers style-wise (worked with Lauren to get those styles).

## How to test

View the UI and check that all the above variations do align in the filters panel.
Make sure the regional dashboard controls still look correct, and the activity report table menu buttons also still work.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-489

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/DropdownMenu.js
+++ b/frontend/src/components/DropdownMenu.js
@@ -83,7 +83,7 @@ export default function DropdownMenu({
     );
   }
   return (
-    <div role="menu" ref={forwardedRef} tabIndex="0" aria-label={menuName} className={classNames} onBlur={onBlur} onKeyDown={onKeyDown}>
+    <div role="menu" ref={forwardedRef} tabIndex="-1" aria-label={menuName} className={classNames} onBlur={onBlur} onKeyDown={onKeyDown}>
       <button
         onClick={onClick}
         className={`${buttonClasses} smart-hub--dropdown-menu-toggle-btn display-flex margin-0`}
--- a/frontend/src/components/DropdownMenu.js
+++ b/frontend/src/components/DropdownMenu.js
@@ -46,7 +46,7 @@ export default function DropdownMenu({
     }
 
     if (canBlur(e)) {
-      // setMenuIsOpen(false);
+      setMenuIsOpen(false);
     }
   };
 
--- a/frontend/src/components/DropdownMenu.js
+++ b/frontend/src/components/DropdownMenu.js
@@ -83,7 +83,7 @@ export default function DropdownMenu({
     );
   }
   return (
-    <div role="menu" ref={forwardedRef} tabIndex="-1" aria-label={menuName} className={classNames} onBlur={onBlur} onKeyDown={onKeyDown}>
+    <div role="menu" ref={forwardedRef} tabIndex="0" aria-label={menuName} className={classNames} onBlur={onBlur} onKeyDown={onKeyDown}>
       <button
         onClick={onClick}
         className={`${buttonClasses} smart-hub--dropdown-menu-toggle-btn display-flex margin-0`}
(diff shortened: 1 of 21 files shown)


kryswisnaskas (commented):
Looks good! 👍 

The filter dropdowns align vertically now 🎉 

One small observation - could we change the border color for `Date `under the `is before` and `is after` conditions to match the other borders? (the `within `condition is already matching)

image

jasalisbury:
Ran into a bug
Screenshot from 2021-11-29 17-07-48
To reproduce:

 1) Go to the TTA History tab
 2) Open the filters, change the date filter to specialists
 3) Add two roles (I used ECS and FES, doubt it matters)
 4) Click apply

jasalisbury:
Another more general thought, having the custom date selectors below the selection of `Year to Date` vs `Custom Date Range` really pushes the date selectors pretty far down the page. It is especially bad when the end date calendar is open. I wonder if we can move the start/end date selector to the right of the year/custom selector. Something to potentially bring up at the next UI review.

jasalisbury:
I'm seeing quite a few proptype errors when I load the page. Also `pregnant woman` shows up twice as a target population even for grantees with no activity reports (may be a bigger issue that requires its own ticket).

frontend/src/components/DropdownMenu.js
```diff
@@ -83,7 +83,7 @@ export default function DropdownMenu({
     );
   }
   return (
-    <div role="menu" ref={forwardedRef} tabIndex="-1" aria-label={menuName} className={classNames} onBlur={onBlur} onKeyDown={onKeyDown}>
+    <div role="menu" ref={forwardedRef} tabIndex="0" aria-label={menuName} className={classNames} onBlur={onBlur} onKeyDown={onKeyDown}>
```
jasalisbury, line 86: When I tab through the page I stay on the `filters` dropdown for two presses of my tab key. The first time when I press enter nothing happens. The next element that is focused on tab opens the dropdown menu when I press enter. Took a screenshot of the first element in focus
Screenshot from 2021-11-29 17-22-38
and the second
Screenshot from 2021-11-29 17-22-43

jasalisbury (commented):
:+1: good work getting everything more responsive to screen size

jasalisbury:
commenting here so it isn't lost... We probably need to add a max width to the profile page that matches the TTA History page.

frontend/src/components/DropdownMenu.js
thewatermethod: Really scratching my head as to why I changed this

thewatermethod:
> Ran into a bug Screenshot from 2021-11-29 17-07-48 To reproduce:
> 
>     1. Go to the TTA History tab
> 
>     2. Open the filters, change the date filter to specialists
> 
>     3. Add two roles (I used ECS and FES, doubt it matters)
> 
>     4. Click apply

Could you double check that? It seems like you may have not hit apply on the specialist select. If that the case, both you and Krys had the same issue, and that seems like a sign that something is amiss in the UI. Perhaps related to the apply button being hidden under the fold/your other comment about the length of the dropdowns, not sure. 

I changed the pill so that it says "None selected" if you haven't applied or you have none selected. Maybe that will alleviate the issue somewhat.

> commenting here so it isn't lost... We probably need to add a max width to the profile page that matches the TTA History page.

Added the max width to the profile

> I'm seeing quite a few proptype errors when I load the page. Also pregnant woman shows up twice as a target population even for grantees with no activity reports (may be a bigger issue that requires its own ticket).

It was in the constants.js twice. Removing it also resolved pretty much all of the prop type errors.

jasalisbury (approved):

PR #657, opened 2021-11-30
Activity reports table component uses 'useDeepCompareEffect'

jasalisbury:
## Description of change

Props passed into the activity reports table sometimes are new objects. Use deep compare use effect to prevent multiple API calls on props that have not changed.

## How to test

Open the TTA History page on main with your network tab open. Note multiple API calls to `activity-reports`. Now reload the page with this branch, not only a single API call.


thewatermethod (approved):
🦘

PR #659, opened 2021-11-30
TTAHUB-498: Remove Program Type from AR Entry

AdamAdHocTeam:
## Description of change

Remove Program Type from AR Entry

## How to test

Program Type should no longer appear on the AR form. This includes all relevant views and export files.

Exporting the CSV should still populate 'Program Specialists' column and the newly added 'Program Type' column (both new line separated).

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-498

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


thewatermethod:
Are we breaking this jira into two parts? I don't see the changes to the CSV or the Grantee selection field in this PR

AdamAdHocTeam:
> Are we breaking this jira into two parts? I don't see the changes to the CSV or the Grantee selection field in this PR

The CSV change was added just recently. I will make this change now.

src/migrations/20210106160931-create-activity-reports.js
```diff
@@ -47,10 +47,6 @@ module.exports = {
         allowNull: false,
         type: Sequelize.STRING,
       },
-      programTypes: {
```
thewatermethod, line 50: Wouldn't best practice be to write a new migration rather than edit this one?

src/lib/transform.js
```diff
@@ -62,14 +62,31 @@ function transformRelatedModel(field, prop) {
   return transformer;
 }
 
-function transformGrantModel(field) {
+function transformGrantModel(prop, secondaryProp) {
```
thewatermethod, line 65: Maybe it doesn't have to be this exact PR, but I can't help thinking there is more abstracted way to write this that could reuse ```transformRelatedModel``` - like a simpler function that just passes the activity recipients down into it (so instance[field] gets the grants, etc).

src/migrations/20210106160931-create-activity-reports.js
AdamAdHocTeam: I don't think we are removing the database column. I almost wonder if I shouldn't change the migrations at all.

thewatermethod: yeah thats a good point. so would it have to stay here and where its changed on the model?

src/lib/transform.js
AdamAdHocTeam: Spoke with Josh we will break this into two functions so its easier to read and understand later.

src/migrations/20210106160931-create-activity-reports.js
jasalisbury: Yeah once a migration hits main it needs to not change. Any modifications to the schema need to happen with a new migration.

src/models/grant.js
```diff
@@ -51,7 +51,8 @@ module.exports = (sequelize, DataTypes) => {
     name: {
       type: DataTypes.VIRTUAL,
       get() {
-        return `${this.grantee.name} - ${this.number}`;
+        const grantName = this.programs && this.programs.length > 0 ? ` - ${this.programs.map((p) => (p.programType)).sort().join(', ')}` : '';
```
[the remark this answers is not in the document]
AdamAdHocTeam: This is a great idea! :-) I have cleaned up the code to use the new virtual property.

src/lib/transform.js
```diff
@@ -194,6 +199,7 @@ const arTransformers = [
   'endDate',
   'startDate',
   transformRelatedModel('activityRecipients', 'name'),
+  transformGrantModel('programTypes'),
```
jasalisbury, line 202: pretty minor but can we keep the columns in the same order? I'd hate for someone to have formulas setup in excel that depend on the order of the columns. Actually I'm not even sure how the header is set, so these may need to stay in the same order so values match the header.

src/lib/transform.js
```diff
@@ -62,14 +62,20 @@ function transformRelatedModel(field, prop) {
...
+    const values = await instance.activityRecipients;
+    if (values) {
+      const distinctList = [
+        ...new Set(
+          values.filter(
+            (recipient) => recipient.grant && recipient.grant[prop] !== null,
+          ).map((r) => r.grant[prop]).flat(),
+        ),
+      ];
+      const programSpecialistNames = distinctList.sort().join('\n');
```
jasalisbury, line 77: Also minor, but since this method is now generic I think we will want a different name for `programSpecialistNames`

AdamAdHocTeam: I have renamed the vars.

jasalisbury (approved):

src/models/activityReport.js
```diff
@@ -142,9 +142,6 @@ export default (sequelize, DataTypes) => {
     requester: {
       type: DataTypes.STRING,
     },
-    programTypes: {
-      type: DataTypes.ARRAY(DataTypes.STRING),
-    },
```
thewatermethod, line 147: Is there an issue with this being removed here without changing the DB?

thewatermethod (approved):
Looks good!

src/models/activityReport.js
AdamAdHocTeam: I don't think so as all BE tests pass and I can create and retrieve a report on the site without issue.

PR #664, opened 2021-12-02
[TTAHUB-359] add filters to regional dashboard

thewatermethod:
## Description of change
Add filters to the regional dashboard page
- Add new filter options 
- Condense configuration for Filter Menu/Filter pills
- Make filters in filter menu configurable
- Filters are reflected in the URLs
- Mark up improvements

I am creating the state filter, originally part of this Jira, separately as it requires its own logic.

## How to test
Add filters to the dashboard. Check that all the behaviors are as expected. Check that the data is updated properly.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-359

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


jasalisbury:
Should this be merging into main?

thewatermethod:
> Should this be merging into main?

Sure - seems like its less files that way! Some of them are the same changes as PR #666, be warned

jasalisbury:
> > Should this be merging into main?
> 
> Sure - seems like its less files that way! Some of them are the same changes as PR #666, be warned

Can we push through #666 first? If there is still a long discussion on what the names are going to be can we just change those names in a different PR? I'm just seeing a lot of changes I've already seen and it is hard to determine what is new and what is old.

thewatermethod:
> Can we push through #666 first? If there is still a long discussion on what the names are going to be can we just change those names in a different PR? I'm just seeing a lot of changes I've already seen and it is hard to determine what is new and what is old.

It is the name discussion holding things up. I will see if I can get some closure at stand up.

frontend/src/components/filter/FilterMenu.js
```diff
@@ -161,20 +178,31 @@ export default function FilterMenu({ filters, onApplyFilters }) {
     setItems(newItems);
   };
 
+  const clearAllFilters = () => {
+    // this looks a little strange, right?
+    // well, we don't want to clear out things like the region, just the filters that can be set
+    // in the UI
+    const newItems = items.filter((item) => prohibitedFilters.includes(item.topic));
```
jasalisbury, line 178: If this component is only passed filters that can be changed in the UI (in the grantee page the `granteeId` filter isn't passed to this component) does it need to worry about not trampling "static" filters? If I'm mistaken can we introduce some concept of "static" filters that the `filter` components don't even see. So this component wouldn't be able to clear out the region, because it isn't included in the filters this components has access to via the filters prop.

frontend/src/pages/GranteeRecord/pages/TTAHistory.js
```diff
@@ -10,44 +10,32 @@ import Overview from '../../../widgets/DashboardOverview';
...
 }) {
-  const [filters, setFilters] = useState([
+  const [filters, setFilters] = useUrlFilters([
     {
       id: uuidv4(),
       topic: 'startDate',
       condition: 'Is within',
       query: defaultDate,
     },
+    {
```
jasalisbury, line 33: Should region be included in the filters ~line 46 since it can't be changed in the UI?

frontend/src/pages/RegionalDashboard/index.js
```diff
@@ -39,113 +41,97 @@ export default function RegionalDashboard({ user }) {
...
       id: uuidv4(),
       topic: 'startDate',
       condition: 'Is within',
-      query: dateRange,
+      query: defaultDate,
     },
-  ];
+  ]);
+
+  const startDateFilter = filters.find((filter) => filter.topic === 'startDate');
```
jasalisbury, line 61: Do we still need to show the date range on filters in the regional TTA page? Or do we need to start on the other pages? We don't do it on any other page. Seems kinda of strange to show only a single filter condition stated on the widget when we add the ability to filter on so may other things.

frontend/src/pages/GranteeRecord/index.js
```diff
@@ -68,15 +65,6 @@ export default function GranteeRecord({ match, location }) {
 
   return (
     <>
-      <Helmet>
```
jasalisbury, line 71: Whenever I load the profile page from the search page I see the title `Grantee Profile - - TTA Hub` If I switch the tab to `TTA History` and back to `Profile` I see `Grantee Profile - ${Grantee name} - TTA Hub`

frontend/src/pages/RegionalDashboard/index.js
thewatermethod: Good point, I'll ask about this at the UI meeting

frontend/src/components/filter/FilterMenu.js
thewatermethod: Right now we have filters that
- are in the URL and are editable in the UI (most common)
- are not in the URL and can't be edited (Grantee ID in the recipient record)
- are in the URL and can't be edited (region in the Recipient Record)

This is the case I'm handling above. The filters in the parent component appear in the URL and in the component below. My solution was to create a prop (probably poorly named) to hold such filters. The other option I thought of would have been to maintain three arrays of filters instead of two. I think if you have a way to make it simpler (or an idea for me to chew on) I'd love to hear it

frontend/src/pages/GranteeRecord/pages/TTAHistory.js
thewatermethod: Related to my comment above, right now I've got filters that are applied to the widgets, which are different then the filters that are applied to the URL, which are different then are what's editable in the UI. It's a mess. Spending some time now pondering.

frontend/src/components/filter/FilterMenu.js
jasalisbury: > are in the URL and are editable in the UI (most common)
are not in the URL and can't be edited (Grantee ID in the recipient record)
are in the URL and can't be edited (region in the Recipient Record)

Thinking out loud, but of these three it seems to me only the first should be passed to this component. `granteeId` should be able to be added into the parent of the `FilterMenu` whenever needed (when the query string is being built to send off to the API). I think `regionId` can be handled the same way as `granteeId`, in the parent component right before the request is sent off.

I think the last point may bring up an issue with how we are structuring URLs on the frontend. I think the case could be made that having the region as a query param isn't appropriate and it should be a part of the path. Especially since it is required to load the resource. It would make sense to me to give the user a 404 if they try to view the recipient tta record page with an invalid `regionId`, one that the recipient has no grants in.

_Might_ be a good rule to have, only change-able filters show up in the filter menu/pills and only change-able filters show up as url params.

thewatermethod: >  think the last point may bring up an issue with how we are structuring URLs on the frontend. I think the case could be made that having the region as a query param isn't appropriate and it should be a part of the path. Especially since it is required to load the resource. It would make sense to me to give the user a 404 if they try to view the recipient tta record page with an invalid regionId, one that the recipient has no grants in.

>Might be a good rule to have, only change-able filters show up in the filter menu/pills and only change-able filters show up as url params.

That makes sense. We're devoting a lot of energy in the code to partitioning things off in this manner. A lot of props and filtering would go away if we just did as you said - all filters that are query params get passed into the filter menu, and all filters in the menu can be changed. 

Everything else needs to be in the path or not in the URL at all and can't be changed.

jasalisbury (commented):
Starting to feel close! Some things I noticed just poking around

 * Date 'is within' last 30 days blur behavior feels strange. Should we close them menu on blur?
 * Space needed between Region and the rest of the title
Screenshot from 2021-12-17 09-58-20
 * (recipient tta record) Slashes in url param need to be encoded
 * Mentioned this to Adam as well, but multi-selects are not consistent, some use 'AND' and some 'OR'. Try selecting every topic in the regional dashboard. You'll get zero results. Try selecting every specialists role and you'll get all reports (or at least close to all)

thewatermethod:
> Mentioned this to Adam as well, but multi-selects are not consistent, some use 'AND' and some 'OR'. Try selecting every topic in the regional dashboard. You'll get zero results. Try selecting every specialists role and you'll get all reports (or at least close to all)

Talked about this with Adam this morning as well. It looks like one of the scope methods (filterArray) uses OR for excludes and AND for includes, whereas filterAssociation is the opposite. I guess maybe they should "OR" in all cases?

jasalisbury:
> > Mentioned this to Adam as well, but multi-selects are not consistent, some use 'AND' and some 'OR'. Try selecting every topic in the regional dashboard. You'll get zero results. Try selecting every specialists role and you'll get all reports (or at least close to all)
> 
> Talked about this with Adam this morning as well. It looks like one of the scope methods (filterArray) uses OR for excludes and AND for includes, whereas filterAssociation is the opposite. I guess maybe they should "OR" in all cases?

That would be how I lean. It wouldn't make sense to me to have a multi-select where all selected items are AND'd together (but I'd double check with the whole group). These are the first multi-select filters so this is the first time we are running into this issue.

thewatermethod:
>Date 'is within' last 30 days blur behavior feels strange. Should we close them menu on blur?

There's [another jira](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-506) that I'd really like to save working on the date range picker for, since this should be a non-issue either way afterwards

frontend/src/hooks/useUrlFilters.js
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState } from 'react';
```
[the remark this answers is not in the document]
thewatermethod: They should be coming in on page load, hopefully. (They are for me, although they might have been broken when you originally wrote this comment)

AdamAdHocTeam (commented):
We probably want to show the region filter pill if the user has access to more than one region and there is a filter for a single region applied.

src/services/accessValidation.js
```diff
@@ -76,17 +87,6 @@ export async function setReadRegions(query, userId, useFirstReadRegion = false)
 
   // if region.in is part of query (user has requested specific regions)
   if ('region.in' in query && Array.isArray(query['region.in']) && query['region.in'][0]) {
-    // first check to see if "all regions (central office)" is selected
-    // if so, return all regions has access to
-
-    if (query['region.in'].length === 1 && parseInt(query['region.in'][0], DECIMAL_BASE) === 14) {
```
jasalisbury, line 82: This change causes central office users to see no reports/data on the landing page.

frontend/src/components/filter/FilterMenu.js
```diff
@@ -126,8 +125,7 @@ export default function FilterMenu({ filters, onApplyFilters, applyButtonAria })
 
   const onUpdateFilter = (id, name, value) => {
     //
-    // this is bonkers... we need to do more than map the array
-    // what escaped me originally was that just an array spread creates a new
+    // just an array spread creates a new
```
jasalisbury, line 128: Do we want to keep some of the comment above?

src/routes/widgets/handlers.js
```diff
@@ -22,7 +22,7 @@ export async function getWidget(req, res) {
     }
 
     // This returns the query object with "region" property filtered by user permissions
-    const query = await setReadRegions(req.query, req.session.userId, true);
+    const query = await setReadRegions(req.query, req.session.userId, false);
```
jasalisbury, line 25: I ran into an issue with this param looking at Adam's PR adding filters to the landing page. I think we can remove the param completely after both PRs make it up and just always assume `false`

frontend/src/components/filter/FilterMenu.js
thewatermethod: I'll remove it

src/services/accessValidation.js
thewatermethod: I'm confused by this - I'm looking at the site as a Central office user and am seeing plenty of data. Is it that central office users do not have permissions to any regions on the live app?

jasalisbury: Region in query params is `region.in[]=14`. The user is central office with permissions on region 1 and 3.

frontend/src/components/filter/FilterMenu.js
```diff
@@ -6,22 +6,31 @@ import PropTypes from 'prop-types';
...
 /**
  * Renders the entire filter menu and contains the logic for toggling it's visibility
  * @param {Object} props
  * @returns JSX Object
  */
-export default function FilterMenu({ filters, onApplyFilters }) {
+export default function FilterMenu({
+  filters, onApplyFilters, allowedFilters, dateRangeOptions,
+}) {
   const [items, setItems] = useState([...filters.map((filter) => ({ ...filter }))]);
```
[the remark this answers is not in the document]
thewatermethod: whoops sorry, missed this somehow - pretty sure that useState will create a new array but the contents will be refs not values, since they are objects, where we want fresh objects of the same value in this situation

src/services/accessValidation.js
thewatermethod: There won't be a region.in[]=14 in the query any more, as far as I know.

The way it's coded now, the user above should see data for 1 & 3 unless they've filtered by region.
Is this correct? Should they instead see all region data?

jasalisbury: I still get `regionId[]=14` locally on this branch on the landing page

thewatermethod: ah gotcha, the **landing page**, my bad

src/routes/widgets/handlers.js
jasalisbury: Actually I think we'll need to change it here, I only get a single region in the bottom table when I'm a central user. https://github.com/HHS/Head-Start-TTADP/blob/main/src/routes/activityReports/handlers.js#L511

frontend/src/components/filter/constants.js
```diff
@@ -0,0 +1,226 @@
...
+const handleArrayQuery = (q) => {
+  if (q.length) {
+    return q.join(', ');
+  }
+  return '';
+};
+
+const handleStringQuery = (q) => q;
+
+export const START_DATE_FILTER = {
```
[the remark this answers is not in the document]
jasalisbury: In the meantime can we hardcode the `dateRangeOptions` in the `startDate` filter so we can remove `dateRangeOptions` as a prop?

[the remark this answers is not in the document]
jasalisbury: Honestly I think I like the way you had it before better I think (before we had our chat on zoom today), specifically `FilterMenu` seems more complicated. Maybe we can revert it back to what you had before and revisit this in a refactor down the road? What do you think?

frontend/src/components/filter/FilterPopulationSelect.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import FilterSelect from './FilterSelect';
+import { TARGET_POPULATIONS } from '../../Constants';
+
+const POPULATION_OPTIONS = TARGET_POPULATIONS.filter((population) => population !== '--------------------').map((label, value) => ({ value, label }));
+
+const PREGNANT_WOMEN_OPTION = POPULATION_OPTIONS[2];
+// add border to the bottom of this option
+Object.assign(PREGNANT_WOMEN_OPTION, { endGroup: true });
```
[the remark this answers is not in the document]
thewatermethod: react select lets you do groups I'm pretty sure, so we can convert this to that, but maybe in a future PR

frontend/src/components/filter/constants.js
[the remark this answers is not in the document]
thewatermethod: > In the meantime can we hardcode the dateRangeOptions in the startDate filter so we can remove dateRangeOptions as a prop?

Not until the new PR - right now different pages use different date range options

jasalisbury (approved):

jasalisbury:
One thing I just noticed, we have `program types` and `topics` as plurals but other multi-selects are singular, like `target population` and `reason`

PR #666, opened 2021-12-02
Recipient record filter validation, UI fixes

thewatermethod:
## Description of change
- Filters should validate on blur. Depending on which field is missing
- Clicking the individual apply buttons are not necessary
- The date picker doesn't blur when you click within it
- The date input has a border

Additionally, a bug with data reloading was fixed (useDeepCompareEffect seemed to be the culprit - to see what I mean, go to prod, change the query in the date filter without adding or removing a new filter. I observed that it wasn't reloading the data, which I also ran into when adding the onChange handlers to the dropdown components). Fixing that required a little bit of extra poking around in the codebase.

Not attempting to solve every UI issue with the date picker here - theres a ticket that needs refinin' that goes deeper.

## How to test
- The filter validation works as expected
- The blur behavior in general makes sense
- You can update the roles and the date without clicking the "secondary" apply buttons, just the big fella at the bottom
- API should not be called excessively

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-500

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


frontend/src/components/ActivityReportsTable/index.js
```diff
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types';
 import {
   Table, Checkbox, Grid, Alert,
 } from '@trussworks/react-uswds';
-import useDeepCompareEffect from 'use-deep-compare-effect';
+// import useDeepCompareEffect from 'use-deep-compare-effect';
```
AdamAdHocTeam, line 7: Probably can remove this if not needed.

frontend/src/components/DateRangeSelect.js
```diff
@@ -107,17 +109,19 @@ const CUSTOM_DATE_RANGE = OPTIONS[1].value;
...
+  options,
+  styleAsSelect,
+  updateDateRange,
+  disabled,
+  onChange,
+}) {
   const [selectedItem, setSelectedItem] = useState(1);
   const [showDateError, setShowDateError] = useState(false);
 
+  // this is a ref for the dropdown menu so that we can
```
AdamAdHocTeam, line 122: Comment appears cut off.

frontend/src/pages/Landing/index.js
```diff
@@ -272,7 +291,7 @@ function Landing({ user }) {
           filters={filters}
           showFilter
           onUpdateFilters={handleApplyFilters}
-          tableCaption={`Region ${regionLabel} Activity reports`}
+          tableCaption={`Region ${appliedRegion === 14 ? 'All' : appliedRegion.toString()} Activity reports`}
```
AdamAdHocTeam, line 294: Should this get moved to a function. Not sure if its worth it.

frontend/src/widgets/withWidgetData.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable react/jsx-props-no-spreading */
-import React, { useState } from 'react';
-import useDeepCompareEffect from 'use-deep-compare-effect';
+import React, { useState, useEffect } from 'react';
+// import useDeepCompareEffect from 'use-deep-compare-effect';
```
AdamAdHocTeam, line 3: Prob can remove.

AdamAdHocTeam (commented):
Looks like some minor code cleanup. Plus found a bug with filter validation where the topic stays red even though a value is set (as we discussed).

I experienced one issue where I couldn't get the filter menu to apply and close even though all values where valid. However when I went back and attempted to reproduce I couldn't. So this might have just been a weird one off.

jasalisbury:
Found a bug

1) Load the page fresh
2) Add a new specialist filter
3) Click cancel before the filter is valid (How the validation shows up when you click cancel is also strange, but not the main issue)
4) Change the date range filter to "before" or "after"
5) Clicking apply doesn't update the page

frontend/src/components/CheckboxSelect.js
```diff
@@ -31,29 +31,37 @@ export function renderCheckboxes(
...
+  const [checkboxes, setCheckboxes] = useState(
+    makeCheckboxes(options, toggleAllChecked),
+  );
 
   const menu = createRef();
 
+  useEffect(() => {
+    const checked = Object.keys(checkboxes).filter((checkbox) => checkboxes[checkbox]);
+    onChange(checked);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
jasalisbury, line 62: Are you running into issues adding `onChange` in the dependency list? This eslint rule is one I am really hesitant to ignore. It is probably the rule that has frustrated me the most, but ultimately ignoring it has always caused strange bugs and weird issues down the road. If you haven't already found it I found this post helpful https://github.com/facebook/react/issues/14920#issuecomment-471070149

frontend/src/widgets/withWidgetData.js
```diff
@@ -18,7 +18,7 @@ const withWidgetData = (Widget, widgetId) => {
 
     const { filters } = props;
 
-    useDeepCompareEffect(() => {
+    useEffect(() => {
```
jasalisbury, line 21: :tada: Thanks for tracking these down!

kryswisnaskas:
Noticed that duplicate filters are allowed for the specialists, but not for the dates. Is this on purpose?

frontend/src/components/ActivityReportsTable/index.js
AdamAdHocTeam: This also fixes TTAHUB-516 the filters not being updated.

thewatermethod:
> Noticed that duplicate filters are allowed for the specialists, but not for the dates. Is this on purpose?

I logged a bug about this issue awhile ago (if I understand what you're referring to)
https://ocio-jira.acf.hhs.gov/browse/TTAHUB-441

frontend/src/components/CheckboxSelect.js
thewatermethod: You're right to point this out, just the process of tracking down the re-renders led me to multiple bugs that needed fixing

thewatermethod:
> Found a bug

Believe this is resolved

kryswisnaskas:
> > Noticed that duplicate filters are allowed for the specialists, but not for the dates. Is this on purpose?
> 
> I logged a bug about this issue awhile ago (if I understand what you're referring to) https://ocio-jira.acf.hhs.gov/browse/TTAHUB-441

What I am seeing could be related, but not fully, I believe. It appears that sometimes the UI will not accept the same date, e.g.

image

The duplicate date above was flagged and the focus shifted to the top date. 

There is a way to make it accept it though, as visible on the screenshot. In this case, I am seeing a sequelize error, which I think might be due to the issue in TTAHUB-441, but not 100% sure.
`
SequelizeDatabaseError: invalid input syntax for type date:`

thewatermethod:
> The duplicate date above was flagged and the focus shifted to the top date.
> 
> There is a way to make it accept it though, as visible on the screenshot. In this case, I am seeing a sequelize error, which I think might be due to the issue in TTAHUB-441, but not 100% sure. ` SequelizeDatabaseError: invalid input syntax for type date:`

Thanks for the screenshot, @kryswisnaskas!
I think fixed the UI and can confirm that the additional issue is TTAHUB-441...

frontend/src/components/CheckboxSelect.js
```diff
@@ -48,6 +48,7 @@ export default function CheckboxSelect({
   disabled,
   hideToggleAll,
   onChange,
+  previousValue,
```
jasalisbury, line 51: Can we use something like https://blog.logrocket.com/how-to-get-previous-props-state-with-react-hooks/ so the user of `CheckboxSelect` don't have to worry about passing in the previous value explicitly?

frontend/src/components/filter/FilterMenu.js
```diff
@@ -54,27 +65,43 @@ export default function FilterMenu({ filters, onApplyFilters }) {
...
-  const onUpdateFilter = (id, name, value, toggleAllChecked) => {
-    const newItems = [...items];
+  const onUpdateFilter = (id, name, value) => {
+    //
+    // this is bonkers... we need to do more than map the array
+    // what escaped me originally was that just an array spread creates a new
+    // array of references... to the same objects as before
+    // therefore, this function was mutating state in unexpected ways
+    //
+    // hence this real humdinger of a line of javascript
```
jasalisbury, line 80: :upside_down_face: Another option is https://lodash.com/docs/4.17.15#cloneDeep (I don't really have a preference, I think this is fine as is)

jasalisbury:
> > Found a bug
> 
> Believe this is resolved

:tada: I can not recreate anymore

frontend/src/components/__tests__/Filter.js
```diff
@@ -128,6 +128,8 @@ describe('filter', () => {
     let topic = await screen.findByRole('combobox', { name: 'topic' });
     userEvent.selectOptions(topic, 'startDate');
 
+    screen.logTestingPlaygroundURL();
```
jasalisbury, line 131: TIL testing-playground. You should give a small demo sometime, I'm interested in how it works. I could be misunderstanding but seems like we can remove this line.

jasalisbury (approved):

frontend/src/components/__tests__/Filter.js
thewatermethod: whoops yes -  maybe at the engineering meeting

AdamAdHocTeam (commented):
Found a bug where we don't validate an empty date range filter.

Create a date range filter with 'Is Within' then select 'Custom Date Range' and apply.

You will see the Date Range pill is created with invalid date but no validation was caught.

thewatermethod:
This is waiting on finalized names for each of the fields, once that is provided, the labels for each control need to be updated to match the validation names

PR #684, opened 2021-12-14
TTAHUB-313: Update Activity Report Landing to Use New Filter

AdamAdHocTeam:
## Description of change

Update activity report landing to use the new filter. 

The three components (Overview, Alerts, AR) on the landing page should now all respect a single filter at the top of the page.

When visiting the page thee filter should only set the region filter (based on the users permissions). 

It should not implement any date constraints by default.

## How to test

View the AR landing page. Adjusting the filter at the top should update all three components on the page. All existing page functionality like exporting reports and sorting should still work as it did before.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-313

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


frontend/src/pages/Landing/index.js
```diff
@@ -40,16 +41,21 @@ export function renderTotal(offset, perPage, activePage, reportsCount) {
...
-    condition: 'Contains',
-    query: regionId.toString(),
-  };
-}
-
 function Landing({ user }) {
+  // Determine Default Region.
   const regions = allRegionsUserHasPermissionTo(user);
+  const regionToUse = regions[0] || user.homeRegionId || 0;
+  const defaultRegion = user.homeRegionId === 14 ? 14 : regionToUse;
```
GarrettEHill, line 48: What are all the checks for region 14 when there are only 12 regions?

frontend/src/components/filter/constants.js
```diff
@@ -156,6 +156,23 @@ export const FILTER_CONFIG = [
       />
     ),
   },
+  {
+    id: 'reportId',
+    display: 'Report ID',
+    conditions: SELECT_CONDITIONS,
+    defaultValues: EMPTY_TEXT_INPUT,
+    displayQuery: handleStringQuery,
+    renderInput: (id, condition, query, onUpdate, onApplyQuery) => (
```
thewatermethod, line 165: ```renderInput: (id, condition, query, onApplyQuery) => (```

Refactored to drop the onUpdate since they were functionally the same

frontend/src/pages/Landing/index.js
AdamAdHocTeam: Region 14 is all regions.

frontend/src/components/filter/constants.js
AdamAdHocTeam: Thanks 👍

jasalisbury:
Ran into some issues (most of these seem to be issues for different PRs, this is just where I noticed them)

1) The styling on the multiselect selected items is different from the report
Screenshot from 2021-12-15 13-56-19
vs 
Screenshot from 2021-12-15 14-00-02

2) Strange state behavior on initial load inside the filter component. If I open the region dropdown and close it (without selecting anything) when focus leaves that filter I get a validation error
Screenshot from 2021-12-15 14-01-52

3) The filters aren't combining in a way I'd expect, if I have a filter for all regions, and add another filter with the top two reasons no new data is loaded (I still see the request hit the API as `region.in[]=14&reason.in[]=Below%20Competitive%20Threshold%20(CLASS),Below%20Quality%20Threshold%20(CLASS)`. Seems like the two conditions are being "OR'd" instead of "AND'd". May be an issue specific to the reason filter, doesn't seem to happen with `region` and `specialist role`

4) This one is definitely unrelated to any new code, but if you can I'd try to match up all the widths of the components on the page, so each section is flush with one another. If it turns into a big change we can write a separate ticket for it, just want to write it down somewhere because it has been in my head for a while.

jasalisbury:
Another issue I just noticed, when the focus is on a value the focus needs to leave that input box before I can click `Apply`. If I make a region selection, for example, the select box closes with the input still focused, which requires a click away to clear the focus before I can click apply. This feels strange when I try to clear the focus on the `Apply` button and nothing happens. Makes it feel like I have to click the `Apply` button twice for it to do anything.

jasalisbury:
Also (last comment for now, I swear!) it feels a little strange to me that we can remove the `region` filter, especially when we can select `all regions`. Doubly so for users with access to a single region where removing the region filter results in nothing changing (they still only get reports from a single region). I can think of two solutions (probably are more options!)

1. We remove the region filter for single region users, only show it for multi-region users and they will hopefully figure out having no region filter is the same as `all regions`
2. We remove the ability to remove the `region` filter (make it a "required" filter passed into the filter component or something?) and only allow users to change the value of that filter

Those two options don't feel exclusive, could probably do both. Also related to single region users can we remove the "all regions" option if the user can only see reports in a single region (obviously moot point if we go with something like option 1 above).

thewatermethod:
> Another issue I just noticed, when the focus is on a value the focus needs to leave that input box before I can click Apply. If I make a region selection, for example, the select box closes with the input still focused, which requires a click away to clear the focus before I can click apply. This feels strange when I try to clear the focus on the Apply button and nothing happens. Makes it feel like I have to click the Apply button twice for it to do anything.

This is fixed on #664 (fyi @AdamAdHocTeam )

thewatermethod:
Re: the region filter. Seems like we have to confer with the team since its a departure from what happens now, but here's what seems to me to be the most elegant solution

- Remove all regions as an option. All regions === no filter
- By default, a non-central office user will have a region filter set to the first of their regions (like now). If they remove it, they see data for all their regions
- By default, a central office user has no region filter

GarrettEHill (approved):

frontend/src/pages/Landing/index.js
```diff
@@ -40,16 +43,14 @@ export function renderTotal(offset, perPage, activePage, reportsCount) {
...
-    topic: 'region',
-    condition: 'Contains',
-    query: regionId.toString(),
-  };
-}
-
 function Landing({ user }) {
+  // Determine Default Region.
   const regions = allRegionsUserHasPermissionTo(user);
+  const regionToUse = regions[0] || user.homeRegionId || 0;
```
jasalisbury, line 49: I think we want to flip `regions[0]` with `user.homeRegionId` so we use `homeRegionId` if we can, the first region numerically if `homeRegionId` isn't set, and fall back to 0 if they have no permissions on any region. Honestly we can probably remove the `|| 0` completely and treat users with no home region and no premissions to read any region as a "central office" user that just sees nothing on this page.

If we keep this as is a user with a `homeRegionId` of 2 and permission to read both region 1 and 2 will default to region 1 on this page.

jasalisbury:
> Re: the region filter. Seems like we have to confer with the team since its a departure from what happens now, but here's what seems to me to be the most elegant solution
> 
> * Remove all regions as an option. All regions === no filter
> * By default, a non-central office user will have a region filter set to the first of their regions (like now). If they remove it, they see data for all their regions
> * By default, a central office user has no region filter

To add to this, for the future unless it is easy to do, if a user has a single region no region filter is applied by default and there is no option to add a region filter.

jasalisbury (commented):
:rocket: Only issue I've found is the bottom table doesn't load the correct data for central office users with no region filter. The bottom table only loads reports for a single region. I believe the culprit is the last param [here](https://github.com/HHS/Head-Start-TTADP/blob/main/src/routes/activityReports/handlers.js#L511) needs to be flipped to false (once we understand why it was true to begin with).

frontend/src/pages/Landing/index.js
AdamAdHocTeam: So do you think the code should just become this?
`
 const defaultRegion = user.homeRegionId || regions[0];
`

jasalisbury: I _think_ so, assuming the page doesn't just explode for a user with zero read permissions

jasalisbury (approved):
:+1: I'd wait to merge this into main once #664 gets merged into main

PR #689, opened 2021-12-16
Update maintenance mode docs

jasalisbury:
## Description of change

Maintenance mode persists through deploys

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,7 +419,7 @@ Ex.
 ./bin/maintenance -e dev -m on
 ```
 
-If you are not logged into the cf cli, it will ask you for an sso temporary password. You can get a temporary password at https://login.fr.cloud.gov/passcode.
+If you are not logged into the cf cli, it will ask you for an sso temporary password. You can get a temporary password at https://login.fr.cloud.gov/passcode. The application will stay in maintenance mode even through deploys of the application. You need to explicitly run `./bin/maintenance -e ${env} -m off` to turn off maintenance mode.
 
 
 <!-- Links -->


README.md
```diff
@@ -419,7 +419,7 @@ Ex.
 ./bin/maintenance -e dev -m on
 ```
 
-If you are not logged into the cf cli, it will ask you for an sso temporary password. You can get a temporary password at https://login.fr.cloud.gov/passcode.
+If you are not logged into the cf cli, it will ask you for an sso temporary password. You can get a temporary password at https://login.fr.cloud.gov/passcode. The application will stay in maintenance mode even through deploys of the application. You need to explicitly run `./bin/maintenance -e ${env} -m off` to turn off maintenance mode.
```
GarrettEHill, line 422: Could we make a maintenance script like the following for any time we need maintenance:
`#!/bin/bash
echo "Begin maintenance"
./bin/maintenance -e dev -m on
bash
./bin/maintenance -e ${env} -m off
echo "End maintenance"`

This would guarantee that maintenance is ended correctly at the end of the session.

jasalisbury: I don't think there is an issue forgetting to turn off maintenance, the reason I added this line is because I wasn't sure if a deploy to the environment would cause maintenance mode to turn off. One thing to improve maintenance mode would be to add a way for authorized users (devs) to still view the application when in maintenance mode.

kryswisnaskas: Yes, adding the ability to view the app behind the maintenance mode would be helpful.

thewatermethod (approved):

kryswisnaskas (approved):

GarrettEHill (approved):

PR #699, opened 2021-12-23
[TTAHUB-506] Redo date picker UI

thewatermethod:
## Description of change
- Remove react-dates in favor of the react-usdws date picker.
- Remove code for previous filter implementation, since it isn't being used
- Date picker now has four options for conditions (Is after, Is before, Is within, and **in**)
- "In encompasses preset date ranges (last 30 days, etc)

## How to test
Use the date picker in the filter menu and in the activity report creation screen.
Ensure that everything works as expected.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-506

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


jasalisbury (commented):
Everything works for me! The only issue I ran into is setting the `Date Range` filter to `In` causes the validation to show even through "Year to date" is selected. I have to change it to "Last thirty days" and back to get the validation error to go away.

jasalisbury (commented):
A found some issues, all of these are on the `Regional TTA Activity Dashboard`

1) In the filter with `Date range` `is before` I can type text in the box and click apply with no validation message showing up. The input is colored red. Nothing is actually applied. If I reopen the filter menu I still see what I typed in the text box (with the red outline and no message)
2)  If I start with a valid filter of `Date range` `Is before` and switch to `Is within` the date format flips from `MM/DD/YYYY` to `YYYY/MM/DD` in the display (so "01/03/2020" to "2020/01/03") in the text box. When I open the menu I don't see a start date or end date. Screenshot from 2022-01-05 15-42-43
3) Typing a valid date ("01/03/2022") into the `Date range` `Is before` text box is not accepted as a valid date. If I remove the last digit and re-add the validation error goes away. Also if I add it as a completely new filter and use my browser's auto fill feature to select a previously type value there is no validation error
4) Looks like the date format between for the date range text box is different from the individual date pickers below (related to my second issue?)
Screenshot from 2022-01-05 15-47-57
5) If I already have a valid date filter (`Date range` `Is after` "01/03/2022") and type a new date ("05/20/2021") the apply button closes the menu but no new data is loaded. If I remove the last digit and re-add new data is loaded but the date is wrong in the pill (note the year of "0202")
Screenshot from 2022-01-05 15-52-37
6) Selecting a start date after the end date can cause the end date to push out past today. Select an end date of Jan 1, start date a few months back. Update the start date to today and the end date is pushed out past today. The end date text box is red with no validation message, but you can still click `apply` to close the menu

In the form

1) The calendar has the year scrunched, seems like only Nov and Sept
Screenshot from 2022-01-05 16-03-04
2) Similar issue with text input (as in letters), the box shows as red but no validation message

kryswisnaskas:
The new date picker looks good. I am still reviewing, but so far I run across the date picker accepting future dates. This sequence was used to reproduce:

- on a new AR creation select today's start and end date followed by "save draft"
- select 1/1/22 as the start and end date; don't save
- move the start date to 1/12/22 
- note everything works well; the end date moves to 1/12/22
- move the start date to 1/5/22
- still everything looks ok; the end date stays at 1/12/22
- move the end date to today's date (e.g. 1/19/22)
- note the end date moves to 1/26/22 and saving is allowed

thewatermethod:
> The new date picker looks good. I am still reviewing, but so far I run across the date picker accepting future dates. This sequence was used to reproduce:
> 
>     * on a new AR creation select today's start and end date followed by "save draft"
> 
>     * select 1/1/22 as the start and end date; don't save
> 
>     * move the start date to 1/12/22
> 
>     * note everything works well; the end date moves to 1/12/22
> 
>     * move the start date to 1/5/22
> 
>     * still everything looks ok; the end date stays at 1/12/22
> 
>     * move the end date to today's date (e.g. 1/19/22)
> 
>     * note the end date moves to 1/26/22 and saving is allowed

good catch! should be addressed now

jasalisbury:
Is this ready for another look?

thewatermethod:
> Is this ready for another look?

yes sir

jasalisbury (approved):
:rocket: Nice! Everything seems to work as expected.

kryswisnaskas:
The issue with a future date is gone 👍 

There is still a minor inconsistency in the date picker. In the scenario used to reproduce the "future date", if in the second to last step, a different start date is selected (not today's, but something between 1/12 and 1/20), the end date adjusts to something random, it appears.

thewatermethod:
> There is still a minor inconsistency in the date picker. In the scenario used to reproduce the "future date", if in the second to last step, a different start date is selected (not today's, but something between 1/12 and 1/20), the end date adjusts to something random, it appears.

Hmm. When I try this, following the steps you provided above, the end date is set to today's date on the second to last step (like, if the delta between the two would push the end date after today, it is set to today instead) 

Is that not what you are seeing?

kryswisnaskas:
I am seeing the end date being set to today's date in most of the cases, but not all, e.g. sometimes it was an earlier date (1/19/22).

kryswisnaskas:
I was expecting to see the end date would be set to the start date in the case rather than today's date - so that's the inconsistency. However, it is minor, so perhaps we could go with it.

kryswisnaskas (approved):

PR #707, opened 2022-01-03
[TTAHUB-553] remove grantee_record_page feature flag from users/backend

thewatermethod:
## Description of change
Remove the grantee record page feature flag from the user model, and bulk update all users to remove that feature flag. (It's already been removed from the frontend)

## How to test
Run the migration. Ensure you have no users with the feature flag, that the feature flag cannot be assigned, and that everything else works as expected

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-553

### Every PR

- [x] Meets issue criteria

### Production Deploy

- [] Staging smoke test completed

### After merge/deploy

### DIFF
--- a/src/migrations/20220103142419-remove-grantee-record-flag.js
+++ b/src/migrations/20220103142419-remove-grantee-record-flag.js
@@ -1,0 +1,12 @@
+module.exports = {
+  up: async (queryInterface) => queryInterface.sequelize.transaction(
+    async (transaction) => {
+      await queryInterface.bulkUpdate('Users', {
+        flags: [],
+      }, {}, { transaction });
+    },
+  ),
+  down: async (queryInterface) => queryInterface.sequelize.transaction(
+    // whoops, don't think this can be rolled back
+  ),
+};
--- a/src/models/user.js
+++ b/src/models/user.js
@@ -19,9 +19,7 @@ const roles = [
   'System Specialist',
 ];
 
-export const featureFlags = [
-  'grantee_record_page',
-];
+export const featureFlags = [];
 
 const generateFullName = (name, role) => {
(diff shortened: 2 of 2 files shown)


src/migrations/20220103142419-remove-grantee-record-flag.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  up: async (queryInterface) => queryInterface.sequelize.transaction(
+    async (transaction) => {
+      await queryInterface.bulkUpdate('Users', {
+        flags: [],
+      }, {}, { transaction });
+    },
+  ),
+  down: async (queryInterface) => queryInterface.sequelize.transaction(
```
AdamAdHocTeam, line 9: Is there ever going to be a case where we might want to add it back (if possible)?

thewatermethod: I don't see how/why at this point, maybe when we're storing more historical data that would be possible

AdamAdHocTeam: I guess we could always add it back manually. I was just thinking hypothetically if some issue was found and we wanted to put it back behind the feature flag.

AdamAdHocTeam (approved):
Removes the recipient record page flag as expected.

src/migrations/20220103142419-remove-grantee-record-flag.js
thewatermethod: no that makes sense - that was kind of why I left the flag in place on the backend for a little bit, in case something came up in the first couple of weeks

jasalisbury (approved):

PR #712, opened 2022-01-06
Update backend dependencies

jasalisbury:
## Description of change

Update backend dependencies that were relatively easy to update.

Only small changes were required to get the code to work with new packages. Most of the changes were around required eslint changes that honestly might not even be required if I can't get eslint updated. Other big change was around how models are loaded in `src/models/index.js` since sequelize removed the `sequelieze.import` method.

Another positive is I was able to completely remove the audit ignored issues file. `yarn audit` only shows three moderate vulnerabilities, one for jest and the other two for our API documentation viewer.

There are a few packages that I haven't updated (yet!):

`@cucumber` - Don't want to update to release candidate
`csv-*` - Had a hard time with these, might try again
`eslint` - The eslint plugin we use to use multiple formatters blows up when eslint is updated. Will explore solutions
`jest*` - Also had a hard time updating these
`file-type` - Caused some errors when I updated

I will take another pass updating these packages, specifically `file-type`, `eslint` and the `csv-*` packages. Jest feels like a larger project and cucumber isn't _really_ out of date.

`yarn outdated` before:

Screenshot from 2022-01-06 15-46-09

and after:

Screenshot from 2022-01-07 13-47-03

## How to test

Browse the site in sandbox and make sure everything still works. Run migrations and other scripts locally to verify they still function properly.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-458

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


src/env.js
```diff
@@ -1,4 +1,3 @@
-import dotenv from 'dotenv';
+require('dotenv').config();
```
thewatermethod, line 1: There's a couple spots here were you swapped from CommonJS to esm or vice versa. 

I'm just curious why, and also I wanted to point out that [there's a ticket to possibly switch us completely over to CommonJS on node](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-331)

Anyways, I think newer versions of Node will support esm out of the box, but it could be awhile before we get there. (#702)

thewatermethod (approved):
Everything works (the app on sandbox, app & migrations locally)

I assume most of the formatting changes are some kind of eslint being run locally on your machine?

Thanks for doing this!

jasalisbury:
> I assume most of the formatting changes are some kind of eslint being run locally on your machine?

Yeah I updated `eslint` and had to roll back because of the `eslint-formatter-multiple` package didn't play nice with the new version of `eslint`. I need to look into it more and figure out if we can get everything to work with the updated `eslint`.

src/env.js
jasalisbury: I had to convert all our models to CommonJS to get the importing in `src/models/index.js` to work. I agree it would be nice to land on CommonJS or ESM, but I am not knowledgeable enough on the differences to know which one to switch to (or how much work it will be)

AdamAdHocTeam (approved):
Site works seems to work same as before.

GarrettEHill (approved):

src/models/index.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable import/no-dynamic-require */
+/* eslint-disable global-require */
```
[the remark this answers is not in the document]
jasalisbury: I don't think so in this case. I tried to make as few code changes as possible to get running on new versions of packages. Sequelize changed the way models are imported. I haven't found an example of how they expect models to get imported after they removed the `import` function. Figuring out how to import all our models non-dynamically at the top of the file seems like a bigger projected then appropriate for a change focused on updating packages.

PR #732, opened 2022-01-18
Ttahub 598 db errors

GarrettEHill:
## Description of change

Check that all foreign keys references have been removed prior to deletion from each table.

## How to test
 On CircleCI in the backend test, the "Container circleci/postgres:12.4-ram" does not have any errors.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-598

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/testUtils.js
+++ b/src/testUtils.js
@@ -172,17 +172,27 @@ export async function destroyReport(report) {
         },
       });
 
-      await Grant.destroy({
-        where: {
-          id: grant.id,
-        },
-      });
+      await ActivityRecipient.findAll({ where: { grantId: grant.id } })
+        .then(async (results) => {
+          if (results.length === 0) {
+            await Grant.destroy({
+              where: {
+                id: grant.id,
+              },
+            });
+          }
+        });
 
-      await Recipient.destroy({
-        where: {
-          id: grant.recipientId,
-        },
-      });
+      await Grant.findAll({ where: { recipientId: grant.recipientId } })
+        .then(async (results) => {
+          if (results.length === 0) {
+            await Recipient.destroy({
+              where: {
+                id: grant.recipientId,
+              },
+            });
+          }
+        });
     } catch (e) {
       // ignore fk errors
     }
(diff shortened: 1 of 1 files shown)


src/testUtils.js
```diff
@@ -172,17 +162,27 @@ export async function destroyReport(report) {
...
-          id: grant.id,
-        },
-      });
-
-      await Recipient.destroy({
-        where: {
-          id: grant.recipientId,
-        },
-      });
+      await ActivityRecipient.findAll({ where: { grantId: grant.id } })
```
jasalisbury, line 165: We typically don't chain promises, can we change this to something like
```
const result = await ActivityRecipient.findAll(....)
if (results.length === 0) {
  await Grant.destroy(....)
}
```

GarrettEHill: Sure

thewatermethod (approved):

src/testUtils.js
```diff
@@ -152,16 +152,6 @@ export async function destroyReport(report) {
     },
   });
 
-  try {
```
thewatermethod, line 155: Why remove this and not wrap it in a similar "if" statement to the ones below?

GarrettEHill: It was not removed, but moved to after the ActivityReport entries are removed which reference the user table.

GarrettEHill: Should I remove the try/catch's as all the fk issues have now been addressed?

jasalisbury: I'd say yes, I think if we run into an error that is not a fk error we want to error out so we can see the error

src/testUtils.js
```diff
@@ -152,39 +152,30 @@ export async function destroyReport(report) {
...
-    User.destroy({
+  const destroys = dbReport.activityRecipients.map(async (recipient) => {
```
jasalisbury, line 155: Nothing is actually returned from the map, so I'd assume `destroys` is always going to be empty. With `await`s being used I would think using `for (const recipient of dbReport.activityRecipients) {` instead of the  map would be clearer (and removing `the `await Promise.all(destroys)` below).

GarrettEHill: Its needed, as the content of the mapped function is async and all the calls within are awaits. Changing to a for of loop makes awaits much less clear.

jasalisbury: Just had a chat with Garrett, using map here seems good

PR #734, opened 2022-01-18
[TTAHUB-565] Add filters and scopes that apply to the goals model

thewatermethod:
## Description of change
Add the filter panel to the Goals & objectives tab
Create filters and scopes to filter goals and objectives. (createDate, reasons, status, topics)

## How to test
There is no data to be filtered yet, so confirm scopes and unit tests are written properly as best you can. Also confirm that the filter panel appears on the Goals & objectives tab, and that it shows the correct filters and filter options.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-565

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


frontend/src/components/filter/constants.js
```diff
@@ -41,6 +47,35 @@ const handleArrayQuery = (q) => {
 const handleStringQuery = (q) => q;
 
 export const FILTER_CONFIG = [
+  {
```
jasalisbury, line 50: I'm sure I've said this before, so sorry for bringing it up again! Also I think this is fine as is but just thinking about a possible refactor (now or later!). I feel having the filter config for any possible filter in one object is going to get pretty confusing especially when we have a three different "topics" each of these filters deal with (`activity reports`, `grants`, and `goals`). And most filters only apply to a single "topic", the status filter doesn't really make sense for activity reports (at least how it is currently defined). Actually I think status is a good point in favor of splitting this constant out a bit, because if we decide we want a "status" filter for "activity reports" and for "goals" we have duplicated entries in this array with only a different `renderInput`.

It would feel better to me to have a common "library" of possible filters individually exported that pages could pull in to create their own `FILTER_CONFIG`. Even if the "library" idea is bad I think we could still just hardcode the constants directly in (or next to) each page that uses filters. Maybe we don't have to have each page have it's own `FILTER_CONFIG` because some can be shared between pages but I think most pages have unique filters. So the regional dashboard would have it's own `FILTER_CONFIG`, and the recipient record page would as well. That would also mean we can simplify the filter panel component so it takes in the config directly and doesn't have to worry about `allowed filters`.

Having each page have it's own independent `FILTER_CONFIG` _should_ allow better control on how each filter is rendered. For example we could have different options for a `start date` filter on different pages (maybe the regional dashboard needs an "is" condition of "last quarter" that we don't want shown on the landing page).

jasalisbury (approved):
:+1: Looks good, main comment is on how we pass which filters to use to the frontend `FilterPanel` component and it is more just thoughts about a possible refactor, I don't think it is anything you need to take action on now unless you feel like it.

frontend/src/components/filter/constants.js
thewatermethod: I was going to write a jira ticket to refactor the filters but it feels a little strange to ship code with a built in tech debt ticket so I'm going to attempt a refactor. I'm moving towards making the filters more modular, so you can assemble a config per page as you suggest, killing the ```allowedFilters``` prop. 

They'll likely still be pretty option free on take one but breaking them out as you say will at least leave the door a little more open in the future to customize them a little more per page.

frontend/src/components/filter/goalFilters.js
```diff
@@ -0,0 +1,102 @@
...
+  displayQuery: handleArrayQuery,
+  renderInput: (id, condition, query, onApplyQuery) => (
+    <FilterReasonSelect
+      inputId={`reason-${condition}-${id}`}
+      onApply={onApplyQuery}
+      query={query}
+    />
+  ),
+};
+
```
thewatermethod, line 73: @jasalisbury Building on your suggestions, I broke apart the filters and they are now imported into individual files. All the actual display components still live in the root of the /filter folder. Right now, there is only need for one version of each as far as React components, if that makes sense, so it feels a bit premature to sort them. (They'd largely go in one "shared" folder right now, I think.)

I figured that in these ```activityReportFilters```, ```goalFilters```, etc, where objects are currently being exported we could convert them to functions as necessary, and that would allow us to do the configuration in the constant files where they are being imported on a per page basis. In the example of the status filter, for example, we could create a function that passes in the options for the multiselect depending on what model we are filtering on.

What do you think? Does this make sense for a first pass?

jasalisbury (approved):
:shipit:

PR #747, opened 2022-01-24
Goal entry allows pasting from right click

jasalisbury:
## Description of change

The input was 2px wide set from react-select. This change makes the input width 100%. Had to move the placeholder because it was stacking under the input (so the placeholder was half outside of the input box). Had to set the color of placeholder for accessibility reasons. Class name is set because I'm not sure the signature of the `cx` prop (if it can take additional classes or not) so figured the safest bet is to assume `cx` returns a string (which is required by className) and just jam everything together in an interpolated string.

## How to test

Start a report, add recipients. Try to paste into the goal creation/selection box using only the mouse.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-481

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


thewatermethod (approved):
Works great!

Do you think we could just remove the placeholder entirely? This is the only field that uses it as far as I know, and it creates more issues than it solves (for example, ```::placeholder``` doesn't do anything for IE)

jasalisbury:
> Do you think we could just remove the placeholder entirely? This is the only field that uses it as far as I know, and it creates more issues than it solves (for example, `::placeholder` doesn't do anything for IE)

We can check at standup tomorrow. I would guess no since we display the number of goals selected in that dropdown.

thewatermethod:
> We can check at standup tomorrow. I would guess no since we display the number of goals selected in that dropdown.

You know, on second thought, I wouldn't worry about it, since we are replacing the form on the activity report anyways (AFAIK)

PR #749, opened 2022-01-26
Ttahub 614/missing transaction

GarrettEHill:
## Description of change
For the audit system to function as intended all modifications to the DB need to be part of a transaction. Some of the modifications to the DB are running outside of a transaction either due to no transaction being specified or the transaction is specified incorrectly.

## How to test
All current tests still pass, as business logic was not changed.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-614

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


src/tools/changeReportStatus.js
```diff
@@ -20,9 +21,12 @@ export default async function changeReportStatus(ids, status) {
...
       promises.push(
-        report.update({
-          submissionStatus: status,
-        }),
+        await sequelize.transaction(async (transaction) => report.update(
+          {
+            submissionStatus: status,
+          },
+          transaction,
+        )),
```
jasalisbury, line 29: I don't think we want the `await` in front of the `sequelize.transaction` line because the transaction is added to the `promises` array that eventually passed into `Promise.all`, which will wait for all promises in the array to finish before resolving.

jasalisbury: I actually think the argument can be made to push the creation of the transaction outside the `for await (const id of idsArray)` on line 18 so all report update calls happen in the same transaction.

src/tools/changeReportStatus.js
```diff
@@ -13,21 +14,25 @@ export default async function changeReportStatus(ids, status) {
...
-        report.update({
-          submissionStatus: status,
-        }),
-      );
-    } else {
-      auditLogger.info(`Couldn't find any reports with the id: ${id}`);
+      if (report) {
+        auditLogger.info(`Changing status of report: ${id} to ${status}`);
+        promises.push(
+          await report.update(
```
jasalisbury, line 24: I think we either need to remove the `await` from the `report.update` call or remove the `promises` array altogether. Calling `await report.update(...)` will wait for the promise to return before moving on to the next id in `idsArray`. The `promise.all` at the end does nothing because all promises in the array have already resolved (because they were await'ed).

thewatermethod (approved):
This looks good to me. Re-ran the effected the tests locally and everything seems kosher.

jasalisbury (approved):

PR #751, opened 2022-01-26
remove slack notification for now

thewatermethod:
## Description of change

Removing slack notifications from the circle ci jobs "test_frontend" and "test_backend" for now
We could enable them on for test_frontend & test_backend on every run, but not just for the daily scan, or at least that's how it looks to me

Maybe we want to do so, but I think we should discuss first in case someone is worried about channel spam
Removing them here because I've nerfed the builds to main and prod where a slack credential is not provided, as things are now

## Issue(s)

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,6 @@ version: 2.1
 orbs:
   node: circleci/node@4.1.0
   browser-tools: circleci/browser-tools@1.2.3
-  slack: circleci/slack@4.5.1
 executors:
   docker-executor:
     # for docker you must specify an image to use for the primary container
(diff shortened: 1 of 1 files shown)


thewatermethod:
> There is a way to get notified of build failures via circle CI via email. I get hit with daily scan failures when they happen (along with other failures). I wonder if we can just enable the email notifications and then we won't have to worry about notifying slack, unless anyone feels a slack message will spur more action.

I like slack better on balance, but not if its gonna be a pain to get it to work

PR #762, opened 2022-02-03
[TTAHUB-252] save AR ordering and sorting

thewatermethod:
## Description of change
Add some cookies to preserve pagination and sorting on the activity report component.

## How to test
Observe when changing pages and refreshing that your sort and page is preserved. You can see what's happening by looking closely at your browser cookies.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-252

### Every PR

- [X] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -16,6 +16,7 @@
     "draftjs-to-html": "^0.9.1",
     "html-to-draftjs": "^1.5.0",
     "http-proxy-middleware": "^2.0.1",
+    "js-cookie": "^3.0.1",
     "lodash": "^4.17.20",
     "moment": "^2.29.1",
     "moment-timezone": "^0.5.33",
@@ -68,7 +69,7 @@
     "ansi-regex": "^5.0.1",
     "nanoid": "^3.2.0",
     "node-fetch": "^2.6.7",
-    "follow-redirects":"^1.14.7"
+    "follow-redirects": "^1.14.7"
   },
   "eslintConfig": {
     "root": true,
--- a/frontend/src/components/ActivityReportsTable/__tests__/index.js
+++ b/frontend/src/components/ActivityReportsTable/__tests__/index.js
@@ -1,5 +1,6 @@
 import '@testing-library/jest-dom';
 import React from 'react';
+import Cookies from 'js-cookie';
 import {
   render, screen, fireEvent, waitFor, act,
 } from '@testing-library/react';
@@ -368,13 +369,17 @@ describe('Table sorting', () => {
   it('clicking Topics column header will sort by topics', async () => {
     const columnHeader = await screen.findByText(/topic\(s\)/i);
 
+    const mockSet = jest.fn();
+    Cookies.set = mockSet;
+
     fetchMock.get(
       '/api/activity-reports?sortBy=topics&sortDir=asc&offset=0&limit=10&region.in[]=1',
       { count: 2, rows: activityReportsSorted },
     );
 
-    await act(async () => fireEvent.click(columnHeader));
+    act(async () => fireEvent.click(columnHeader));
     await waitFor(() => expect(screen.getAllByRole('cell')[15]).toHaveTextContent(/Behavioral \/ Mental Health CLASS: Instructional Support click to visually reveal the topics for R14-AR-1$/i));
+    expect(mockSet).toHaveBeenCalled();
   });
 
   it('clicking Creator column header will sort by author', async () => {
--- a/frontend/src/components/ActivityReportsTable/__tests__/index.js
+++ b/frontend/src/components/ActivityReportsTable/__tests__/index.js
@@ -1,6 +1,5 @@
 import '@testing-library/jest-dom';
 import React from 'react';
-import Cookies from 'js-cookie';
 import {
   render, screen, fireEvent, waitFor, act,
 } from '@testing-library/react';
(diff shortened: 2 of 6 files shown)


jasalisbury:
Should this also save filters? If I set custom filters, open a report, and then click the `Activity Reports` link the sorting is preserved but my filters are the default filters for the page.

thewatermethod:
> Should this also save filters? If I set custom filters, open a report, and then click the `Activity Reports` link the sorting is preserved but my filters are the default filters for the page.

We discussed at standup - filters are shared through URLs and bookmarking, pagination and sorting are not shareable and saved per user

jasalisbury:
> We discussed at standup - filters are shared through URLs and bookmarking, pagination and sorting are not shareable and saved per user

It feels strange to me (but maybe it is okay!) that I can set a filter and sort, browse away from the page, and when I come back only the sorting is saved.

thewatermethod:
> It feels strange to me (but maybe it is okay!) that I can set a filter and sort, browse away from the page, and when I come back only the sorting is saved.

Well that seems like something we can solve, hopefully (resetting the saved positions if the filters aren't saved), or maybe it just makes sense to also save the filters to cookies

Of course, if we save filters to cookies, we do start opening weird situations for ourselves as far as what would take precedence, the URL, the cookies, etc, but probably solvable just as well

jasalisbury:
Now when I set a sort column, browse away, and come back my sorting isn't saved

thewatermethod:
Did you try clearing your cookies? It's working for me

jasalisbury:
> Did you try clearing your cookies? It's working for me

Yeah tried clearing cookies and running in incognito, still wouldn't save

thewatermethod:
> Yeah tried clearing cookies and running in incognito, still wouldn't save

Ok, maybe you can ping me when you get in/get a chance in the morning and just show me real quick what you are seeing? Not sure how to repro

jasalisbury (commented):
Working well for me! One thing, when I make a request for data all the filter info is sent up in the cookie to the server. Can we switch to local storage (or something else) so we aren't sending up stuff we don't need in the cookie?

jasalisbury:
Also, doesn't seem to handle setting a sort option and changing the page (on the activity reports table I set the sort to start date and the page to 3)

jasalisbury:
Are we sure we want to have different filter/sorting saved for each recipient's activity report table on the `TTA History` tab? I could see a user that wants to compare reports across different recipients wanting to keep the filters/sorting/paging the same between recipients.

I was going to ask a similar question for all pages in general but I think with the different filtering options we can't save the filters "globally" across different pages. At least it seems that way, maybe there is a half step? Maybe it isn't worth it at all?

frontend/src/hooks/useUrlFilters.js
```diff
@@ -11,18 +12,36 @@ const { history } = window;
  * @param {Object[]} defaultFilters
  * @returns {[ Object[], Function ]}
  */
-export default function useUrlFilters(defaultFilters) {
+export default function useUrlFilters(key, defaultFilters) {
```
jasalisbury, line 15: I'm going to preface this suggestion by saying it may not be a good one, but it seems like this hook is no longer the `useUrlFilters` hook but `useSessionFiltersAndReflectInUrl` hook. Can we split these loading/reflecting filters in the URL and storing/loading filters from session storage into separate hooks and call one from the other?

frontend/src/hooks/useUrlFilters.js
```diff
@@ -11,18 +12,36 @@ const { history } = window;
...
-  const params = useMemo(() => queryStringToFilters(new URL(window.location).search.substr(1)), []);
-  const [filters, setFilters] = useState(params.length ? params : defaultFilters);
+  const initialValue = useMemo(() => {
+    const params = queryStringToFilters(new URL(window.location).search.substr(1));
+    if (params.length) {
+      return params;
+    }
+
+    try {
+      const cookieFromStorage = sessionStorage.getItem(key);
```
jasalisbury, line 25: will want to rename this variable

frontend/src/pages/Landing/index.js
```diff
@@ -320,11 +324,13 @@ function Landing() {
           setDownloadAlertsError={setDownloadAlertsError}
           downloadAllAlertsButtonRef={downloadAllAlertsButtonRef}
         />
-        <ActivityReportsTable
-          filters={filtersToApply}
-          showFilter={false}
-          tableCaption="Approved activity reports"
-        />
+        <FilterContext.Provider value={{ filters, filterKey: FILTER_KEY }}>
```
jasalisbury, line 327: Three questions:
1) Should this we be putting `filtersToApply` in this context instead of `filters`?
2) I can't remember, why do we need to pass `ActivityReportsTable` `filtersToApply` instead of just filters?
3) If the filters sent to the context matches the prop sent to the `ActivityReportsTable` component (either both `filters` or both `filtersToApply`) can we drop the context altogether?

frontend/src/testHelpers.js
```diff
@@ -1,7 +1,3 @@
-// Disable eslint rule making this a default export. This file will,
```
jasalisbury, line 1: Only took a year and a half to accumulate more helper functions :joy:

AdamAdHocTeam (commented):
Not sure if this is something we have to worry about our users getting.

image

Turns out I had an explicit 'no cookies' for my local host. This does make me wonder if our gov users will have specific security settings in regards to cookies.

AdamAdHocTeam (approved):
Works as advertised.

I would make sure any concerns Josh has get addressed before shipping.

thewatermethod:
> Turns out I had an explicit 'no cookies' for my local host. This does make me wonder if our gov users will have specific security settings in regards to cookies.

Was the site crashing for you? Or just failing to save to storage?

thewatermethod:
> I would make sure any concerns Josh has get addressed before shipping.

Will do! Thanks.

GarrettEHill (approved):

PR #765, opened 2022-02-04
Ttahub 588/db based audit

GarrettEHill:
## Description of change
Postgresql has a make table/trigger/function name length of 63 bytes.
In order to more likely stay under this limit all names are abbreviated where possible.
Statically defined names are known to be short enough, but are still abbreviated to keep
  consistency.
All dynamically generated table/trigger/function are abbreviated where possible.

The following list are the abbreviations used:
Z - Prefix to keep all audit tables, triggers, & function at the end of any list
A - Audit
L - Log
F - Function
T - Trigger

Examples:
  ZAL - Z Audit Log
  ZALF - Z Audit Log Function
  ZALT - Z Audit Log Trigger

All table/type/function/trigger create/alter/drops are monitored and logged to a ddl audit log.
All calls to truncate a table are logged into the ddl audit log for the corresponding table.
All methods to create a table will trigger the creation of a corresponding set of audit
  table/function/triggers necessary to audit all activity on the triggering table.
When a table is dropped the functions and triggers monitoring that table are removed, the audit
   log it self is not removed to maintain audit integrity.
Truncate and Update to the audit log tables is not permitted through the used of triggers.
Delete on audit tables is only allowed when the 'ARCHIVE AUDIT LOG' descriptor is passed within
  the transaction. This is to allow stale records to be removed only under archival process.
  All other attempts to delete from an audit log table are prevented.
All insert/update/deletes performed on a monitored table will be logged.
Updates will only store the fields that change as part of the specific call.
Inserts and Deletes will keep the whole record.
Specific fields across all tables or for a specific table can be flagged as not being of
  significance requiring a audit log entry if they are the only field to change. If any other
  field also changes, all field that have changes will be recorded.
A passed in user id will be recorded with any log generated, If no user id was passed, -1 is used.
A passed in transaction id will be recorded with any log generated, if no transaction id was passed,
  then the built in postgresql transaction id will be used. The same transaction id should be used
  for all db calls generated form a single UI generated event to aid in associating all
  corresponding changes.
A passed in session sig will be recorded with any log generated, if no session sig is passed, null
  will be used.
An optional descriptor can be passed and will be recorded with any log generated if supplied.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-588

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


src/models/auditModelGenerator.js
```diff
@@ -0,0 +1,149 @@
...
+//   throw new Error(
+//     'Audit log only allows reading and inserting data,'
+//     + ' all modification and removal is not allowed.'
+//   );
+// };
+
+const tryJsonParse = (data) => {
+  const newData = data;
+  if (typeof data === 'object') {
+    Object.entries(data).forEach(([key, value]) => {
```
thewatermethod, line 18: something to be aware of: both ```null``` and an array are ```typeof "object"``` in JS

GarrettEHill: Updated to correctly handle null and array use case.

src/models/auditModelGenerator.js
```diff
@@ -0,0 +1,149 @@
...
+    },
+    data_id: { type: DataTypes.INTEGER },
+    dml_type: {
+      type: DataTypes.ENUM(...dmlType),
+      allowNull: false,
+    },
+    old_row_data: {
+      type: DataTypes.JSON,
+      allowNull: true,
+      get() { return tryJsonParse(this.getDataValue('old_row_data')); },
```
thewatermethod, line 73: sequelize doesn't automagically handle returning JSON for us?

GarrettEHill:
> 

How/where did you get/see that value?

GarrettEHill:
> > How/where did you get/see that value?
> 
> That's a column in my ZALActivityReports table, that was the first change I made

This should be addressed now with the better cookie processing

src/routes/apiDirectory.js
```diff
@@ -27,28 +27,29 @@ router.use(cookieSession);
 router.use(authMiddleware.unless({ path: [join('/api', loginPath)] }));
 
 router.use((req, res, next) => {
+  const getSessionSig = (cookie) => {
+    try {
+      if (cookie !== undefined && (typeof cookie === 'string' || cookie instanceof String)) {
+        const sessionSigs = cookie.split(' ').filter((s) => s.includes('session.sig='));
+        return sessionSigs.lenght > 0
```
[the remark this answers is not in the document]
GarrettEHill: Yep

src/routes/apiDirectory.js
```diff
@@ -20,9 +22,40 @@ authMiddleware.unless = unless;
...
 router.use(authMiddleware.unless({ path: [join('/api', loginPath)] }));
 
+router.use((req, res, next) => {
+  const getSessionSig = (cookie) => {
+    try {
+      if (cookie !== undefined && (typeof cookie === 'string' || cookie instanceof String)) {
+        const sessionSigs = cookie.split(' ').filter((s) => s.includes('session.sig='));
+        return sessionSigs.length > 0
+          ? sessionSigs[0].replace('session.sig=', '')
+          : '';
```
thewatermethod, line 36: Not sure if it'll mess things up but if I have more than one cookie from localhost:3000, the database entry for session.sig has a semicolon

GarrettEHill: hmmm, can you post your value as a comment here?

jasalisbury:
When I create a report and approve it I see entries in the tables I'd expect. But the `ActivityReportApprovers` table doesn't seem to get the userId or the transaction id properly.
Screenshot from 2022-02-14 09-23-27
Similar issue with the `ActivityReports` table but I imagine it is because of the sequelize hooks in that case (that set calculated status from submission status)
Screenshot from 2022-02-14 09-23-43

src/models/auditModelGenerator.js
```diff
@@ -0,0 +1,149 @@
...
+// };
+
+const tryJsonParse = (data) => {
+  let newData = data;
+  if (data) {
+    if (typeof data === 'object') {
+      Object.entries(data).forEach(([key, value]) => {
+        newData[key] = tryJsonParse(value);
+      });
+    } else if (typeof data === 'string') {
```
jasalisbury, line 22: Do we need to consider if the data is a number or bool?

src/models/auditModelGenerator.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+  return newData;
+};
+
+const addAuditTransactionSettings = async (sequelize, instance, options, type, name) => {
+  const loggedUser = httpContext.get('loggedUser') ? httpContext.get('loggedUser') : '';
+  const transactionId = httpContext.get('transactionId') ? httpContext.get('transactionId') : '';
+  const sessionSig = httpContext.get('sessionSig') ? httpContext.get('sessionSig') : '';
+  const auditDescriptor = httpContext.get('auditDescriptor') ? httpContext.get('auditDescriptor') : '';
+  if (loggedUser !== '' || transactionId !== '' || auditDescriptor !== '') {
```
jasalisbury, line 38: I assume the audit functions can handle not having any of the variables below set (if this `if` is false)

src/models/auditModelGenerator.js
```diff
@@ -0,0 +1,149 @@
+// Load in our dependencies
+import { Model, DataTypes } from 'sequelize'; // eslint-disable-line import/no-import-module-exports
+import httpContext from 'express-http-context'; // eslint-disable-line import/no-import-module-exports
+import { auditLogger } from '../logger'; // eslint-disable-line import/no-import-module-exports
+
+const dmlType = ['INSERT', 'UPDATE', 'DELETE'];
+
+// const exception = () => {
```
jasalisbury, line 8: Can remove

src/models/auditModelGenerator.js
```diff
@@ -0,0 +1,149 @@
...
+    const result = await sequelize.queryInterface.sequelize.query(
+      `SELECT
+        -- Type: ${type} Model: ${name}
+        set_config('audit.loggedUser', '${loggedUser}', TRUE) as "loggedUser",
+        set_config('audit.transactionId', '${transactionId}', TRUE) as "transactionId",
+        set_config('audit.sessionSig', '${sessionSig}', TRUE) as "sessionSig",
+        set_config('audit.auditDescriptor', '${auditDescriptor}', TRUE) as "auditDescriptor";`,
+      { transaction: options.transaction },
+    );
+    // console.log(JSON.stringify(result)); // eslint-disable-line no-console
```
jasalisbury, line 48: Can remove this as well

src/models/tests/auditModelGenerator.test.js
```diff
@@ -0,0 +1,363 @@
...
+          {
+            sequelize: db.sequelize,
+            createdAt: false,
+            updatedAt: false,
+            transaction: t,
+          },
+        );
+
+        await Test.sync({ force: true, alter: true, transaction: t })
+          .catch((err) => auditLogger.error(err));
```
jasalisbury, line 89: If we use `await` we need to wrap the call in a `try catch` instead of adding a `.catch` at the end of the promise chain. Adding the `.catch` at the end mixes two ways of dealing with promises, chaining vs await/async.

src/migrations/20220115000000-audit-system.js
```diff
@@ -0,0 +1,921 @@
...
+  corresponding changes.
+A passed in session sig will be recorded with any log generated, if no session sig is passed, null
+  will be used.
+An optional descriptor can be passed and will be recorded with any log generated if supplied.
+*/
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => queryInterface.sequelize.transaction(
+    async (transaction) => {
+      let syncronizer;
```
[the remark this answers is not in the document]
GarrettEHill: I was having race issues without forcing it to wait for the await.

src/models/auditModelGenerator.js
GarrettEHill: Defaults will be used for normal data flow, but this was also needed to support tests as there is no good way to set the httpContext for tests.

src/models/auditModelGenerator.js
GarrettEHill: its used in the model definition, should I just define it in the model?

src/migrations/20220115000000-audit-system.js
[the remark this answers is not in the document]
jasalisbury: Should be able to `await` without assigning the result to a variable

src/models/auditModelGenerator.js
jasalisbury: This commented out code is used in the model definition?

src/models/auditModelGenerator.js
GarrettEHill: Yes, the DB will use default values for anything not passed. This check is basically only useful for test cases, as the lack of a httpContext was making it not possible to  set the values manually.

GarrettEHill:
> When I create a report and approve it I see entries in the tables I'd expect. But the `ActivityReportApprovers` table doesn't seem to get the userId or the transaction id properly. Screenshot from 2022-02-14 09-23-27 Similar issue with the `ActivityReports` table but I imagine it is because of the sequelize hooks in that case (that set calculated status from submission status) Screenshot from 2022-02-14 09-23-43

Here is the issue. We do a very bad job about using transactions for complete operations. A user input should result in a single transaction to the database for all modifications of data across all tables. This would allow a complete roll back if any issue occurred. We don't currently do this.  We should use the same transaction for all requests of data as well to keep context correct across all interaction. 

To accomplish this we would need to take this as a project as a team as it looks like it could be alot of work. 

Without doing this work some of the db changes will be logged without the triggering user information. and most of the queries for data don't use a transaction. This will start to be a problem if it tries to reference data the is currently in an uncommitted transaction.

src/migrations/20220115000000-audit-system.js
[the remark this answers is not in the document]
GarrettEHill: I was having a race condition issue with several of the calls, forcing each to wait for output has corrected this.

src/models/auditModelGenerator.js
GarrettEHill: numbers and bools work as expected

jasalisbury:
> > When I create a report and approve it I see entries in the tables I'd expect. But the `ActivityReportApprovers` table doesn't seem to get the userId or the transaction id properly. Screenshot from 2022-02-14 09-23-27 Similar issue with the `ActivityReports` table but I imagine it is because of the sequelize hooks in that case (that set calculated status from submission status) Screenshot from 2022-02-14 09-23-43
> 
> Here is the issue. We do a very bad job about using transactions for complete operations. A user input should result in a single transaction to the database for all modifications of data across all tables. This would allow a complete roll back if any issue occurred. We don't currently do this. We should use the same transaction for all requests of data as well to keep context correct across all interaction.
> 
> To accomplish this we would need to take this as a project as a team as it looks like it could be alot of work.
> 
> Without doing this work some of the db changes will be logged without the triggering user information. and most of the queries for data don't use a transaction. This will start to be a problem if it tries to reference data the is currently in an uncommitted transaction.

Created and moved https://ocio-jira.acf.hhs.gov/browse/TTAHUB-657 to the top of the tech debt epic

src/migrations/20220115000000-audit-system.js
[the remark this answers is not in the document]
jasalisbury: I could see running into an issue when there was mixed usage between `await` and chaining (when exceptions were caught with `.catch` instead of `try/catch`) but there should be no difference between assigning the result of what is being `await`ed and not assigning. The `await` will wait for the promise to be resolved regardless of if the result is assigned to a variable or ignored.

src/models/auditModelGenerator.js
```diff
@@ -0,0 +1,149 @@
...
+    modelName: auditModelName,
+    createdAt: false,
+    updatedAt: false,
+  });
+  module.exports[auditModelName] = auditModel;
+  return auditModel;
+};
+
+const attachHooksForAuditing = (sequelize, model) => {
+  model.addHook(
```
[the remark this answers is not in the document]
GarrettEHill: I have replaced the hooks per model with permanent global hooks. all the types of hooks still exist.

jasalisbury:
Looks good to me, I say we get it in dev/sandbox and let people (as in people on our team) use the site a few days to make sure we don't hit any bugs and we generate audit data as we expect. I have a branch that is being used in dev but I imagine it won't be needed for too long, not sure if sandbox is currently in use

GarrettEHill:
Its still up and running in sandbox. With through more use we have identified several new issues:
[TTAHUB-614](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-614) - Transactions missing or misconfigured on some calls to the DB
[TTAHUB-657](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-657) - Ensure write operations are wrapped in a single transaction
[TTAHUB-670](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-670) - Change permissions for a user removes all permissions and then add permission rows back
[TTAHUB-671](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-671) - Autosave deleted existing data and then inserted it
[TTAHUB-672](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-672) - Need HSES nightly data changes metadata for audit log
[TTAHUB-674](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-674) - REFINE: Add reason for data change for audit log

What else is needed to complete the review and move this forward?

kryswisnaskas:
Nice work!

Do we have an interim plan for the handling of the production data transform script? Perhaps entering a separate ticket for it if not addressing here.

GarrettEHill:
> Nice work!
> 
> Do we have an interim plan for the handling of the production data transform script? Perhaps entering a separate ticket for it if not addressing here.

The script adds tables functions and triggers. It does not modify any of the current tables.

kryswisnaskas:
> > Nice work!
> 
> Do we have an interim plan for the handling of the production data transform script? Perhaps entering a separate ticket for it if not addressing here.
> 
> The script adds tables functions and triggers. It does not modify any of the current tables.

If we take a backup of the prod db and run `processData `on it, it will transform only the existing tables, right? If so, we end up with unobfuscated data in the audit tables. We could potentially truncate the audit tables as a way to still have the ability to refresh the data in the non-prod environments while working on a different approach.

GarrettEHill:
> > > Nice work!
> > 
> > 
> > Do we have an interim plan for the handling of the production data transform script? Perhaps entering a separate ticket for it if not addressing here.
> > The script adds tables functions and triggers. It does not modify any of the current tables.
> 
> If we take a backup of the prod db and run `processData `on it, it will transform only the existing tables, right? If so, we end up with unobfuscated data in the audit tables. We could potentially truncate the audit tables as a way to still have the ability to refresh the data in the non-prod environments while working on a different approach.

[TTAHUB-681](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-681) - processData script needs to be updated to obfuscate data in audit logs

kryswisnaskas (approved):

jasalisbury:
> [TTAHUB-614](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-614) - Transactions missing or misconfigured on some calls to the DB
> [TTAHUB-657](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-657) - Ensure write operations are wrapped in a single transaction
> [TTAHUB-670](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-670) - Change permissions for a user removes all permissions and then add permission rows back
> [TTAHUB-671](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-671) - Autosave deleted existing data and then inserted it
> [TTAHUB-672](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-672) - Need HSES nightly data changes metadata for audit log
> [TTAHUB-674](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-674) - REFINE: Add reason for data change for audit log
> [TTAHUB-681](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-681) - processData script needs to be updated to obfuscate data in audit logs

Do any of these need to be addressed before we move forward merging this PR?

GarrettEHill:
> > [TTAHUB-614](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-614) - Transactions missing or misconfigured on some calls to the DB
> > [TTAHUB-657](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-657) - Ensure write operations are wrapped in a single transaction
> > [TTAHUB-670](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-670) - Change permissions for a user removes all permissions and then add permission rows back
> > [TTAHUB-671](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-671) - Autosave deleted existing data and then inserted it
> > [TTAHUB-672](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-672) - Need HSES nightly data changes metadata for audit log
> > [TTAHUB-674](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-674) - REFINE: Add reason for data change for audit log
> > [TTAHUB-681](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-681) - processData script needs to be updated to obfuscate data in audit logs
> 
> Do any of these need to be addressed before we move forward merging this PR?

For the system to be functional no, for better data maybe, for PII TTAHUB-681 needs to be thought through.

GarrettEHill:
To temporarily disable audit tracking run the following at the start of the transaction:
`SET session_replication_role = replica;`
and to reenable tracking run the following at the end of the transaction:
`SET session_replication_role = DEFAULT;`

PR #770, opened 2022-02-08
Add goal status graph widget on backend

jasalisbury:
## Description of change

Needed to add two missing scopes to goals, region and recipient id. Added some test utilities to create/destroy goals to aid in testing. Check pointing at the backend so someone else can pickup the ticket and do the frontend as I won't have time to work on the frontend for a few days.

## How to test

Make sure tests make sense and pass.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-456

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/widgets/goalStatusGraph.js
+++ b/src/widgets/goalStatusGraph.js
@@ -3,14 +3,29 @@ import {
   Goal, Grant, Recipient, sequelize,
 } from '../models';
 
+export const GOAL_STATUS = {
+  NOT_STARTED: 'Not Started',
+  IN_PROGRESS: 'In Progress',
+  CLOSED: 'Closed',
+  CEASED: 'Ceased/Suspended',
+  DRAFT: 'Draft',
+};
+
+const STATUSES_TO_INCLUDE = [
+  GOAL_STATUS.NOT_STARTED,
+  GOAL_STATUS.IN_PROGRESS,
+  GOAL_STATUS.CLOSED,
+  GOAL_STATUS.CEASED,
+];
+
 export default async function goalStatusGraph(scopes) {
-  const goals = await Goal.findAll({
+  const goalsFromDb = await Goal.findAll({
     where: {
       [Op.and]: [
         scopes.goal,
         {
           status: {
-            [Op.not]: null,
+            [Op.in]: STATUSES_TO_INCLUDE,
           },
         },
       ],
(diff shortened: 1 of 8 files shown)


src/widgets/goalStatusGraph.js
```diff
@@ -0,0 +1,40 @@
...
+} from '../models';
+
+export default async function goalStatusGraph(scopes) {
+  const goals = await Goal.findAll({
+    where: {
+      [Op.and]: [
+        scopes.goal,
+        {
+          status: {
+            [Op.not]: null,
```
thewatermethod, line 13: Elsewhere, we've been using "needs status" if the goal was created in the hub and therefore doesn't have a status. Do you think we should use it here?

src/widgets/index.js
```diff
@@ -19,4 +20,5 @@ export default {
   topicFrequencyGraph,
   targetPopulationTable,
   frequencyGraph,
+  goalStatusGraph,
```
thewatermethod, line 23: Probably not a huge deal, but querying this API returns 238 goals with a status of "" on my DB. Possibly something weird from imported data?

Screen Shot 2022-02-09 at 7 59 19 AM

src/widgets/goalStatusGraph.js
jasalisbury: "needs status" and "draft" goals are ignored in the widget. I'm updating the backend to explicitly ignore draft goals

src/widgets/index.js
jasalisbury: Yeah those goals come in from smartsheet as empty strings. I'd imagine we'd want to update those goals to have a status that matches "needs status" (either null or explicitly "needs status")

thewatermethod (approved):

src/widgets/goalStatusGraph.js
```diff
@@ -3,14 +3,29 @@ import {
...
 export default async function goalStatusGraph(scopes) {
-  const goals = await Goal.findAll({
+  const goalsFromDb = await Goal.findAll({
     where: {
       [Op.and]: [
         scopes.goal,
         {
           status: {
-            [Op.not]: null,
+            [Op.in]: STATUSES_TO_INCLUDE,
```
thewatermethod, line 28: ✨

thewatermethod (approved):

PR #771, opened 2022-02-09
[TTAHUB-559] Update creator and collaborator data on legacy reports

thewatermethod:
## Description of change
Some legacy reports are missing creator or collaborator data in the database despite having that data present in the smartsheet data column. This script attempts to assign collaborators from the "otherSpecialists" field and creators from the "createdBy" field.

## How to test
Run the script. You may not have data in your DB that works (emails in the "imported" data in my database don't match any of the users), so you might have to create some.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-559

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


src/tools/updateLegacyCreatorsAndCollaborators.js
```diff
@@ -0,0 +1,116 @@
...
+        through: {
+          attributes: [],
+        },
+        required: false,
+      },
+    ],
+  });
+
+  // so here is where the filtering happens
+  const reports = rawReportData.filter((r) => r.collaborators.length === 0 || r.userId === null);
```
jasalisbury, line 112: Are we worried about a situation where a report has some collaborators but not all?

src/tools/updateLegacyCreatorsAndCollaborators.js
```diff
@@ -0,0 +1,116 @@
...
+    } else {
+      logger.info('No matching creator found');
+    }
+  }
+
+  return sequelize.transaction(async (transaction) => {
+    await Promise.all([
+      ...updatedCollaborators.map((c) => ActivityReportCollaborator.create({
+        activityReportId: id,
+        userId: c.id,
```
jasalisbury, line 74: If we need to handle a report that has some collaborators but not all we will have to make sure we don't dup `ActivityReportCollaborators`

thewatermethod:
Based on a conversation with Angela today and some concerns about trying to parse this data programmatically with a script, we've scaled back the expectation here. Basically - if there are emails in those fields, they should be added to the appropriate places. 

Not accounting for any of the weirdness around names, etc, at this point. That will likely either have to be a separate script or something that is addressed manually, either through the new UI that's been discussed a little, or through a situation where the a CSV is generated with all the needed ids and then run through a script.

package.json
```diff
@@ -71,23 +71,9 @@
...
     "changeReportStatus": "node ./build/server/tools/changeReportStatusCLI.js",
     "changeReportStatus:local": "./node_modules/.bin/babel-node ./src/tools/changeReportStatusCLI.js",
-    "updateLegacyPopulations": "node ./build/server/tools/updateLegacyPopulationsCLI.js",
-    "updateLegacyPopulations:local": "./node_modules/.bin/babel-node ./src/tools/uupdateLegacyParticipantsCLI.js",
-    "updateLegacyParticipants": "node ./build/server/tools/updateLegacyPopulationsCLI.js",
-    "updateLegacyParticipants:local": "./node_modules/.bin/babel-node ./src/tools/updateLegacyParticipantsCLI.js",
-    "importActivityReportDuration": "node ./build/server/tools/importActivityReportDuration.js",
-    "importActivityReportDuration:local": "./node_modules/.bin/babel-node ./src/tools/importActivityReportDuration.js",
-    "updateTopicNames": "node ./build/server/tools/updateTopicNamesCLI.js",
-    "updateTopicNames:local": "./node_modules/.bin/babel-node ./src/tools/updateTopicNamesCLI.js",
```
jasalisbury: :taco:

jasalisbury (approved):

PR #793, opened 2022-02-23
Csv limit increase

jasalisbury:
## Issue(s)

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/activityReports.js
+++ b/src/services/activityReports.js
@@ -791,7 +791,6 @@ async function getDownloadableActivityReports(where) {
           attributes: ['userId'],
           as: 'approvers',
           required: false,
-          separate: true,
           include: [
             {
               model: User,
@@ -831,11 +830,13 @@ export async function getAllDownloadableActivityReportAlerts(userId, filters) {
     [Op.and]: scopes,
     [Op.or]: [
       { // User is approver, and report is submitted or needs_action
-        [Op.or]: [
-          { submissionStatus: REPORT_STATUSES.SUBMITTED },
-          { calculatedStatus: REPORT_STATUSES.NEEDS_ACTION },
-        ],
-        '$approvers.userId$': userId,
+        [Op.and]: [{
+          [Op.or]: [
+            { calculatedStatus: REPORT_STATUSES.SUBMITTED },
+            { calculatedStatus: REPORT_STATUSES.NEEDS_ACTION },
+          ],
+          '$approvers.userId$': userId,
+        }],
       },
       { // User is author or collaborator, and report is approved
         [Op.and]: [
--- a/src/services/activityReports.js
+++ b/src/services/activityReports.js
@@ -691,7 +691,7 @@ export async function possibleRecipients(regionId) {
 }
 
 async function getDownloadableActivityReports(where) {
-  return ActivityReport.findAndCountAll(
+  return ActivityReport.findAll(
     {
       where,
       attributes: {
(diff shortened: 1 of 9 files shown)


AdamAdHocTeam (commented):
Initial testing looks good.

I was able to export 15k+ reports on SB and comparing them to DEV looked good.

My initial testing of the normal site functionality also looked good.

Will do some more testing later.

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
thewatermethod, line 1: 🎉

src/services/activityReports.js
```diff
@@ -21,12 +21,50 @@ import {
...
+
+      When DB objects have the extra sequelize "stuff" we run into memory
+      issues. When we get only data from the DB we don't run into memory
+      issues because all the sequelize "stuff" we don't need is garbage
+      collected at some point.
+
+      See https://github.com/sequelize/sequelize/issues/3897
+    */
+    const raw = JSON.parse(JSON.stringify(rows));
+    finalResult.push(...raw);
```
thewatermethod, line 58: not saying that there is a better way, but this is kind of an expensive operation and does not reflect well on sequelize for making us do it

thewatermethod:
Works for me! One thing I was thinking is that there is a substantial period after the button is clicked with little/no feedback for the users - no warning beforehand that this could take 2 minutes and little sign that something is working in the background. It may be a separate ticket, it may not, but it does seem like an issue.

src/services/activityReports.js
jasalisbury: Yeah definitely. Especially because the issue in the comments was created in 2015

jasalisbury:
> Works for me! One thing I was thinking is that there is a substantial period after the button is clicked with little/no feedback for the users - no warning beforehand that this could take 2 minutes and little sign that something is working in the background. It may be a separate ticket, it may not, but it does seem like an issue.

Yeah good point. I tried to diagnose why it takes so long. Locally it seems DB calls took up around 20 seconds before they were chunked up, so might be different now. I tried looking at indexes but the query was too complex for me to make sense of what was going on. Now that models are split into separate calls it would be easier to diagnose slowdowns in SQL. The biggest slowdown in the `trasnformer` is removing HTML tags from two fields. I think the solution there would be to store an HTML and non-HTML version of those fields when they are saved.

jasalisbury:
I've added the "loading" indicator whenever a CSV is being downloaded

jasalisbury:
> The loader is definitely an improvement! Long term, I think we may still want to loop this into a UI discussion and have these big reports built async.

Yeah I think going async would be where we will take this in the future (assuming it is important enough to work on!)

> I was able to crash sandbox by 1) exporting all reports 2) clicking any link while the reports are downloading. Does this mean that multiple people using the site while a large report is being downloaded would crash an instance? Perhaps these giant site-crashing reports aren't downloaded enough to make this a concern?

I was able to browse around while the report was being downloaded. Sandbox is more likely to crash because it is set to 1 gb memory (vs 2 gb in prod) and there is a single instance vs 3 in prod.

thewatermethod:
> I was able to browse around while the report was being downloaded. Sandbox is more likely to crash because it is set to 1 gb memory (vs 2 gb in prod) and there is a single instance vs 3 in prod.

Good, I mean, as long as we're aware and thinking about it, it's probably not a dealbreaker given those facts about prod

AdamAdHocTeam (approved):
Looks good.

Its a shame Sequelize has us jumping through hoops for memory.

PR #796, opened 2022-02-25
[TTAHUB-617] Fix legacy imported target population

AdamAdHocTeam:
## Description of change

This fixes the "Affected by Child Welfare involvement" issue on the imported json (what we display).

## How to test

Run the script all legacy reports that had "Affected by Child Welfare involvement" should now have ""Affected by Child Welfare Involvement".

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-617

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/tools/updateChildInvolvementLegacyTargetPopulation.js
+++ b/src/tools/updateChildInvolvementLegacyTargetPopulation.js
@@ -1,0 +1,38 @@
+import { Op } from 'sequelize';
+import {
+  ActivityReport,
+} from '../models';
+import { auditLogger } from '../logger';
+
+export default async function updateChildInvolvementLegacyTargetPopulation() {
+  const legacyReports = await ActivityReport.findAll({
+    attributes: [
+      'id',
+      'imported',
+    ],
+    where: {
+      [Op.and]: [
+        {
+          imported: { [Op.not]: null },
+        },
+        {
+          imported: {
+            targetPopulations: {
+              [Op.like]: '%Affected by Child Welfare involvement%',
+            },
+          },
+        },
+      ],
+    },
+  });
+
+  auditLogger.info(`${legacyReports.length ? legacyReports.length : 0} reports with affected data found...`);
+  return Promise.all(legacyReports.map(async (report) => {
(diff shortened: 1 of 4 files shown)


thewatermethod:
Can we double check that we don't want to fix whats in the targetPopulations column rather than modify the imported data? I think its preferable to leave the imported data alone

src/tools/updateChildInvolvementLegacyTargetPopulation.js
```diff
@@ -0,0 +1,38 @@
+import { Op } from 'sequelize';
+import {
+  ActivityReport,
+} from '../models';
+import { auditLogger } from '../logger';
+
+export default async function updateChildInvolvementLegacyTargetPopulation() {
```
thewatermethod, line 7: this
```js
   where: {
      [Op.and]: [
        {
          imported: { [Op.not]: null },
        },
        {
          imported: {
            targetPopulations: {
              [Op.like]: '%Affected by Child Welfare involvement%',
            },
          },
        },
      ],
```

can probably just be

```js
   where: {     
          imported: {
            targetPopulations: {
              [Op.like]: '%Affected by Child Welfare involvement%',
            },
          },
   },    
 ```

thewatermethod:
oh disregard my previous question I guess, I thought we were trying to avoid updating imported data but I see that you've already updated the regular DB column so this must be an additional request

jasalisbury (approved):
:rocket:

jasalisbury:
We should add acceptance criteria to these one off script tickets to delete the script after they are ran so we don't forget

kryswisnaskas:
Looks good as far as updating of the values in the db! 👍 

Can we also make sure that the UI for the legacy reports displays the new data? For example
`activity-reports/legacy/R06-AR-000464`

It would be ideal to source the data from the target populations column, but in the past we also modified the `imported `column to accomplish the above. 🦖

src/tools/updateChildInvolvementLegacyTargetPopulation.js
AdamAdHocTeam: This has been changed. I guess I was worried that the null value would break the where clause.

PR #799, opened 2022-02-28
[TTAHUB-374] save progress to local storage when creating an AR

thewatermethod:
## Description of change
When creating an activity report, data is saved to local storage. If a report is not retrieved or saved from the backend, you can proceed through the report as normal. There are messages that attempt to explain this to the user. In addition, submit report behavior has been modified so that if the backend fails to save or retrieve a report, submission is disabled until such time as the report can be retrieved.

In my conception, this is meant to 1) mitigate some issues with low/no internet and 2) provide peace of mind to our users that there is an additional layer of saving happening, which is **somewhat** protected against any weird errors they might see in the UI.

I also spent a while in the unit tests, so you'll see a few changes related to cleaning up the console.

Before deployment, the language on these error messages will have to be workshopped amongst the larger group since this is basically whatever fell out of my fingers at the time.

## How to test
Create a report as normal. It should work! If you check while creating a report in your local storage, you should see that the report has been "backed up" there. It is cleaned up once its in submitted or approved state (there are certain browsers and environments that limit the amount of local storage space, so cleanup is necessary).

For additional testing, you can add errors and return statements to the appropriate fetchers (getReport and saveReport) and create a report (the report will need to have an id first, so probably save one and then add the errors, and then continue working). You should be able to, and if you reconnect it should be fairly seemless. 

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-374

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


frontend/src/NetworkContext.js
```diff
@@ -0,0 +1,5 @@
+import { createContext } from 'react';
+
+const NetworkContext = createContext();
+
```
thewatermethod, line 4: I know I've been creating/using a lot of contexts. I swear I'm not context happy, there just wasn't an easy path on how to pass down "the backend failed" through the navigator.

frontend/src/pages/ActivityReport/index.js
```diff
@@ -169,7 +211,23 @@ function ActivityReport({
             || report.calculatedStatus === REPORT_STATUSES.NEEDS_ACTION);
 
         updateAdditionalData({ recipients, collaborators, availableApprovers });
-        updateFormData(report);
+
+        let shouldUpdateFromNetwork = true;
+
```
thewatermethod, line 216: I'm not sure if there is more elegant way to handle this without recreating the git language.

frontend/src/pages/ActivityReport/testHelpers.js
```diff
@@ -0,0 +1,68 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import React from 'react';
```
thewatermethod, line 2: This file isn't in _tests_ because if it was it would match our test regex, which I didn't want to mess with. The eslint warning says that its all right these packages are in devDependencies and not dependencies in package.json.

The stuff has been pulled out so I could write a smaller test file and not get weird run off from all kinds of other tests.

frontend/src/testHelpers.js
```diff
@@ -11,3 +7,18 @@ export const withText = (text) => (content, node) => {
 
   return nodeHasText && childrenDontHaveText;
 };
+
```
thewatermethod, line 10: cherry picked from #762

jasalisbury:
Been using it a bit, seem to work well, haven't run into anything major! A few questions/concerns before I dive into the code, so I apologize if these questions are answered in code:
* How are we going to clean up local storage? Should we delete the entries after the report has been submitted? What about collaborators that don't submit the report they worked on, how is that report cleared?
* I like the banner at the top telling the user the report was loaded locally, but can we add/update the "report was last saved" message to indicate the report didn't make it up to the server?
* Do we need to handle the case where the user lost internet connection before saving a report for the first time as a special case? I was able to enter info and move pages while my report was still "new" and when I regained connection everything worked as expected. However I'm worried about what happens when the user leaves the report while it is still "new". They would never be able to return to the report to submit it to the server
* When I create a new report and click `Save draft` on the first page I don't get an entry in local storage for the report. I have to click `Save & Continue` or use the navigator to get an entry in local storage
* Could use a message next to the `submit` button when it is disabled explaining why
* Similar to above (but more a general question then something we definitely want to do), should we add messages explaining why there are no `Goals` available for selection if the user has no internet connection? I guess same issue with the list of grants, approvers, and collaborators

thewatermethod:
> How are we going to clean up local storage? Should we delete the entries after the report has been submitted? What about collaborators that don't submit the report they worked on, how is that report cleared?

The report is cleaned from local storage if it gets submitted or approved, or is viewed in one of those states. Not sure if there's anything about that collaborator case -- it might not be cleaned up if they never view the report when submitted.  I'm honestly not sure what the fix would be for that. It would probably have to be a user meta table of some kind. I think that I should probably add a cleanup hook to the approved report page as well, just for extra cleaning power.

> I like the banner at the top telling the user the report was loaded locally, but can we add/update the "report was last saved" message to indicate the report didn't make it up to the server?

Yes, I can do that.

> Do we need to handle the case where the user lost internet connection before saving a report for the first time as a special case? I was able to enter info and move pages while my report was still "new" and when I regained connection everything worked as expected. However I'm worried about what happens when the user leaves the report while it is still "new". They would never be able to return to the report to submit it to the server

This is an interesting problem. I'm a bit worried about ensuring that a report saved without an id could lead to annoying problems. I can try saving the "new" report and see if we clear out the "new" report data when the report is granted an ID

> When I create a new report and click Save draft on the first page I don't get an entry in local storage for the report. I have to click Save & Continue or use the navigator to get an entry in local storage

Ok, must be a bug, will fix

> Could use a message next to the submit button when it is disabled explaining why

I can add this

> Similar to above (but more a general question then something we definitely want to do), should we add messages explaining why there are no Goals available for selection if the user has no internet connection? I guess same issue with the list of grants, approvers, and collaborators

Yes, probably. Not sure if it would be apparent already given they likely have alerts notifying them that the network is having issues, but probably there is always room for more communication

jasalisbury:
> The report is cleaned from local storage if it gets submitted or approved, or is viewed in one of those states. Not sure if there's anything about that collaborator case -- it might not be cleaned up if they never view the report when submitted. I'm honestly not sure what the fix would be for that. It would probably have to be a user meta table of some kind. I think that I should probably add a cleanup hook to the approved report page as well, just for extra cleaning power.

Should we clear out local storage for the report when it is submitted for approval? I doubt I would personally visit a report again after submitting it for approval (assuming it was approved).

> > I like the banner at the top telling the user the report was loaded locally, but can we add/update the "report was last saved" message to indicate the report didn't make it up to the server?

> Yes, I can do that.

I see both "saved" messages when I'm able to save to the server. My thinking (and this may not be the way to do it!) is to only show the "saved locally" message when the report fails to hit the server. Another thing is I see the message saying the report was saved to the network when the server is off.

> > Do we need to handle the case where the user lost internet connection before saving a report for the first time as a special case? I was able to enter info and move pages while my report was still "new" and when I regained connection everything worked as expected. However I'm worried about what happens when the user leaves the report while it is still "new". They would never be able to return to the report to submit it to the server

> This is an interesting problem. I'm a bit worried about ensuring that a report saved without an id could lead to annoying problems. I can try saving the "new" report and see if we clear out the "new" report data when the report is granted an ID

Maybe we shouldn't save "new" reports to local storage?

thewatermethod:
> Should we clear out local storage for the report when it is submitted for approval? I doubt I would personally visit a report again after submitting it for approval (assuming it was approved).

Good call, local storage is now cleared out on successful submit

> I see both "saved" messages when I'm able to save to the server. My thinking (and this may not be the way to do it!) is to only show the "saved locally" message when the report fails to hit the server. Another thing is I see the message saying the report was saved to the network when the server is off.

My thought was that this was primarily a trust building exercise, so maybe we really hammer the users over the head that this is being saved in multiple ways. Probably once we're happy with the functionality, it would be a question for the whole group or at least the designers how many alerts and where and of what flavor.

> Maybe we shouldn't save "new" reports to local storage?

That was my original approach, but I think it works ok now that I've switched it around. "New" reports are cleared out from local storage when they get an id.

jasalisbury:
Probably got lost in my comment but I'm seeing the "successfully saved to the server" message when it is only getting saved locally (when I have my server off).

I think we might want to have a last save (and probably created at?) stored in local storage in case we want to clean out reports based on age (for example maybe we clear out reports older then 2 months, or have a popup to ask the user about it?). Don't think we'd use it directly now but would be nice to have the ability in the future.

frontend/src/hooks/useARLocalStorage.js
```diff
@@ -0,0 +1,46 @@
...
+      )
+    ) {
+      toSave = false;
+    }
+
+    setSaveReport(toSave && localStorageAvailable);
+  }, [key, localStorageAvailable, setStoredValue, storedValue]);
+
+  return [storedValue, (v) => {
+    const savedToStorage = new Date().toISOString();
```
jasalisbury, line 34: :+1: didn't see this while saving the report, disregard my comment about "last saved" time

frontend/src/pages/ActivityReport/Pages/goalsObjectives.js
```diff
@@ -57,12 +59,19 @@ const GoalsObjectives = () => {
       )}
       {showGoals
         && (
-        <Fieldset className="smart-hub--report-legend margin-top-4" legend="Goals and objectives">
-          <div id="goals-and-objectives" />
-          <GoalPicker
-            availableGoals={availableGoals}
-          />
-        </Fieldset>
+        <>
```
jasalisbury, line 62: Why do we need a fragment here?

frontend/src/pages/ActivityReport/index.js
```diff
@@ -32,6 +41,9 @@ import {
   reviewReport,
   resetToDraft,
 } from '../../fetchers/activityReports';
+import useLocalStorage from '../../hooks/useLocalStorage';
+import NetworkContext from '../../NetworkContext';
+import { storageAvailable } from '../../hooks/helpers';
```
jasalisbury, line 46: Seems like we are checking if storage is available in a few places, any way we can cut it down to a single place? (if not that is fine!)

frontend/src/pages/ActivityReport/index.js
```diff
@@ -264,38 +339,53 @@ function ActivityReport({
 
   const onSave = async (data) => {
     const approverIds = data.approvers.map((a) => a.User.id);
-    if (reportId.current === 'new') {
-      const { startDate, endDate, ...fields } = data;
-      let startDateToSave = startDate;
-      if (startDateToSave === 'Invalid date' || startDateToSave === '' || !moment(startDateToSave, 'MM/DD/YYYY').isValid()) {
-        startDateToSave = null;
-      }
+    setConnectionActive(false);
```
jasalisbury, line 342: Should we only set this to false on an error?

frontend/src/pages/ActivityReport/Pages/Review/Submitter/Draft.js
```diff
@@ -146,7 +154,15 @@ const Draft = ({
         >
           Save Draft
         </Button>
-        <Button type="submit">Submit for approval</Button>
+        { !connectionActive && (
+        <Alert>
+          There&#39;s an issue with your network connection,
+          and you can&#39;t submit your report right now.
```
jasalisbury, line 160: Think you need a `{' '}` after the line above.

jasalisbury (approved):
I'd get some reviews from other engineers, but this LGTM. A few questions that I have that shouldn't hold up this PR for the team as a whole:
* Should we only show a single "successfully save" in the case the user can save reports to the server?
* Should we use a different color then the green "success" color when the report is only saved to local storage? I know we don't want to spook our users making them think their report isn't saved but I think it is important that they understand when the report they saved hasn't reached the server and can't be seen by other users.

thewatermethod:
FYI - this is with the UI team, they are working on the precise messaging

frontend/src/pages/ActivityReport/index.js
```diff
@@ -207,8 +259,27 @@ function ActivityReport({
           && (report.calculatedStatus === REPORT_STATUSES.DRAFT
             || report.calculatedStatus === REPORT_STATUSES.NEEDS_ACTION);
         updateAdditionalData({ recipients, collaborators, availableApprovers });
+
+        let shouldUpdateFromNetwork = true;
+
+        if (formData && savedToStorage) {
```
AdamAdHocTeam, line 265: Just add a comment here explaining that this is checking if the last local storage save is older than the report last updated at.

frontend/src/hooks/useARLocalStorage.js
```diff
@@ -0,0 +1,46 @@
...
+        || storedValue.calculatedStatus === REPORT_STATUSES.SUBMITTED
+      )
+    ) {
+      toSave = false;
+    }
+
+    setSaveReport(toSave && localStorageAvailable);
+  }, [key, localStorageAvailable, setStoredValue, storedValue]);
+
+  return [storedValue, (v) => {
```
AdamAdHocTeam, line 33: As we discussed lets give 'v' a meaningful name. 

It also might be cleaner to move the second return param that's using 'v' to its own function.

frontend/src/pages/ActivityReport/index.js
```diff
@@ -139,22 +161,41 @@ function ActivityReport({
...
-  const [formData, updateFormData] = useState();
-  const [initialAdditionalData, updateAdditionalData] = useState({});
+
+  const [lastSaveTime, updateLastSaveTime] = useState(null);
+
+  const [formData, updateFormData, localStorageAvailable] = useARLocalStorage(
+    LOCAL_STORAGE_DATA_KEY(activityReportId), null,
+  );
+
+  const savedToStorage = formData ? formData.savedToStorage : null;
```
AdamAdHocTeam, line 175: This sounds like a boolean might want to give it a better name.

frontend/src/pages/ActivityReport/index.js
```diff
@@ -259,7 +332,7 @@ function ActivityReport({
     return <Redirect to="/" />;
   }
 
-  if (error) {
+  if (error && !formData) {
```
AdamAdHocTeam, line 335: Maybe a comment.

frontend/src/pages/ActivityReport/index.js
```diff
@@ -232,7 +303,9 @@ function ActivityReport({
 
         updateError();
       } catch (e) {
-        updateError('Unable to load activity report');
+        setConnectionActive(false);
```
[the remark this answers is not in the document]
AdamAdHocTeam: We found: https://stackoverflow.com/questions/62943328/how-to-check-if-the-user-has-internet-connection-in-react

AdamAdHocTeam (commented):
We need to make sure that when a user deletes a draft report from the AR Landing we clear that report from local storage. I could see this number of deleted reports accumulating over time.

Another minor thing I noticed. If I create a new report makes some changes to the first page. Then click 'Save Draft' and leave. There is no report created in local storage until I revisit the saved report.

frontend/src/pages/ActivityReport/index.js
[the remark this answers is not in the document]
AdamAdHocTeam: We probably need to handle this as a failure to retrieve goals seems to have tripped the no internet connection.

thewatermethod:
@AdamAdHocTeam Think I've addressed all those issues if you want to take another look

thewatermethod:
@AdamAdHocTeam Think that the issue you found with inconsistent behavior with save to draft/save and continue has been fixed

jasalisbury:
What else needs to happen to get this PR merged?

thewatermethod:
Hoping to merge it today, messaged Patrice to confirm the new messaging works

GarrettEHill (approved):

PR #808, opened 2022-03-08
[TTAHUB-611/612/615] Create goal & objectives form

thewatermethod:
## Description of change

- Add create goal form under the goals & objectives tab
- Backend handlers
- Only users with read/write or approve permissions can create goals, and only recipients with active grants can have goals created 

Note that this PR does not include behavior when returning to landing page - it should show the recently created goals at the top of the table (not sure how I'm gonna pull that off anyways)

## How to test
Run migrations. Create multiple goals/objectives for a recipient. Make sure validation, creation works and makes sense.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-611

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


thewatermethod:
converting to draft so I can check over the big merge I just did

frontend/src/components/CreateGoal/Form.js
```diff
@@ -0,0 +1,212 @@
...
+      { canAddNewObjective ? (
+        <div className="margin-top-4">
+          <PlusButton onClick={onAddNewObjectiveClick} text="Add new objective" />
+        </div>
+      ) : null }
+    </div>
+  );
+}
+
+Form.propTypes = {
```
AdamAdHocTeam, line 158: That's a lot of props.

frontend/src/components/CreateGoal/Form.js
```diff
@@ -0,0 +1,212 @@
...
+  errors,
+  validateGoalName,
+  validateEndDate,
+  validateGrantNumbers,
+  objectives,
+  setObjectives,
+  setObjectiveError,
+  topicOptions,
+}) {
+  const onUpdateText = (e) => setGoalName(e.target.value);
```
AdamAdHocTeam, line 37: Not sure if you need this as a stand alone func.

frontend/src/components/CreateGoal/ObjectiveForm.js
```diff
@@ -0,0 +1,139 @@
...
+  errors,
+  topicOptions,
+}) {
+  // the parent objective data from props
+  const { title, topics, resources } = objective;
+
+  // onchange handlers
+  const onChangeTitle = (e) => setObjective({ ...objective, title: e.target.value });
+  const onChangeTopics = (newTopics) => setObjective({ ...objective, topics: newTopics });
+  const setResources = (newResources) => setObjective({ ...objective, resources: newResources });
```
AdamAdHocTeam, line 32: Again not sure if these could just be moved inline.

AdamAdHocTeam (commented):
If I create a new goal and fill out all information click 'Save & Continue'. Then on the redirect page I click the three ellipsis and 'Edit'. The page that I'm redirected back to is missing the  'Recipient Grant Numbers'. If I add an objective (not realizing the blank field) and hit save I get a 500 error.

thewatermethod:
> If I create a new goal and fill out all information click 'Save & Continue'. Then on the redirect page I click the three ellipsis and 'Edit'. The page that I'm redirected back to is missing the 'Recipient Grant Numbers'. If I add an objective (not realizing the blank field) and hit save I get a 500 error. I was unable to recover from this error ie had to leave the page.

Oh, did you run the migrations?

frontend/src/components/CreateGoal/Form.js
thewatermethod: Sure is

AdamAdHocTeam:
> > If I create a new goal and fill out all information click 'Save & Continue'. Then on the redirect page I click the three ellipsis and 'Edit'. The page that I'm redirected back to is missing the 'Recipient Grant Numbers'. If I add an objective (not realizing the blank field) and hit save I get a 500 error. I was unable to recover from this error ie had to leave the page.
> 
> Oh, did you run the migrations?

Yes I have run the migrations I restored DB and ran again to be sure.

Seems to still be an issue.

1. Create a new goal Save & Continue.
2. Click three ellipses to edit goal.
3. Add some objectives 
4. Click Save & Continue
5. Get 500 Error

src/policies/goals.js
```diff
@@ -0,0 +1,38 @@
...
+  canDelete() {
+    return this.canWriteInRegion();
+  }
+
+  canCreate() {
+    return this.canWriteInRegion();
+  }
+
+  canWriteInRegion() {
+    const permissions = _.find(
```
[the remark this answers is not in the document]
thewatermethod: Just trying to keep it consistent with the way the other permissions are modeled

thewatermethod:
> Seems to still be an issue.

Found the bug, seems to be fixed

jasalisbury:
Works well! Two observations

1) Clicking "save draft" kicks me back to the goal list. Do we want to do that or just hit the server with a save and keep the user on the form?
2) When I add objectives to a "Not Started" goal (one I click "submit" for) I see zero objectives listed for the goal, which is because there are no activity reports for the goal. But I could see a user getting confused why there are no objectives for a goal when they just added objectives to that goal

GarrettEHill:
> 2. When I add objectives to a "Not Started" goal (one I click "submit" for) I see zero objectives listed for the goal, which is because there are no activity reports for the goal. But I could see a user getting confused why there are no objectives for a goal when they just added objectives to that goal
 This seems like a good idea, but to be done right will require additional work. Should a new jira be opened to address this?

jasalisbury:
> This seems like a good idea, but to be done right will require additional work. Should a new jira be opened to address this?
> 
> > 2. When I add objectives to a "Not Started" goal (one I click "submit" for) I see zero objectives listed for the goal, which is because there are no activity reports for the goal. But I could see a user getting confused why there are no objectives for a goal when they just added objectives to that goal

Yeah I'm not sure what the solution is, something we need to bring up at a standup

thewatermethod:
> Clicking "save draft" kicks me back to the goal list. Do we want to do that or just hit the server with a save and keep the user on the form?

I changed it to behave consistently. Will have to double check with the group that it works as expected since I've heard a couple of different things.

> When I add objectives to a "Not Started" goal (one I click "submit" for) I see zero objectives listed for the goal, which is because there are no activity reports for the goal. But I could see a user getting confused why there are no objectives for a goal when they just added objectives to that goal

I think I've fixed this by making the join to AR's not required in the backend route. I can't think of any way that has unexpected side effects, but maybe there was a reason it wasn't like that to begin with.

thewatermethod:
Note I also wrote another Jira to handle the landing page behavior as far as showing the goals: https://ocio-jira.acf.hhs.gov/browse/TTAHUB-740

src/policies/goals.js
```diff
@@ -0,0 +1,38 @@
+/*
+  This class handles authorization for Activity Reports. In express handlers this
```
jasalisbury, line 2: Will want to remove this comment

src/routes/goals/handlers.js
```diff
@@ -30,3 +59,36 @@ export async function changeGoalStatus(req, res) {
...
 }
+
+export async function deleteGoal(req, res) {
+  try {
+    const { goalId } = req.params;
+    const { regionId } = req.body;
+
+    const user = await userById(req.session.userId);
+
+    const policy = new Goal(user, {
```
jasalisbury, line 70: I would expect this to be passed an instance of a goal. Also would allow us to not care about regionId in the body at all, which cleans up the call (you only need the goal id to delete). Plus someone could call the API with a goal id from a different region and be able to delete the goal

thewatermethod: Do you think I should put the sequelize logic for looking up the goal region in this file?

thewatermethod: right right, but goals don't have regions, so I'd need to get the grants attached, loop through and check that at least one matches the user's permissions. But it sounds like maybe the services is the best place for that, not the policy

jasalisbury: Oh gotcha. Yeah I'd probably add something to the goals service going from the goal to all grants to grab all regions. This will change in the future to be a bit nicer once a goal has a single grant

frontend/src/components/CreateGoal/Form.css
```diff
@@ -0,0 +1,8 @@
+/** text inputs **/
+.ttahub-create-goals-form .usa-error-message + :invalid, 
+/** the <Select /> component **/
+.ttahub-create-goals-form .usa-error-message + .usa-select, 
+/** URL fields are wrapped in a div **/
+.ttahub-create-goals-form .usa-error-message + .display-flex :invalid  {
+    border: 2px solid #b50909;
```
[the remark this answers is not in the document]
thewatermethod: For your first and last question, I think/assume the answer is a yes, and that will happen when we start working on the updates to the AR shortly. We should definitely keep an eye everything and make sure my assumption is correct

As far as the red line here, it should have that, so I've added it in (thanks for catching that)

src/policies/goals.js
```diff
@@ -1,38 +1,67 @@
...
+        && permission.regionId === region),
     );
-    return !_.isUndefined(permissions);
+    return !isUndefined(permissions);
+  }
+
+  // this expects goal to have been found with associated data, specifically
+  // goalByIdWithActivityReportsAndRegions in services/goals
+  // you can see the structure expected in the conditions below
+  isOnApprovedActivityReports() {
```
jasalisbury, line 61: Not for now, but I could see this method being moved as a virtual method on the goal model because I assume it is a question we are going to frequently ask about goals.

src/policies/goals.js
```diff
@@ -1,38 +1,67 @@
...
   canDelete() {
-    return this.canWriteInRegion();
+    if (this.isOnApprovedActivityReports()) {
+      return false;
+    }
+
+    // I assume that if you can delete in ANY of the goals regions,
+    // than you can delete it
+    const regions = this.goal.grants.map((grant) => grant.regionId);
+    return regions.reduce((previous, region) => {
```
jasalisbury, line 22: Can we use `regions.every` here instead of the reducer? If so it might be a bit easier to understand what is going on, and probably can remove the explicit true check on line 24. Not a big deal either way, especially since this code is going to be updated very soon after the schema update.

frontend/src/components/CreateGoal/ObjectiveForm.js
```diff
@@ -0,0 +1,139 @@
...
+        <Textarea id="objectiveTitle" name="objectiveTitle" required value={title} onChange={onChangeTitle} onBlur={validateObjectiveTitle} />
+      </FormGroup>
+      <FormGroup error={errors[OBJECTIVE_FORM_FIELD_INDEXES.TOPICS].props.children}>
+        <Label htmlFor="topics">
+          Topics
+          {' '}
+          <span className="smart-hub--form-required font-family-sans font-ui-xs">*</span>
+        </Label>
+        {errors[OBJECTIVE_FORM_FIELD_INDEXES.TOPICS]}
+        <Select
```
jasalisbury, line 94: In the activity report we have the topics multi-select stay open after a selection. Do we want to match that functionality here?

jasalisbury (approved):
LGTM Nice work! Just have a few minor questions

src/policies/goals.js
thewatermethod: I think we want ```regions.some```, right? What I'm asking is, if a user can delete in any of the regions then should they be able to delete here? Or would they need permissions in all a goals regions? (I know its almost a non-issue given how few goals transcend 1 region, but just checking in case)

jasalisbury: Yeah you are correct it doesn't really matter because this code isn't making it to release with the schema updates. But I'd typically land on the side of not letting users do things when there are permission questions like this that could go either way. Rather be more constrained then more permissive when letting users mutate data.

frontend/src/components/GoalsTable/GoalsTableHeader.js
```diff
@@ -0,0 +1,112 @@
...
+}) {
+  const { user } = useContext(UserContext);
+  const hasButtonPermissions = canEditOrCreateGoals(user, parseInt(regionId, DECIMAL_BASE));
+
+  return (
+    <div className="desktop:display-flex">
+      <div className="desktop:display-flex flex-1 desktop:padding-top-0 padding-top-2">
+        <h2 className="font-body-lg margin-left-2 margin-right-1 margin-y-3">{title}</h2>
+        { hasActiveGrants && hasButtonPermissions ? (
+          <span className="smart-hub--table-controls desktop:margin-x-2 desktop:margin-y-0 margin-2 display-flex flex-row flex-align-center">
```
jasalisbury, line 43: We should create a new feature flag for creation of goals since we don't want goal creation to release at the same time as view goals

thewatermethod: I didn't realize that was the case... I wonder if we could just use the existing feature flag and the wrapping around the parts as we release them.

thewatermethod: either way, if its all right with you, I'm getting ready to merge this in and will write another Jira/PR to make some of more recent changes to the mockup and can include the feature flag on that one if needed

jasalisbury: yeah we could probably use the same flag, either way up to you. And putting it in the next PR is fine.

src/policies/goals.js
thewatermethod: thats a smart idea. J/C, would we have associate the activity reports directly with goals? I don't think you can go like ```this.objectives[0].activityReports```. Is it weird to do a find all within a virtual field? Feels weird but maybe I should get over it.

GarrettEHill (approved):

PR #815, opened 2022-03-11
[TTAHUB-707] Update DB schema for the create goals MVP

thewatermethod:
## Description of change

In order to support the new create goals form, some changes to the DB schema will be necessary.

Referenced this doc when writing the new migrations: https://docs.google.com/drawings/d/1xIDlhbKtZ6UiujSX0eqiELvWSPiYRKf-h44jqhvDe8k/edit

## How to test
Make sure changes are logical and line up with the long term plan 

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-707

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20220311135140-update-topics-table.js
+++ b/src/migrations/20220311135140-update-topics-table.js
@@ -39,40 +39,13 @@ const TOPICS = [
 ];
 
 module.exports = {
-  up: async (queryInterface, Sequelize) => queryInterface.sequelize.transaction(
+  up: async (queryInterface) => queryInterface.sequelize.transaction(
     async (transaction) => {
-    // create an "isSelectable" column, denoting that this topic
-    // is in active use
-      await queryInterface.addColumn(
-        'Topics',
-        'isSelectable',
-        {
-          type: Sequelize.DataTypes.BOOLEAN,
-        },
-        { transaction },
-      );
-
-      // set all inactive topics to not selectable
-      await queryInterface.bulkUpdate('Topics', {
-        isSelectable: false,
-      }, {
-        name: {
-          [Sequelize.Op.notIn]: TOPICS,
-        },
-      }, { transaction });
-
       // delete all existing active topics
-      // (couldn't figure out how to do a targeted bulk update with
-      // the sequelize query interface object)
-      await queryInterface.bulkDelete('Topics', {
-        name: {
-          [Sequelize.Op.in]: TOPICS,
-        },
-      }, { transaction });
+      await queryInterface.bulkDelete('Topics', {}, { transaction });
 
       const topicsToInsert = TOPICS.map((topic) => ({
         name: topic,
-        isSelectable: true,
         createdAt: new Date(),
         updatedAt: new Date(),
       }));
@@ -84,7 +57,7 @@ module.exports = {
     },
   ),
 
-  down: async (queryInterface) => {
-    await queryInterface.removeColumn('Topics', 'isSelectable');
+  down: async () => {
+  // out of luck here I think
   },
 };
--- a/src/migrations/20220311135140-update-topics-table.js
+++ b/src/migrations/20220311135140-update-topics-table.js
@@ -1,0 +1,90 @@
+const TOPICS = [
+  'Behavioral / Mental Health / Trauma',
+  'Child Assessment, Development, Screening',
+  'CLASS: Classroom Organization',
+  'CLASS: Emotional Support',
+  'CLASS: Instructional Support',
+  'Coaching',
+  'Communication',
+  'Community and Self-Assessment',
+  'Culture & Language',
+  'Curriculum (Instructional or Parenting)',
+  'Data and Evaluation',
+  'ERSEA',
+  'Environmental Health and Safety / EPRR',
+  'Equity',
+  'Facilities',
+  'Family Support Services',
+  'Fiscal / Budget',
+  'Five-Year Grant',
+  'Home Visiting',
+  'Human Resources',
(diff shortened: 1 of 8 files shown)


src/migrations/20220311135140-update-topics-table.js
```diff
@@ -0,0 +1,90 @@
...
+  'Teaching Practices / Teacher-Child Interactions',
+  'Technology and Information Systems',
+  'Transition Practices',
+  'Transportation',
+];
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => queryInterface.sequelize.transaction(
+    async (transaction) => {
+    // create an "isSelectable" column, denoting that this topic
```
jasalisbury, line 44: Might be better to use https://sequelize.org/v6/manual/paranoid.html

jasalisbury (approved):
I think this is fine to merge to main but we will want to be careful about merging breaking DB changes to main with the rest of the DB schema updates. Not that there are necessarily changes that will breaking (there might not be), just something to think about as we get more of these DB updates.

src/migrations/20220311135140-update-topics-table.js
thewatermethod: You're suggesting not bothering with the isSelectable column, just making the model paranoid and deleting everything that would be unselectable?

jasalisbury: I think we can hard delete any topics not currently in use (because no other models reference this table). In the future when we won't be able to hard delete sequelize has a built in way of handling soft deletes that we use elsewhere, would be nice to handle it in the same way unless there is something I'm not thinking about. If we use paranoid we don't have to worry about adding `where: { isSelectable: true }` to our queries.

src/tools/importPlanGoals.test.js
```diff
@@ -21,7 +21,7 @@ describe('Import TTA plan goals', () => {
         const fileName = 'GranteeTTAPlanTest.csv';
         downloadFile.mockResolvedValue({ Body: readFileSync(fileName) });
         await Role.destroy({ where: {} });
-        await Topic.destroy({ where: {} });
+        await Topic.destroy({ where: {}, force: true });
         await Goal.destroy({ where: {} });
```
thewatermethod, line 25: Apparently findOrCreate fails if a topic has been soft deleted (i.e. the model has a "paranoide: true" flag on it)

src/models/resource.js
```diff
@@ -0,0 +1,42 @@
...
+      allowNull: false,
+    },
+    updatedAt: {
+      allowNull: false,
+      type: DataTypes.DATE,
+    },
+    createdAt: {
+      allowNull: false,
+      type: DataTypes.DATE,
+    },
```
GarrettEHill, line 36: Sequelize adds these columns automatically, do you need to define them? 
[https://sequelize.org/v5/manual/models-definition.html](https://sequelize.org/v5/manual/models-definition.html)
> By default, Sequelize will add the attributes createdAt and updatedAt to your model so you will be able to know when the database entry went into the db and when it was updated last.

src/migrations/20220311135140-update-topics-table.js
```diff
@@ -0,0 +1,68 @@
...
+  'Pregnancy Services / Expectant Families',
+  'Program Planning and Services',
+  'Quality Improvement Plan / QIP',
+  'Recordkeeping and Reporting',
+  'Safety Practices',
+  'Staff Wellness',
+  'Teaching Practices / Teacher-Child Interactions',
+  'Technology and Information Systems',
+  'Transition Practices',
+  'Transportation',
```
GarrettEHill, line 38: Now that these are being populated into the database, can we remove them from frontend/src/Constants.js and/or frontend/src/pages/ActivityReport/constants.js and/or src/constants.js and just pull the values from the DB?

Would we also need a front end under admin to allow adding/editing the list?

thewatermethod: I think these are both separate Jiras, which I will write

src/models/resource.js
thewatermethod: good to know! thanks

src/migrations/20220311135140-update-topics-table.js
thewatermethod: see 
- https://ocio-jira.acf.hhs.gov/browse/TTAHUB-709
- https://ocio-jira.acf.hhs.gov/browse/TTAHUB-710

src/models/resource.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+  }
+  Resource.init({
+    id: {
+      type: DataTypes.INTEGER,
+      allowNull: false,
+      autoIncrement: true,
+      primaryKey: true,
+    },
+    url: {
```
hardwarehuman, line 21: As these will be the frozen (if an AR linked to the Objective is approved) `ObjectiveResource` records, I'm not sure if we want to leave it called `url` or not. Possibly the canonical `Resource` has `canonicalUrl` instead of `url`.

hardwarehuman (changes requested):
Sorry it took me so long! I have a stronger opinion on ObjectiveResource than the `url` column.

src/models/resource.js
hardwarehuman: That's a really good point about the "canonicalUrl". I'm not at all committed to having separate column names between ObjectiveResource and Resource. Having said that, we don't even have to decide until we have an actual `Resources` table in addition to `ObjectiveResources`.

PR #819, opened 2022-03-15
[TTAHUB-706] Add Creator Role Selection for NEEDS_ACTION

AdamAdHocTeam:
## Description of change

We encountered an issue when deploying the new creator role field to production. If a report was in NEEDS_ACTION state the user has no way of setting the required creator role. This prevented them from resubmitting the report. We also found out that the same issue existed if the user had one role and a report was in the DRAFT state.

## How to test

All steps needs to be tested with a user that has a single **AND** multiple roles:

**Draft:**
1. Create a report in DRAFT state. 
2. Manually set the creator role to NULL in the DB for that report. 
3. Revisit the DRAFT report. 
4. Saving the report should automatically populate the creator role OR if you have multiple roles fore the user to select the creator role.

**Needs Action:**
1. Create a report and put it in the NEEDS_ACTION state. 
2. Manually set the creator role to NULL in the DB for that report. 
3. Revisit the NEEDS_ACTION report.
4. If the user only has one role they should be able to resubmit without error. If they user has more than one role they should be able to set the creator role on the review and submit page.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-706

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/pages/ActivityReport/index.js
+++ b/frontend/src/pages/ActivityReport/index.js
@@ -181,11 +181,15 @@ function ActivityReport({
         } else {
           report = {
             ...defaultValues,
-            creatorRole: user && user.role && user.role.length === 1 ? user.role[0] : null,
             pageState: defaultPageState,
             userId: user.id,
             regionId: region || getRegionWithReadWrite(user),
           };
+        }
+
+        // If creator role is null and user has single role set it.
+        if (!report.creatorRole) {
+          report.creatorRole = user && user.role && user.role.length === 1 ? user.role[0] : null;
         }
 
         const apiCalls = [
@@ -337,6 +341,7 @@ function ActivityReport({
 
   const onFormSubmit = async (data) => {
     const approverIds = data.approvers.map((a) => a.User.id);
+    console.log('On Submit Parent value: ', data.creatorRole);
     const reportToSubmit = {
       additionalNotes: data.additionalNotes,
       approverUserIds: approverIds,
--- a/frontend/src/pages/ActivityReport/index.js
+++ b/frontend/src/pages/ActivityReport/index.js
@@ -341,7 +341,6 @@ function ActivityReport({
 
   const onFormSubmit = async (data) => {
     const approverIds = data.approvers.map((a) => a.User.id);
-    console.log('On Submit Parent value: ', data.creatorRole);
     const reportToSubmit = {
       additionalNotes: data.additionalNotes,
       approverUserIds: approverIds,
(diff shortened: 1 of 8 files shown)


thewatermethod:
When I test this, if I have multiple roles, I am able to select and it seems to work fine, for both draft and needs action.

If I have a single role in draft, I set the creatorRole to null in the DB, and then I go back to the report and save it, the database does not change.

The same is true for a report in needs action. In that case, I am unable to re-submit a report, and the DB says null for creatorRole.

AdamAdHocTeam:
> When I test this, if I have multiple roles, I am able to select and it seems to work fine, for both draft and needs action.
> 
> If I have a single role in draft, I set the creatorRole to null in the DB, and then I go back to the report and save it, the database does not change.
> 
> The same is true for a report in needs action. In that case, I am unable to re-submit a report, and the DB says null for creatorRole.

I have updated the code to handle the single role issue.

frontend/src/pages/ActivityReport/index.js
```diff
@@ -323,6 +323,13 @@ function ActivityReport({
       reportId.current = savedReport.id;
       window.history.replaceState(null, null, `/activity-reports/${savedReport.id}/${currentPage}`);
     } else {
+      // If user has one role set creator role.
+      if (!data.creatorRole && userHasOneRole) {
+        const [onlyUserRole] = user.role;
```
AdamAdHocTeam, line 328: This seems to be the best location in code to set the creator role if they only have one.

I'm open to other suggestions.

thewatermethod: You could remove the if statement and the eslint warning by writing that more declaratively, something like:

```js
const creatorRole = !data.creatorRole && userHasOneRole ? user.role[0] : data.creatorRole;
const updatedFields = findWhatsChanged({ ...data, creatorRole}, formData);
```

thewatermethod:
Works great!

The one thing I would say is missing is an updated test... if they were passing before the most recent commit but I couldn't successfully submit a report, maybe we need a better test?

frontend/src/pages/ActivityReport/index.js
AdamAdHocTeam: Agree this way is much cleaner for the linter. C# habits are still hard to change :-).

Will change.

thewatermethod (approved):

GarrettEHill (approved):

PR #830, opened 2022-03-23
[TTAHUB-760] Always Maintain Region Filters in URL

AdamAdHocTeam:
## Description of change

We always want to maintain the region filters in the URL. This will ensure that when a URL filter is shared the user will be correctly notified if they are attempting to view a region they can't access.

## How to test

Test on both AR Landing and Regional Dashboard pages. 

Test for both a regular and Central Office user.

1. View the landing/dashboard page in a new browser. The URL should show all user regions.
2. Add a custom region filter. The URL should only show the region specified.
3. Remove the region filter. The URL should add back all regions the user can access.
4. Create a URL that has SOME of the regions the user can access. Paste it in a new browser window. You should see pills for the regions in the URL filter. 
5. Create a URL filter that has ALL of the regions the user can access. Paste it in a new browser window. There shouldn't be any filter pills for region but the URL should reflect all regions.

All preexisting filter functionality should still work as expected.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-684

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/filter/FilterPanel.js
+++ b/frontend/src/components/filter/FilterPanel.js
@@ -8,25 +8,31 @@ export default function FilterPanel({
   onRemoveFilter,
   filters,
   filterConfig,
-  hideRegionFiltersByDefault,
   onApplyFilters,
   applyButtonAria,
+  allUserRegions,
 }) {
-  // Prop 'hideRegionFiltersByDefault' is used to set the
-  // initial hide/show region filter state based on the User's settings.
-  const [hideRegionFilters, setHideRegionFilters] = useState(hideRegionFiltersByDefault);
   const [filtersToShow, setFiltersToShow] = useState([]);
 
   useEffect(() => {
-    // If we are hiding region filters prevent any underlying component from getting region filters.
-    setFiltersToShow(hideRegionFilters ? filters.filter((f) => f.topic !== 'region') : filters);
-  }, [hideRegionFilters, filters]);
+    // Determine if filters contain all regions.
+    const passedRegionFilters = filters.filter((f) => f.topic === 'region').map((r) => r.query);
+    let containsAllRegions = true;
+    if (allUserRegions) {
+      allUserRegions.forEach((r) => {
+        if (!passedRegionFilters.includes(r)) {
+          containsAllRegions = false;
+        }
+      });
+    }
+    // Hide or Show Region Filters.
+    setFiltersToShow(containsAllRegions ? filters.filter((f) => f.topic !== 'region') : filters);
+  }, [filters, allUserRegions]);
 
   const onApply = (items) => {
-    // Check for region filter.
-    const regionFilter = items.find((f) => f.topic === 'region');
-    onApplyFilters(items, !regionFilter);
-    setHideRegionFilters(!regionFilter);
+    // Check for region filters.
+    const regionFilters = items.filter((f) => f.topic === 'region');
+    onApplyFilters(items, regionFilters.length === 0);
   };
 
   const onRemoveFilterPill = (id) => {
@@ -35,20 +41,12 @@ export default function FilterPanel({
     const isRegionFilter = pillToRemove && pillToRemove.topic === 'region';
 
     if (isRegionFilter) {
-      const otherRegion = filters.find((f) => f.id !== id && f.topic === 'region');
-      const haveRemainingRegions = !hideRegionFiltersByDefault && otherRegion;
-      onRemoveFilter(id, !haveRemainingRegions);
-      setHideRegionFilters(!haveRemainingRegions);
+      // Check if we removed the last region filter.
+      const otherRegions = filters.filter((f) => f.id !== id && f.topic === 'region');
+      onRemoveFilter(id, otherRegions.length === 0);
     } else {
       onRemoveFilter(id, false);
     }
-
-    /*
-       onRemoveFilter(id, isRegionFilter);
-    if (isRegionFilter) {
-      setHideRegionFilters(true);
-    }
-    */
   };
 
   return (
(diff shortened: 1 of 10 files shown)


frontend/src/components/Modal.js
```diff
@@ -37,18 +43,33 @@ const Modal = ({
       </div>
       <ModalFooter>
         <ButtonGroup>
-          <ModalToggleButton className={cancelButtonCss} data-focus="true" type="button" modalRef={modalRef} closer>
```
thewatermethod, line 40: Probably makes sense for this branch to track the permissions modal one, I see a bunch of duplicated code here (unless I'm confused, which is possible)

frontend/src/components/RegionPermissionModal.js
```diff
@@ -0,0 +1,100 @@
...
+
+  const showFilterWithMyRegionAccess = () => {
+    showFilterWithMyRegions();
+    modalRef.current.toggleModal(false);
+  };
+
+  const smartSheetAccessLink = 'https://app.smartsheetgov.com/b/form/f0b4725683f04f349a939bd2e3f5425a';
+  const requestSmartSheetAccess = () => {
+    // Open link in new tab and close modal.
+    window.open(smartSheetAccessLink, '_blank');
```
thewatermethod, line 44: [Avoid resorting to window.open()](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#avoid_resorting_to_window.open)

frontend/src/components/filter/FilterPanel.js
```diff
@@ -1,22 +1,73 @@
...
+  onApplyFilters,
+  applyButtonAria,
+  allUserRegions,
+}) {
+  const [filtersToShow, setFiltersToShow] = useState([]);
+
+  useEffect(() => {
+    // Determine if filters contain all regions.
+    const passedRegionFilters = filters.filter((f) => f.topic === 'region').map((r) => {
+      if (isArray(r.query)) {
```
thewatermethod, line 21: As you were saying, I think if we actually want the URL to theoretically match from viewer to destination, we will have to include all visible regions in the URL

thewatermethod: ?region.nin[]=1&region.in[]=2

thewatermethod (commented):
Seems to work when I tested it. A few tweaks as noted. If it were me, I'd probably check with the larger team before even getting into changing/fixing it

AdamAdHocTeam:
> As a user with multiple regions I get a URL like `...?region.in[]=2&region.in[]=3&region.in[]=4&region.in[]=5&region.in[]=6...` on the landing page. This means we are getting all activity reports for "region 2 AND region 3..." and so on. This causes me to have zero activity reports loaded (since no activity report can be for multiple regions).

Right now my main concern going forward with this approach is how do we handle a region **NOT IN** filter. I think we would need to update the URL to contain all regions except the "NOT IN" region. Or we just leave the filter for this case as is for now?

Another option would be a copy filter button that automatically adds all other regions to the clipboard filter.

jasalisbury:
> > As a user with multiple regions I get a URL like `...?region.in[]=2&region.in[]=3&region.in[]=4&region.in[]=5&region.in[]=6...` on the landing page. This means we are getting all activity reports for "region 2 AND region 3..." and so on. This causes me to have zero activity reports loaded (since no activity report can be for multiple regions).
> 
> Right now my main concern going forward with this approach is how do we handle a region **NOT IN** filter. I think we would need to update the URL to contain all regions except the "NOT IN" region. Or we just leave the filter for this case as is for now?
> 
> Another option would be a copy filter button that automatically adds all other regions to the clipboard filter.

Do we need to handle a user who doesn't have access to a "NOT IN" region filter? They don't have access to the region so they are seeing the correct data.

AdamAdHocTeam:
> 

I think I misunderstood what you where saying here. 

Did you set a date on the alerts your expecting to see? If I create reports with multiple regions with no date the alerts don't show until I remove the date filter. There is a separate jira to fix this issue.

AdamAdHocTeam:
> As a user with multiple regions I get a URL like `...?region.in[]=2&region.in[]=3&region.in[]=4&region.in[]=5&region.in[]=6...` on the landing page. This means we are getting all activity reports for "region 2 AND region 3..." and so on. This causes me to have zero activity reports loaded (since no activity report can be for multiple regions).

I think I misunderstood what you where saying here.

Did you set a date on the alerts your expecting to see? If I create reports with multiple regions with no date the alerts don't show until I remove the date filter. There is a separate jira to fix this issue.

jasalisbury:
> > As a user with multiple regions I get a URL like `...?region.in[]=2&region.in[]=3&region.in[]=4&region.in[]=5&region.in[]=6...` on the landing page. This means we are getting all activity reports for "region 2 AND region 3..." and so on. This causes me to have zero activity reports loaded (since no activity report can be for multiple regions).
> 
> I think I misunderstood what you where saying here.
> 
> Did you set a date on the alerts your expecting to see? If I create reports with multiple regions with no date the alerts don't show until I remove the date filter. There is a separate jira to fix this issue.

You are correct, the date filter was causing no reports to show up, sorry about that!

frontend/src/components/filter/FilterPanel.js
AdamAdHocTeam: I will discuss this with Josh.

frontend/src/components/RegionPermissionModal.js
AdamAdHocTeam: This has been changed at the modal level to take a button. We now pass a link to open the smart sheet page.

frontend/src/components/filter/FilterPanel.js
AdamAdHocTeam: Region not in has been removed.

jasalisbury (approved):
:rocket:

thewatermethod (approved):

PR #834, opened 2022-03-24
[TTAHUB-684] Region Permission Modal

AdamAdHocTeam:
## Description of change

We display a modal on the AR Landing and Regional Dashboard page if the user is attempting to use a filter with a region they cant access. The user must choose one of two options. The first is to view the filter with regions they do have access to. The second opens the TTA Smart Sheet request page in a new tab then shows the filter with regions they have access.

Note: This change ONLY includes the Region Permission's modal.

** 03/30 Fix for Region **IS NOT**:
If a user is given a **region is not** filter we have no way of knowing the other regions the user was intended to see. For now this code will ignore a region not in filter and only display the modal for specified regions.

**For Example:** User A has regions 1, 2, 3 and User B only has access to region 1. If a filter is given to user B that says region is not 3. User A would be expecting User B to see the results of regions 1 & 2. But User B can only see region 1. The result of the filter are correct but we have no way of knowing the intended result from the passed URL.

## How to test

Create a filter URL that contains regions your user can and can't access. When clicking or pasting this URL into a new tab you should receive the permission modal with the two options. Using a filter that contains regions you do have access to should not display the modal.

This should be tested on both the AR Landing and Regional Dashboard pages.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-684

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


frontend/src/components/Modal.js
```diff
@@ -37,18 +43,33 @@ const Modal = ({
...
+                  {cancelButtonText}
+                </ModalToggleButton>
+              )
+              : null
+          }
           {
             showOkButton
               ? (
-                <Button className={okButtonCss || 'usa-button usa-button--secondary usa-button'} type="button" aria-label={okButtonAriaLabel} modalRef={modalRef} onClick={onOk}>
+                <Button className={okButtonCss || 'usa-button usa-button--secondary usa-button'} data-focus={!!hideCancelButton} type="button" aria-label={okButtonAriaLabel} onClick={onOk}>
```
thewatermethod, line 58: !! hideCancelButton === hideCancelButton, right?

frontend/src/components/Modal.js
```diff
@@ -20,13 +20,19 @@ const Modal = ({
   okButtonCss,
   cancelButtonCss,
   showTitleRequired,
+  secondaryOkButtonText,
+  onSecondaryOk,
```
thewatermethod, line 24: Nitpick: probably not "secondaryOk" but "secondaryAction", to me OK means acknowledge and dismiss the modal

frontend/src/components/RegionPermissionModal.css
```diff
@@ -0,0 +1,21 @@
...
+    padding: 1.25rem 14% 4rem;
+}
+
+
+.smart-hub--region-permission-modal .usa-modal--lg .usa-modal__heading {
+    font-size: 1.85rem;
+}
+
+.smart-hub--region-permission-modal .popup-modal button {
+    padding: 9px 16px;
```
thewatermethod, line 12: Again, nitpick: we're using the 8px scale most everywhere else, we've got a 9px here, also the font-size: 1.85rem, did that come from somewhere? Would rather use one of the pixel sizes from here if possible: https://designsystem.digital.gov/utilities/font-size-and-family/

frontend/src/pages/Landing/index.js
```diff
@@ -59,6 +60,21 @@ function Landing() {
   const defaultRegion = user.homeRegionId || regions[0] || 0;
   const hasMultipleRegions = regions && regions.length > 1;
 
+  const buildDefaultRegionFilters = () => {
```
thewatermethod, line 63: would this be something that benefits from a useMemo?

frontend/src/pages/RegionalDashboard/index.js
```diff
@@ -38,6 +39,21 @@ export default function RegionalDashboard() {
   const userHasOnlyOneRegion = useMemo(() => regions.length === 1, [regions]);
   const defaultRegion = useMemo(() => regions[0].toString(), [regions]);
 
+  const buildDefaultRegionFilters = () => {
```
thewatermethod, line 42: looks the same both places, maybe you could move it to somekind of helper file

frontend/src/pages/RegionalDashboard/index.js
```diff
@@ -83,11 +99,23 @@ export default function RegionalDashboard() {
 
   const filtersToApply = expandFilters(filters);
 
+  const showFilterWithMyRegions = () => {
+    // Exclude region filters we dont't have access to and show.
```
thewatermethod, line 103: same with this, could be moved to a regionHelpers file

GarrettEHill (approved):

frontend/src/components/Modal.js
AdamAdHocTeam: Yep will fix, maybe auto linter fix.

frontend/src/components/Modal.js
AdamAdHocTeam: Agree sounds better.

frontend/src/pages/RegionalDashboard/index.js
AdamAdHocTeam: Agree, moved.

AdamAdHocTeam:
> LGTM, just left some nitpicking on which your mileage may vary

Code has been updated based on comments if you want to have another look.

thewatermethod (approved):

GarrettEHill (approved):

jasalisbury (approved):
:rocket:

PR #835, opened 2022-03-25
[TTAHUB-698] update goal status UI

thewatermethod:
## Description of change
- You set goal statuses by clicking on the status itself in the goal table
- Previous status is recorded in the database and referenced in the dropdown when reopening a goal
- The backend has been updated, but the handler still isn't connected due to the note that I saw in there saying we needed to wait for DB updates (still the case?)

There was a note in the Jira about which fields would be editable but I'm assuming that doesn't apply to the actual changes here, since we're only talking about the one screen.

## How to test
- The status dropdown works, calls the right functions and shows the right statuses based on the goal's current and historical statuses

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-698

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


src/services/goals.js
```diff
@@ -203,19 +203,22 @@ export async function copyGoalsToGrants(goals, grantIds, transaction) {
...
   goalId,
+  oldStatus,
   newStatus,
   closeSuspendReason,
   closeSuspendContext,
 ) {
   /* TODO:
-    Disable for now until goals are unique grants.
+    Disable for now until goals are unique grants. ?????
   */
```
thewatermethod, line 213: Don't exactly understand this (as far as what we are waiting for), so I've left it out for now, but will gladly finish the job here if/when this is resolved

jasalisbury: It was disabled to prevent us (or other users) accidentally setting the goal status of a lot of different goals at once. Since a single goal can be "owned" by any number of recipients setting the goal status for a single recipient actually sets it for all recipients with that goal.

jasalisbury:
I get a 401 when I try to change a goal status. The regionId being checked from the goal is a string and is being compared to a number from the permission regionId param. If I force both to be numbers I still get a 404 when I try to set the status of the goal.

We will want to update the route to retrieve the goal from the DB and figure out the region ID from there instead of relying on what is passed into the route. Or I guess we can keep it as is with a lot of notes to update once the schema has been updated, that way we don't solve something that may be tricky that will go away in a few weeks anyways.

thewatermethod:
> We will want to update the route to retrieve the goal from the DB and figure out the region ID from there instead of relying on what is passed into the route. Or I guess we can keep it as is with a lot of notes to update once the schema has been updated, that way we don't solve something that may be tricky that will go away in a few weeks anyways.

I just added a new service in #808 that if approved we can port into this

jasalisbury:
I still get a 404 when I try to update goals. If you have a copy of prod loaded locally it is the [goal here](http://localhost:3000/recipient-tta-records/7781/region/2/goals-objectives?createDate.win=2022%2F01%2F01-2022%2F04%2F04). If you don't have the backup loaded and can't recreate locally let me know and we can pair to troubleshoot

thewatermethod:
> I still get a 404 when I try to update goals. If you have a copy of prod loaded locally it is the [goal here](http://localhost:3000/recipient-tta-records/7781/region/2/goals-objectives?createDate.win=2022%2F01%2F01-2022%2F04%2F04). If you don't have the backup loaded and can't recreate locally let me know and we can pair to troubleshoot

The ```changeGoalStatus``` service is still commented out so a 404 is expected (not a 401)

frontend/src/components/GoalsTable/StatusDropdown.js
```diff
@@ -0,0 +1,122 @@
+/* eslint-disable no-unused-vars */
```
jasalisbury, line 1: Can we re-enable this rule?

jasalisbury (approved):
:ship:

PR #839, opened 2022-03-29
[TTAHUB-691] Update Participants

AdamAdHocTeam:
## Description of change

Update other entity participants list to be alphabetical. 

Write a script to update participants.

## How to test

Create an AR of type other entity the participants list should now be alphabetized. 

Run the yarn command to execute the 'updateParticipantsList'. The script should run successfully and only bad participant data should be updated (see JIRA for list).

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-691

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/pages/ActivityReport/constants.js
+++ b/frontend/src/pages/ActivityReport/constants.js
@@ -46,11 +46,11 @@ export const otherEntityParticipants = [
   'Local/State Agency(ies)',
   'OCC Regional Office',
   'OHS Regional Office',
-  'Other',
   'Regional Head Start Association',
   'Regional TTA Team / Specialists',
   'State Early Learning System',
   'State Head Start Association',
+  'Other',
 ];
 
 // Note that if this topic list is changed, it needs also to be changed in
--- a/frontend/src/pages/ActivityReport/constants.js
+++ b/frontend/src/pages/ActivityReport/constants.js
@@ -42,15 +42,15 @@ export const recipientParticipants = [
 ];
 
 export const otherEntityParticipants = [
-  'Local/State Agency(ies)',
   'HSCO',
+  'Local/State Agency(ies)',
   'OCC Regional Office',
   'OHS Regional Office',
+  'Other',
   'Regional Head Start Association',
   'Regional TTA Team / Specialists',
   'State Early Learning System',
   'State Head Start Association',
-  'Other',
 ];
 
 // Note that if this topic list is changed, it needs also to be changed in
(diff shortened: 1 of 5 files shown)


jasalisbury (approved):
:+1: Everything seems to work. One thing is the legacy report view still shows the old participants since it is pulled from the imported data. Are we sure that is the desired behavior? Or are we only worried about the CSV download?

AdamAdHocTeam:
> 👍 Everything works. One thing is the legacy report view still shows the old participants since it is pulled from the imported data. Are we sure that is the desired behavior? Or are we only worried about the CSV download?

I have updated the code to also update the imported participants after checking with Angela.

frontend/src/pages/ActivityReport/constants.js
```diff
@@ -42,15 +42,15 @@ export const recipientParticipants = [
 ];
 
 export const otherEntityParticipants = [
-  'Local/State Agency(ies)',
   'HSCO',
+  'Local/State Agency(ies)',
   'OCC Regional Office',
   'OHS Regional Office',
+  'Other',
```
GarrettEHill, line 49: If this list is displayed in order, should "other" be last instead of alphabetical so the user would need to scroll past all the options first?

src/tools/updateParticipantsList.js
```diff
@@ -0,0 +1,64 @@
...
+    if (updatedImported) {
+      auditLogger.info(`Updating report ${report.id}'s imported participant from [${report.imported.granteeParticipants.replaceAll('\n', ' ')}] to [${importedGranteeParticipants.replaceAll('\n', ' ')}].`);
+    }
+
+    const imported = {
+      ...report.imported,
+      granteeParticipants: importedGranteeParticipants,
+    };
+
+    return report.update({ participants, imported });
```
GarrettEHill, line 62: could your function please take, use, and pass a transaction value for all db interaction. we need to start trying to keep all db interactions within transactions to allow the audit system to effectively track which user caused which action that modified the data.

GarrettEHill (changes requested):

frontend/src/pages/ActivityReport/constants.js
AdamAdHocTeam: Moved to bottom.

src/tools/updateParticipantsList.js
AdamAdHocTeam: Updated to use a transaction.

GarrettEHill (approved):

PR #847, opened 2022-04-06
[TTAHUB-613] edit goal from table

thewatermethod:
## Description of change

The context menu was re-added to the goals table. There is now one option, "edit". Seems reasonable to stop here as it is becoming difficult to test with partial backend functionality and changing requirements. (also, at 44 files changed)

- There are no changes to the "new" goal form for creating draft goals
- The backend is not wired until we have migrations in place 

See "how to test" for variations and details on what needs to work and how

Labels & validation messages were also updated to match the most recent mockup

## How to test

View the edit goal form for goals in various states

### a draft goal
- all fields should be editable
- there is no objective status dropdown

### a "not started" goal
- If the goal is *not* on an activity report, it behaves as a draft goal.

### a goal on an activity report
- You cannot remove or edit previously added topics & resources
- The only other editable fields are "objective status" and "estimated close date"

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-613

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


AdamAdHocTeam (approved):
Looks good. 

Please review the below comments to see if you think they need attention.

- We need a way to label and outline objectives to give context to what the user is looking at (maybe "Objective 1 Summary).

- If I set a date on an goal then click "save and continue". Then click edit to go back the date (only) is blank.

- If a goal was on an AR. I was able to edit Goal topics and Objective fields. Not sure if this is by design

jasalisbury (commented):
I see some propType errors for `datePickerKey`. Also when I edit a `draft` report I get stuck in a loop calling the API (this endpoint http://localhost:3000/api/goals/$id/recipient/$id)

thewatermethod:
> I see some propType errors for datePickerKey. Also when I edit a draft report I get stuck in a loop calling the API (this endpoint http://localhost:3000/api/goals/$id/recipient/$id)

Def addressed the proptype errors. I wasn't able to reproduce the loop, but hopefully its something I fixed when addressing one of the bugs Adam found.

thewatermethod:
> We need a way to label and outline objectives to give context to what the user is looking at (maybe "Objective 1 Summary).

Something to bring up at the UI meeting with Lauren & Kelly

> If I set a date on an goal then click "save and continue". Then click edit to go back the date (only) is blank.

This has been addressed, I believe. Thanks for pointing that out, led to finding a lot of issues.

> If a goal was on an AR. I was able to edit Goal topics and Objective fields. Not sure if this is by design.

You should be able to add topics & resources only. (None of the objectives will have topics or resources right now because that's not how our data model has worked)

thewatermethod:
> I think I figured out why I'm getting the loop. I am opening a goal with no fields filled out that has an objective that is also blank. I'm not sure how the goal originally got saved like that (I can't currently save any goals, which I think is intentional?). Also getting a prop type error (I assume from the goal that is causing the loop) saying objective.arId is required. Since objectives can be created outside of ARs I think we need to make that prop optional.

You wouldn't happen to have the recipientId/goalId of that blank goal, would you?

jasalisbury (approved):
:rocket:

PR #853, opened 2022-04-08
Nightly download sets auditlog metadata

jasalisbury:
## Description of change

Audit log metadata for the nightly download is now set. Since the nightly job has no access to the `httpContext` the audit* variables are only set if they are null. This allows those variables to be set for the transaction before hitting sequelize hooks and without using an `httpContext`.

I also tried moving all DB calls down to the bottom of `processFiles` so we wouldn't have to wrap the whole method in a transaction. Even though I kept the same order of the calls I was seeing foreign key errors. I wasn't able to track down what exactly was happening.

* The user id is set to 0 to distinguish between the "null" case and system user
* Transaction ID is a UUID
* Session signature is the timestamp of the run followed by the sha256 of the zip file processed
* The audit descriptor lets us know the change was from the nightly download

## How to test

Verify tests pass. Also verify you can still make edits and see the correct info in the audit log (such as the user you made the changes with)

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-672

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/models/auditModelGenerator.js
+++ b/src/models/auditModelGenerator.js
@@ -23,19 +23,32 @@ const tryJsonParse = (data) => {
   return newData;
 };
 
+const pgSetConfigIfNull = (settingName, value) => `
+  set_config(
+    '${settingName}',
+    COALESCE(current_setting('${settingName}', true), '${value}'),
+    true
+  )`;
+
 const addAuditTransactionSettings = async (sequelize, instance, options, type) => {
   const loggedUser = httpContext.get('loggedUser') ? httpContext.get('loggedUser') : '';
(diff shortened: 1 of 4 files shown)


thewatermethod (approved):
Tests pass, audit log still needs to work

src/models/auditModelGenerator.js
```diff
@@ -23,19 +23,32 @@ const tryJsonParse = (data) => {
   return newData;
 };
 
+const pgSetConfigIfNull = (settingName, value, alias) => `
```
GarrettEHill, line 26: That is slick. I like it

GarrettEHill (approved):

PR #867, opened 2022-04-15
Add GTM snippet

kryswisnaskas:
## Description of change

Adds Google Tag Manager snippet with Google analytics measurement. Biggest change to the site is the frontend `index.html` has a `nonce` replaced when retrieved from the server.

From https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_inline_script we must use `unsafe-inline`, provide a `nonce`, or provide the hash of the script for that script to execute. The point of google tag manager is the contents of the scripts can change without a software deployment (via CircleCI) so we can't generate a hash at build time (because the hash could be different when users are retrieving the page). Also see https://developers.google.com/tag-platform/tag-manager/web/csp#enable_the_container_tag_to_use_csp.

Now the frontend `index.html` has a `__NONCE__` in the tag manager script snippet. When `index.html` is retrieved from the server `__NONCE__` is replaced with a generated `nonce` value. The same `nonce` is placed in the response `csp` header.

Google tag manager and google analytics were setup with environments that match ours, local, sandbox, dev, staging, and prod (or GA calls it live). There are build variables in CircleCI that insert the correct tag manager options into the `index.html` page depending on the environment. There is a "lookup" variable in google tag manager that matches the google tag manager environment to the proper google analytics `Measurement ID`.

NOTE: Google analytics wouldn't let me name a property with three characters so `dev` is `development`

## How to test

This is currently deployed to sandbox. When you browse to sandbox you should see the google tag manager script in your network tab and after browsing around calls to google analytics. Login to https://analytics.google.com/ and https://tagmanager.google.com/ using the google account in our `1password` vault. You should see your session inside google analytics when viewing the sandbox environment. Some more things to check, sorry they aren't easy to navigate to, let me know if you need help finding:

 * Environment variables in circle ci only show the last four characters but double check those characters match those in google tag manager. Find the values in tag manager by clicking the `Admin` tab, environments, `Actions` context menu, and `Get Snippet`. You'll have to search a bit but they are the query params `gtm_auth` and `gtm_preview`
 * Make sure the mapping between google tag manager and google analytics is correct. Can check by clicking `Variables` in the left hand menu, and selecting `Lookup - GA Tracking IDs`. Can find the values in google analytics by clicking the gear icon in the lower left, `Data Streams`, click the only option, the `Use existing on-page tag`, `Google Analytics`, and the `Measurement ID` can be found on the fourth step

Once dev is no longer in use I'll deploy this branch to both to ensure we can get analytics for `dev` and `sandbox` separately.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-379
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-791

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


thewatermethod:
- Confirmed that my session was being logged in analytics on sandbox
- Confirmed that the mappings in CircleCI match the values in the tag manager snippet
- Confirmed that the tracking ids in the lookup table in tag manager match the values in the data streams on analytics

I had to do the first in Chrome - the analytics snippet was still blocked in my (stock, no privacy addons) Firefox install

thewatermethod (approved):

thewatermethod:
hmm, well I don't have a screenshot, it simply doesn't load as far as I can tell. I saw no session activity in analytics until I visited on Chrome. I do seem to have the cookie though so maybe I am just missing something.

This is my console (hard to tell in here since the line referenced is "1:1" on the minified source, but I'm assuming that those errors are related to Touchpoints)

Screen Shot 2022-05-09 at 10 45 16 AM

The full contents of my network tab after a refresh and page navigation (I don't see any google analytics stuff)

Screen Shot 2022-05-09 at 10 52 19 AM

thewatermethod:
Further curiosity - Firefox has had this built in "enhanced tracking protection" panel for about a year. (It's on for all sites by default, I've never touched the settings for that, so we can assume that is the behavior we'd encounter in the wild). You can see I get no registered trackers on sandbox

Screen Shot 2022-05-10 at 8 46 44 AM

But you can see from Mozilla.com that it does list google analytics if it finds it under "allowed trackers"

Screen Shot 2022-05-10 at 8 47 13 AM

When I expand "tracking content"

Screen Shot 2022-05-10 at 8 47 07 AM

Regardless, I wouldn't think this is a blocker for release because 1) client based analytics always have issues like this, they are kind of fragile by nature and 2) Firefox isn't one of our supported browsers

src/app.js
```diff
@@ -15,26 +17,38 @@ import updateGrantsRecipients from './lib/updateGrantsRecipients';
...
+app.use((req, res, next) => {
+  res.locals.nonce = crypto.randomBytes(16).toString('hex');
+  const cspMiddleware = helmet.contentSecurityPolicy({
     directives: {
       ...omit(helmet.contentSecurityPolicy.getDefaultDirectives(), 'upgrade-insecure-requests', 'block-all-mixed-content', 'script-src', 'img-src', 'default-src'),
       'form-action': ["'self'"],
-      scriptSrc: ["'self'", 'https://touchpoints.app.cloud.gov/touchpoints/7d519b5e.js'],
-      imgSrc: ["'self'", 'data:', 'https://touchpoints.app.cloud.gov'],
-      defaultSrc: ["'self'", 'https://touchpoints.app.cloud.gov/touchpoints/7d519b5e/submissions.json'],
+      scriptSrc: ["'self'", 'https://touchpoints.app.cloud.gov*', '*.googletagmanager.com'],
```
thewatermethod, line 39: I think this should be either https://touchpoints.app.cloud.gov/* or you can possibly just remove it (the survey is just a link now, not a modal)

thewatermethod:
On Firefox with all extensions disabled, I see no GA activity.

Screen Shot 2022-05-19 at 10 27 38 AM

FWIW Firefox and Internet Explorer don't support script-src-elem or script-src-attr, and I guess its only in release candidate in Safari.

Works great in Chrome, think that's probably good enough TBH

AdamAdHocTeam (approved):
Looks good.

Was no longer on sandbox, but was able to see past history and matched variable tags between environments.

PR #868, opened 2022-04-18
[TTAHUB-690/743] Goals and objectives (on Activity Report) page redesign 

thewatermethod:
## Description of change

Changing the goals & objectives tab on the AR to use the new fields. This change includes the new frontend for both recipient and other entity reports.

This is for the **frontend** only! Every time you save, your report will be in a nuked state so please bear that in mind while testing.

A future PR will include the radio buttons/file uploader.

## How to test

On the goals & objectives tab, you can add and edit goals & objectives. The validation works in a similar way to the existing AR form. You will not be able to save any data and return to it. If your report autosaves, and you navigate away from a tab, you will find your goals & objectives all gunked up.

Additionally, form controls conform to the logic here: https://docs.google.com/document/d/1AG1Zycwgapeeo2gjVkW4uAwF9x7VRZhHqFiwJQCK2as/edit - note that the backend doesn't allow right now for the behavior we want in topics and resources, so that's an exception

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-690
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-743

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


jasalisbury:
Just played around with it, some observations:
 
- Add a new goal replaces the current goal when selecting a previously created goal (can only work on a single goal at a time). The objectives added before clicking `new goal` are still visible.
- The add a new goal button when the only goal is a new goal does not work properly. Sometimes it does nothing. Sometimes it clears out the objective (but keeps `ttaProvided`) but keeps the text of the new goal I just typed in the `Recipient's goal` field
- Do we want to prevent adding of a "resource link" until the link before is valid?
- If I have focus on a text box (say the `Recipient's goal` field) and click the date picker menu button the validations fire and the date picker menu does not open

thewatermethod:
> Add a new goal replaces the current goal when selecting a previously created goal (can only work on a single goal at a time). The objectives added before clicking new goal are still visible.

Whoops, my bad. Resolved this. Will need refactoring when the backend is more complete as we'll need to handle stuff moving in and out of edit mode and I don't necessarily know the shape of the API and how we can massage the form stuff to match it a little better

> The add a new goal button when the only goal is a new goal does not work properly. Sometimes it does nothing. Sometimes it clears out the objective (but keeps ttaProvided) but keeps the text of the new goal I just typed in the Recipient's goal field

And this was showing when it shouldn't have been, my bad. It should only appear in the "review" page, and it should work as you expect there, spawning a new, blank form.

> Do we want to prevent adding of a "resource link" until the link before is valid?

I don't believe so, I think Lauren and Kelly explicitly said that we want to let them go wild (probably not in those words) but I could be misremembering something from one of those meetings.

> If I have focus on a text box (say the Recipient's goal field) and click the date picker menu button the validations fire and the date picker menu does not open

Not sure if anything can be done about this. I think it's arguably what we want to happen.

jasalisbury:
Some more observations:

* Should we make the context menu bigger on "saved" goals?
* Validation message isn't cleared. Try to "save goal" with an invalid objective. Make the objective valid and click "save goal". Error message still displayed
* Individual fields that are invalid aren't highlighted
* Having goal `a` "edit-able" causes the "delete" button on goal `b` to not work
* Clicking "edit" on goal `a` while goal `b` is in the edit state causes goal `a` to disappear
* Do we want to allow users to add multiple "empty" objectives (feels different then the `resources` question above, even if it has the same answer)

(ignore if this is because it isn't hitting the backend)
* saving a goal and clicking "edit" makes my objectives disappear

AdamAdHocTeam (commented):
Found some minor things. 

- Recipient Goal and Estimated Goal Close don't persist on the saved goal view or going back via edit.
- Saving a blank Goal only flags Recipient goal field as required and not close date (no red border).
- Saving a blank Objective leaving all other fields blank only flags the title as required (no red border).
- No clear UI distinction between objectives (might be a UX design issue).

jasalisbury:
Still some validation weirdness, if I select an already created goal and add a new objective I see `Please complete all required fields...` at the top. I can't seem to get that alert to ever disappear.

jasalisbury:
Think we are getting close to being able to merge, seeing some more weirdness but hard to say if it is because we aren't actually saving to the backend.

 * Clicking "edit" on a saved goal (goal A) while a different goal (goal B) is already being "edited" causes the goal that was already being "edited" (goal B) to disappear 
 * The context menu for unsaved goals is a uuid. I also saw the context menu have the goal id as "undefined" but not sure how I got that to happen
Screenshot from 2022-05-09 10-02-00

thewatermethod:
I updated the logic, and edited a couple of goals in sequence, including with one already open to edit and I didn't have any issues. I didn't see that undefined thing. Saving things does screw up the flow, so even if you autosave while doing this, you might see additional errors.

I added the UUID thing as a temporary measure that will be removed when the goals come back from the database with an id (when you click save goals, in the actual version of this, we'll go to the database at that point and get an id, which is what will be printed in the button, etc)

jasalisbury:
> I updated the logic, and edited a couple of goals in sequence, including with one already open to edit and I didn't have any issues. I didn't see that undefined thing. Saving things does screw up the flow, so even if you autosave while doing this, you might see additional errors.
> 
> I added the UUID thing as a temporary measure that will be removed when the goals come back from the database with an id (when you click save goals, in the actual version of this, we'll go to the database at that point and get an id, which is what will be printed in the button, etc)

Yeah might be time to merge this in and try to find bugs after everything is saving properly to the database instead of finding half bugs that may be real or may be because an autosave hit.

jasalisbury (approved):

PR #869, opened 2022-04-19
Sequelize uses transactions automatically for all queries

jasalisbury:
## Description of change

See https://sequelize.org/docs/v6/other-topics/transactions/#automatically-pass-transactions-to-all-queries We no longer need to explicitly pass transactions to each query. Sequelize uses `cls-hooked` to pass transactions for us. We still have to kick off the transaction which is handled for all routes with `transactionWrapper`.

Had to add `waitForExpect` to wait on some transactions to end. For the file handler tests it appears `supertest` gives control back to tests once the `res` is set in a handler. The file handler does additional processing after setting the result so the transaction does not finish before tests regain control, causing the queries inside that request to not be visible yet (because the transaction hasn't been committed).

## How to test

* tests pass
* Running locally you are able to use the site as normal (create reports, view reports, etc...)

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-657

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


GarrettEHill (approved):

thewatermethod (approved):
Confirmed the site still functions as before & the tests still pass

PR #870, opened 2022-04-19
[TTAHUB-251] Warn of simultaneous editing of reports

thewatermethod:
## Description of change

- Installed the "ws" and "redis" packages on the backend to handle websockets and the node redis connection respectively.
- Use redis & websockets on the backend to create a pub sub channel 
- Frontend uses websockets when on an activity report to report its location within the report
- Displays an alert based on the messages coming down the pipe

## How to test
- Update packages if necessary
- You need to have Redis running
- Create a report. If you look at your network tab, you should have a websocket connection that publishes data on save.
- If you want to view the message, you can edit ```frontend/src/components/SocketAlert.js:27``` to not filter out the current user. You'll need to have two tabs on the same page, and you'll have to wait for the save event before the alert appears.
- [Mockup](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-720)

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-431
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-251

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


src/index.js
```diff
@@ -10,4 +11,21 @@ const server = app.listen(port, () => {
   auditLogger.info(`Listening on port ${port}`);
 });
```
thewatermethod, line 13: We could obviously put some logic in this parsed message stuff, I think the approach I'd probably take is to maintain the connections/clients in little arrays somehow, based on report id

jasalisbury:
Some questions I have that will require searching/tinkering to answer:
1. Since we have multiple instances we have to store data about who has what report open somewhere (redis?) and not local to the instance (i.e. in memory)
2. Since we route traffic through a load balancer do we need to make connections "sticky" so we route the websocket request to the correct instance? If so can we configure out load balancer in cloud.gov? Are we lucky and websockets "just work"?
3. Are we sure websockets are support with cloud.gov? Should be supported from AWS assuming we aren't on a `Classic Load Balancer` https://aws.amazon.com/elasticloadbalancing/features/

thewatermethod:
FYI, the number of instances is in deployment_config/{env}_vars.yml

thewatermethod:
Can also confirm the demo works on sandbox now, will still have to do some engineering to handle the multiple instances but that's proof of concept at least that websockets works on cloud.gov

src/index.js
jasalisbury: We can't store any state info in memory since prod has multiple instances. Will have to be in postgres or redis (or something else that we spin up just for this purpose)

thewatermethod: I'm looking at the redis pub/sub stuff right now, seems like a pretty decent solution and we shouldn't have to add anything infrastructure-wise

thewatermethod: ex: https://gist.github.com/pietern/348262

jasalisbury:
> Can also confirm the demo works on sandbox now, will still have to do some engineering to handle the multiple instances but that's proof of concept at least that websockets works on cloud.gov

Nice! Three other things to figure out that I can see

1. Get websockets working with multiple instances from the cloud.gov side (might be easy idk)
2. Figure out websocket storage when using multiple instances
3. Come up with a scheme to only send messages to users when needed and allow multiple "types" of messages. Right now we want to know when users have reports open but we may have another need to use websockets. Hopefully we find a way of sending report user messages that allows other types of messages

frontend/package.json
```diff
@@ -178,7 +178,9 @@
       "<rootDir>/src/widgets/Example.js",
       "<rootDir>/src/pages/RegionalDashboard/formatDateRange.js",
       "<rootDir>/src/pages/RegionalDashboard/constants.js",
-      "<rootDir>/src/pages/makecolors.js"
+      "<rootDir>/src/pages/makecolors.js",
+      "<rootDir>/src/NetworkContext.js",
+      "<rootDir>/src/hooks/helpers.js"
```
thewatermethod, line 183: Both network context and helpers are mainly if statements around low level browser APIs and how properties are named differently. It seemed unnecessary to test that

frontend/src/components/ActivityReportsTable/index.js
```diff
@@ -315,12 +313,7 @@ ActivityReportsTable.propTypes = {
...
     }),
   ).isRequired,
-  onUpdateFilters: PropTypes.func,
   tableCaption: PropTypes.string.isRequired,
 };
 
-ActivityReportsTable.defaultProps = {
-  onUpdateFilters: () => { },
-};
-
```
thewatermethod, line 325: removed as it was a passthrough that wasn't getting used in the pass through

frontend/src/components/Navigator/index.js
```diff
@@ -158,6 +164,24 @@ function Navigator({
     };
   });
 
+  const { socket, store, clearStore } = useContext(SocketContext);
+  const { user } = useContext(UserContext);
+
+  const INTERVAL_DELAY = 4000;
```
jasalisbury, line 170: I think we can remove this heartbeat, we should only need to send an update on which page a user is looking at on navigation (unless I'm missing something). Also I would _think_ the server side gets hit with a websocket "disconnected" no matter how the user leaves the page so everything should get unsubscribed when a user closes their browser (or their computer suddenly shuts off, etc..)

frontend/src/components/SocketProvider.js
```diff
@@ -0,0 +1,55 @@
...
+} from 'react';
+import PropTypes from 'prop-types';
+
+export const SocketContext = createContext();
+
+const WS_URL = process.env.REACT_APP_WEBSOCKET_URL;
+
+export const socketPath = (activityReportId, page) => {
+  if (activityReportId && activityReportId !== 'new') {
+    return `/activity-report/edit/${activityReportId}/${page}`;
```
jasalisbury, line 12: Also need to catch the "page" being `undefined`. When a report is first loaded if it doesn't have a page selected you can get a call like `ws://localhost:8080/activity-report/edit/19245/undefined`

frontend/src/components/Navigator/index.js
```diff
@@ -158,6 +164,24 @@ function Navigator({
...
+      user: user.name,
+      lastSaveTime,
+      channel: socketPath(reportId, currentPage),
+    }));
+  };
+
+  useInterval(publishLocation, INTERVAL_DELAY);
+
+  useEffect(() => {
+    clearStore();
```
jasalisbury, line 182: Do we need to explicitly "unsubscribe" from channels on page change?

jasalisbury (commented):
Biggest question is around how often and in what cases we send up messages. Another thing I noticed is I seem to be calling the websocket endpoint twice locally.
Screenshot from 2022-05-23 12-33-17
Only the last call fills up with messages.

frontend/src/components/Navigator/index.js
jasalisbury: We talked on a call about this a bit. I didn't think through the need of having a heartbeat completely. Using the redis pub/sub functionality messages are gone once broadcast, so there is no way for a user that joins a second after a message to receive that message. We might be able to use the autosave heartbeat (every 2 minutes). We will check at standup if a 2 minute lag time is alright or if we need a tighter window

thewatermethod:
Made a few changes, I think things are a little neater now but will have to hold off on final review since I'd like to that on a deployed env and they are both in use right now.

jasalisbury:
> Made a few changes, I think things are a little neater now but will have to hold off on final review since I'd like to that on a deployed env and they are both in use right now.

For after we are done with the goals rework... Dev should be free now to use. Also @kryswisnaskas did the IPT team approve our use of websockets?

kryswisnaskas:
@Josh Salisbury ***@***.***> nothing new yet after the last IPT
meeting

GarrettEHill (approved):

frontend/src/components/SocketAlert.js
```diff
@@ -0,0 +1,63 @@
...
+import PropTypes from 'prop-types';
+import { Alert } from '@trussworks/react-uswds';
+import Sticky from 'react-stickynode';
+import { useMediaQuery } from 'react-responsive';
+import UserContext from '../UserContext';
+import useArrayWithExpiration from '../hooks/useArrayWithExpiration';
+import usePageVisibility from '../hooks/usePageVisibility';
+import './SocketAlert.css';
+
+const THIRTY_SECONDS = 5 * 1000;
```
nvms, line 11: should this be renamed, or should the delay be adjusted?

frontend/src/components/Navigator/index.js
```diff
@@ -215,11 +221,13 @@ function Navigator({
     // Prevent user from making changes to goal title during auto-save.
     setSavingLoadScreen(isAutoSave);
 
+    const endDate = formEndDate && formEndDate.toLowerCase() !== 'invalid date' ? formEndDate : '';
```
nvms, line 224: instead of relying on this exact string, the following might work (untested):

```javascript
moment(formEndDate).isValid()
```

https://momentjscom.readthedocs.io/en/latest/moment/01-parsing/15-is-valid/

frontend/src/components/Navigator/index.js
```diff
@@ -703,6 +712,8 @@ Navigator.propTypes = {
       PropTypes.string,
     ]),
   }),
+  socketMessageStore: PropTypes.shape({
+  }),
```
nvms, line 716: could we update this to represent an object of shape `user: { name: string }`, unless there are additional properties I missed from `SocketAlert`?

frontend/src/components/SocketAlert.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+    return user;
+  }).join(usersToRender.length < 3 ? ' ' : ', ')} ${usersToRender.length === 1 ? 'is' : 'are'} now also working in this section. Your changes may not be saved.`;
+
+  return store && usersToRender.length ? (
+    <Sticky className="ttahub-socket-alert margin-bottom-2" top={71} enabled={!isMobile}>
+      <Alert type="warning">
+        <span>
+          <span className="usa-prose margin-top-0">{message}</span>
+          {' ' /* a very reasonable framework, react */}
```
nvms, line 47: yes, indeed

frontend/src/components/SocketAlert.js
```diff
@@ -0,0 +1,63 @@
...
+import { useMediaQuery } from 'react-responsive';
+import UserContext from '../UserContext';
+import useArrayWithExpiration from '../hooks/useArrayWithExpiration';
+import usePageVisibility from '../hooks/usePageVisibility';
+import './SocketAlert.css';
+
+const THIRTY_SECONDS = 5 * 1000;
+export default function SocketAlert({ store }) {
+  const [users, { push: pushUser }] = useArrayWithExpiration([], THIRTY_SECONDS);
+  const isPageVisible = usePageVisibility();
```
nvms, line 14: my socket connection is active and this component is aware of other users of this resource (as they are aware of my presence), but we are choosing not to alert them of my presence if I've tabbed away from this page or minimized my browser. is that what this is doing?

src/index.js
```diff
@@ -1,13 +1,66 @@
...
+        // subscribe to the channel, the function is a callback for what to
+        // do when a message is received via redis pub/sub
+        await subscriber.subscribe(channelName, (message) => {
+          ws.send(message);
+        });
+
+        // when a message is received via websocket, publish it to redis
+        ws.on('message', async (message) => {
+          const { channel, ...data } = JSON.parse(message);
+          await redisClient.publish(channel, JSON.stringify(data));
```
nvms, line 54: the only message we should ever be receiving is a payload from the client that has been stringified with `JSON.stringify`, right? if that's the case, maybe it's not necessary to do any payload validation here, like try/catching the `parse`, but my immediate thought upon seeing this was: "if parse throws, node is gonna crash". 🤷

nvms (approved):
few minor non-blocking comments, and a conflict needing to be resolved, but this looks great - well done!

frontend/src/components/SocketAlert.js
thewatermethod: correct

nvms: the reason I ask is mostly because I'm questioning if we can rely on the accuracy of this being a "there's nobody else editing this form" if, say, I just minimized my browser for a minute so that I could grab some text from some other resource, after which I'm intending to return to this form to continue editing it

although it's possible I'm making this out to be more of an issue than it really is

thewatermethod: yeah that makes sense, I think the expiration when this was added was 5 minutes instead of thirty seconds, and that was to absolutely minimize the amount of traffic that was flying around, but given how much more quickly everything comes and goes, checking that the browser is open probably no longer makes sense

src/index.js
thewatermethod: yeah, it needs to be in a try catch I guess because  we do need to extract the "channel" from the payload

AdamAdHocTeam (commented):
Wonder if this could be a candidate for an E2E test.

Apologies if you already have this or investigated the possibility.

thewatermethod:
> Wonder if this could be a candidate for an E2E test.

We dismissed that given we can only have 1 user at a time in an e2e test (at least as set up now)

AdamAdHocTeam:
> > Wonder if this could be a candidate for an E2E test.
> 
> We dismissed that given we can only have 1 user at a time in an e2e test (at least as set up now)

👍 

Might be something to revisit with Jon's impersonation down the road.

PR #876, opened 2022-04-25
Process data tweaks

jasalisbury:
## Description of change

Process data needs to be wrapped in a transaction. Also does no good to truncate audit log tables and then make audit log creating edits. Audit logs are now truncated after all other data has been updated


kryswisnaskas (approved):
👍 (confirmed, it works)

thewatermethod (approved):

PR #880, opened 2022-04-27
Goals and objectives are updated in place

jasalisbury:
## Description of change
Goals and objectives are no longer deleted and recreated when updating. Also ran into a bug where 'other-entity' reports were not properly saving objectives. There is now a single cleanup method that is called when saving goals or objectives. This ensures we don't have hanging goals.

The case where we can have hanging goals:
1. Create a recipient report and add some new goals to the report
2. Switch the report to an 'other-entity' report

Now the clean method is called whenever the 'type' of report changes
(other-entity vs recipient).

## How to test

This branch is deployed to `dev`

Create, update, and delete goals and objectives, both `recipient` and `other-entity` reports. Watch the audit log tables for `ActivityReportObjectives`, `Objectives`, and `Goals`. When updating a goal or objective there should be a single "update" audit log and not a `delete` followed by an `insert`. Make sure goals and objectives are still saving as they should. Try switching the report type (`recipient` to `other-entity` and visa versa) and verify all objectives and goals are cleaned when switching.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-671
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-738

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


thewatermethod (approved):
Looks good!

- Confirmed that goals & objectives are updated and not deleted and inserted
- Confirmed that switching back and forth between "recipient" and "other-entity" deletes any hanging goals & objectives
- Report seems to save data as normally otherwise

src/services/goals.js
```diff
@@ -382,10 +382,11 @@ export async function goalsForGrants(grantIds) {
   });
 }
 
-async function removeActivityReportObjectivesFromReport(reportId) {
+async function removeActivityReportObjectivesFromReport(reportId, objectiveIdsToRemove) {
```
AdamAdHocTeam, line 385: Should we be adding a transaction to these types of operations?

src/services/goals.js
```diff
@@ -422,46 +423,44 @@ async function removeObjectives(currentObjectiveIds) {
   });
 }
 
-async function removeUnusedObjectivesGoalsFromReport(reportId, currentGoals) {
+export async function removeUnusedGoalsObjectivesFromReport(reportId, currentObjectives) {
```
AdamAdHocTeam, line 426: Any need for a transaction.

src/services/goals.js
jasalisbury: Transactions are now handled for us automagically. You were out when the PR went in, but we now use https://sequelize.org/docs/v6/other-topics/transactions/#automatically-pass-transactions-to-all-queries. Each request get it's own singular transaction.

jasalisbury: https://github.com/HHS/Head-Start-TTADP/pull/869

AdamAdHocTeam (commented):
Overall looks good. Potentially found one issue described below.

Create a report of type recipient. Added two goals, the first goal had two objectives. The second goal had one objective. After saving and switching to other-entity I noticed one goal still remained in the DB. Going from other-entity to recipient did properly clean up the objectives.

thewatermethod (approved):
confirmed that goals with multiple objectives are removed

src/services/goals.js
AdamAdHocTeam: Nice find, seems like a large change.

AdamAdHocTeam (approved):
Looks good. :+1:

Issue has been fixed.

PR #882, opened 2022-04-27
[TTAHUB-467] apply color palette

thewatermethod:
## Description of change
UX design did an audit of the colors used in production vs. the original TTA Hub design library and found a considerable amount of discrepancies. This change implements that color palette and also introduces SASS as a way to standardize on color usage.

Because we don't simply use CSS but also use various components that require "CSS-in-JS", I am storing the color values in a JS object, and wrote a very simple node script to compile that file to a SCSS file, which will then be updated throughout the entire project. Hopefully our colors won't be changing that much, and going forward we should only be using colors from this palette

## How to test
Check the site, make sure nothing looks anomalous, review the changes requested in the Jira ticket.

## Issue(s)
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-467

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


frontend/src/colors.js
```diff
@@ -0,0 +1,40 @@
+// if adding to this, please run 'yarn makecolors' from the frontend dir
+// to rebuild the matching colors.scss file'
+const colors = {
```
GarrettEHill, line 3: Would grouping these into sub objects provide any additional value? colors.ttahub.blue or color.text.link

thewatermethod: I don't see a benefit right now, but maybe you can expand on that?

frontend/src/makecolors.js
```diff
@@ -0,0 +1,49 @@
...
+
+  const contents = `
+// STOP! Don't edit this file.
+// Instead, edit colors.js and run 'yarn makecolors'
+// from the frontend directory to generate this file'
+
+// figma reference here:
+// https://www.figma.com/file/5Fr0NKQf9MQ5WGd8BWxA6i/TTA_SmartHub-Library-09132021?node-id=0%3A14
+
+${keys.map((key, index) => {
```
jasalisbury, line 33: Can we indent this or will it cause an indention in the resulting file?

jasalisbury (approved):
Thanks for going through all these changes!

frontend/src/makecolors.js
thewatermethod: yeah the ``` `` ``` template strings respect the indentation you pass in

jasalisbury: Can we remove the indention from the other code (line 34-37) to match line 33? Took me a while to realize the block that starts at line 33 does _not_ end at line 47.

.circleci/config.yml
```diff
@@ -260,6 +260,11 @@ jobs:
     steps:
       - attach_workspace:
           at: .
+      - run:
+          name: Audit checksums of color files
+          command: |
+            chmod 744 ./checkcolorhash.sh
```
jasalisbury, line 266: More of a wondering, I know the other scripts change permissions before running, but from a quick google it seems like git should track if a file is executable. Is there a way we can remove the `chmod` line and just run the script? Wouldn't spend much time, more of a curiosity then anything to spend a lot of time on.

checkcolorhash.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
jasalisbury, line 1: :tada:

jasalisbury (approved):

.circleci/config.yml
thewatermethod: Screen Shot 2022-05-10 at 8 42 38 AM

Tried it without chmod and got this. I wonder if its more of a Circle CI thing than a git thing. Sarah Jaine would probably be able to tell us the full story.

PR #892, opened 2022-05-09
Add websockets to boundary diagram

jasalisbury:
## Description of change

Add websockets to boundary diagram

## How to test

Make sure diagram makes sense

## Questions

Is there a better way to represent websockets? They take the same path as HTTPS requests just a different protocol.


thewatermethod:
makes sense to me, do we need to denote redis as an intermediary somehow?

jasalisbury:
> makes sense to me, do we need to denote redis as an intermediary somehow?

Yeah good call, I'll add it

thewatermethod (approved):

AdamAdHocTeam (approved):
Looks good.

PR #893, opened 2022-05-10
[TTAHUB-700] handle filter select overflow

thewatermethod:
## Description of change
Previously, multiple selections on the filter multi select were handled by letting it overflow and simply hiding the overflow when the control wasn't selected. This change looks a little more graceful.

However, when digging into replacing components on the multi-select, it seemed like I would essentially have to replace everything starting at the "container" level since that's where the global focus boolean is available. So instead of going that approach, I created a dummy element that hides the current visual, is only shown on focus, doesn't accept clicks. Because this solution uses a newer CSS selector, "focus-within", it's written in such a way so that Internet Explorer users should get the same experience as before. (Progressive enhancement?)

Doing it that way means the "remove" button is not shown on any of the dummy tags, since it doesn't have access to the internals of react-select.

It's a little clunky but I didn't see a more React-forward way to do it that wouldn't have taken (me) a couple of weeks.

## How to test
Add a filter with a multiple select and make sure everything works in a sensible and complete manner, and you are prevented from doing anything that you might like to do.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-700

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


jasalisbury:
Sometimes the tag is above where it should be
Screenshot from 2022-05-11 12-46-03

1) Add FES and GS roles
2) Close the menu so `+1 more tag` is shown
3) Open the menu and remove the GS role
4) Close the menu and FES should move above and behind the box it is inside

thewatermethod:
whoops, think that's all set now

jasalisbury:
Still see the same issue

thewatermethod:
Hmm, I am unable to reproduce (was able to do so pretty easily before)

jasalisbury:
Still seeing it, on commit `5580fee9e67896874f03d66455d0f050939fe05f`.

https://user-images.githubusercontent.com/6631790/168614293-084515a1-8126-48ea-accc-8cf3b3ed3ab7.mp4

thewatermethod:
ok made a couple of other changes to address it

jasalisbury:
I'm still seeing the same issue

jasalisbury (approved):

PR #987, opened 2022-08-02
[TTAHUB-942] Fixes for Objectives

AdamAdHocTeam:
## Description of change

This PR fixes the below bugs:

- Save Goal then use context menu to edit shows both read only and edit screen.
- Fix deleting Objective would cause FK constraint error for 'ActivityReportObjective'.
- If using a existing Objective we correctly remove it from the Objective drop (also when editing via context).

## How to test

Create a goal on a AR that uses existing goals. When an objective is applied it should be removed from the drop down list. Saving the goal and returning to edit should also reflect the used objectives and no longer display both the edit and readonly views.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-942

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


src/services/goals.js
```diff
@@ -347,6 +347,17 @@ export async function goalsByIds(id) {
...
+            model: ActivityReportObjective,
+            as: 'activityReportObjectives',
+            attributes: [
+              'ttaProvided',
+            ],
+            required: false,
+            where: {
+              activityReportId,
+            },
+          },
```
GarrettEHill, line 360: Goals link to ARs directly through the ActivityReportGoals table as well

AdamAdHocTeam: This join isn't to get the AR. Its to makes sure we get the associated 'ActivityReportObjectives' associated with the 'objective'. If the user deletes the 'objective' we need to clean these up as well.

frontend/src/pages/ActivityReport/Pages/components/Objectives.js
```diff
@@ -34,7 +34,7 @@ export default function Objectives({
     defaultValues,
   });
 
-  const objectiveIds = fields ? fields.map(({ value }) => value) : [];
+  const objectiveIds = fields ? fields.map(({ value, id }) => id || value) : [];
```
thewatermethod, line 37: Is it possible to standardize the sequelize so that we don't have to "always be checking ©️" this?

thewatermethod:
Which bug specifically is this fixing?

frontend/src/pages/ActivityReport/Pages/components/Objectives.js
AdamAdHocTeam: Yes I am adding value to the objective save. This also fixes the issue of Objectives not being removed from the list.

src/services/goals.js
GarrettEHill: If the delete all objectives linking a goal to the AR, the ActivityReportGoals link should also be cleaned up.

AdamAdHocTeam:
> Which bug specifically is this fixing?

Updated in desc.

thewatermethod (approved):

PR #992, opened 2022-08-04
[TTAHUB-942] Fix multiple edit error, add custom "isPageComplete" validation to two AR pages, fix the editing of multiple goals on the RTR

thewatermethod:
## Description of change
On the AR, when creating a new goal or editing an existing one, once it is selected - editing, saving, editing, saving again would put the form into a bad state where it couldn't be edited.

On the RTR, select an existing goal, add or edit objective, save, and then "add another goal". On dev currently, this doesn't work (causes a 500 response on the server). This change fixes that by properly formatting our API response.

Change "estimated close date" to "anticipated close date"

The new goal form, the hook form registration, and the ControlledDatePicker now all agree (the activity report form sections show "complete" when they are complete and "in progress" when they are "in progress."

## How to test
Reproduce the existing on dev as above. On this branch, it should be impossible to recreate the described errors.

On a new activity report, complete the activity summary, the goal and objectives, and the next steps sections. All should show as complete in the left hand sub navigation if they are complete and incomplete if they are not. Submitting a form should still work the same way.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-942

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


kryswisnaskas:
🤔 I am still seeing a case when I am unable to edit.

While looking at the steps to reproduce, I came across this scenario:

- create a new AR with 2 recipients
- add a collaborator with 2 roles (not a central office)
- fill the Activity Summary section half way
- for goals, select an existing goal
- select an existing objective
- provide tta
- click "Save goal"
- a validation message asks to input all required fields

thewatermethod:
@kryswisnaskas I have addressed some additional issues that were causing more errors

AdamAdHocTeam (commented):
Found a possible bug.

Added existing goals and objectives saved. Then edited and was able to change the Goal text. But when saving it said its on a approved activity report and can't be changed. 

image

Seems we need to make sure to lock the goal text on edit if the goal is on an approved report. 

Same for objective title.

AdamAdHocTeam (commented):
Also seeing a strange validation message like Krys.

In my case, I half filled out the summary created my own goal and objective (not pre existing). Saved the goal then edited and made changes and saved again. I completed the remainder of the pages. When I went to review and submit I not only saw the summary page as incomplete but also the Goals and Objectives.

thewatermethod:
@AdamAdHocTeam I believe these have been addressed

frontend/src/components/GoalForm/ReadOnlyObjective.scss
```diff
@@ -0,0 +1,11 @@
+@use '../../colors.scss' as *;
```
thewatermethod, line 1: @AdamAdHocTeam FYI, we don't need to import this unless we are using the colors in the file

AdamAdHocTeam (commented):
Found a potential issue: 

1. Create a goal on a AR
2. Add three objectives and complete the info on each
3. Remove the second objective

Notice that the 'TTA Objective' and 'Topic' fields are cleared.

PR #996, opened 2022-08-05
[TTAHUB-916] Printing goals respects filters

nvms:
## Description of change

`PrintGoals` respects the filters applied at the Goals & Objectives page.

## How to test

Apply some filters and ensure that the goals in the table below are the same as the goals that are listed on the `PrintGoals` page.

https://user-images.githubusercontent.com/17074357/183964612-dc501623-f5cc-47b1-a952-f9ebfe1f68e8.mov

## Issue(s)

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/GoalsTable/GoalsTableHeader.js
+++ b/frontend/src/components/GoalsTable/GoalsTableHeader.js
@@ -41,7 +41,7 @@ export default function GoalsTableHeader({
   const showAddNewButton = hasActiveGrants && hasButtonPermissions && false;
 
   const onPrint = () => {
-    history.push(`/recipient-tta-records/${recipientId}/region/${regionId}/goals-objectives/print`, {
+    history.push(`/recipient-tta-records/${recipientId}/region/${regionId}/goals-objectives/print${window.location.search}`, {
       selectedGoals,
     });
   };
--- a/frontend/src/pages/RecipientRecord/index.js
+++ b/frontend/src/pages/RecipientRecord/index.js
@@ -182,7 +182,7 @@ export default function RecipientRecord({ match }) {
               backLink={(
                 <Link
                   className="ttahub-recipient-record--tabs_back-to-search margin-top-2 margin-bottom-3 display-inline-block"
-                  to={`/recipient-tta-records/${recipientId}/region/${regionId}/goals-objectives`}
+                  to={`/recipient-tta-records/${recipientId}/region/${regionId}/goals-objectives${window.location.search}`}
                 >
                   Back to goals table
                 </Link>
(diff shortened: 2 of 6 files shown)


thewatermethod:
Works great - just a question:

Until now, we've used the sessionStorage/localStorage APIs as enhancements - for example, the filter state works fine (or it's supposed to, at least) if the user's sessionStorage is inaccessible for some reason. 

I know that it's unlikely for a user to have sessionStorage disabled/full but I'm wondering if we want to pass the result through another API (history state via react router, maybe, more generally something that we know to be present since the site is working) instead of having it be dependent on sessionStorage working to function.

nvms:
> Works great - just a question:
> 
> Until now, we've used the sessionStorage/localStorage APIs as enhancements - for example, the filter state works fine (or it's supposed to, at least) if the user's sessionStorage is inaccessible for some reason.
> 
> I know that it's unlikely for a user to have sessionStorage disabled/full but I'm wondering if we want to pass the result through another API (history state via react router, maybe, more generally something that we know to be present since the site is working) instead of having it be dependent on sessionStorage working to function.

Yeah, that's a good point. Since the filters will I think always be available at `location.search`, maybe `PrintGoals` should just rely on that to build its query string? This would remove the dependency on any storage API.

Something like:

```typescript
const filters = queryStringToFilters(window.location.search);
const query = filtersToQueryString(filters);
```

I'm not sure the above works yet, but it illustrates what I'm thinking

thewatermethod:
> Yeah, that's a good point. Since the filters will I think always be available at location.search, maybe PrintGoals should just rely on that to build its query string? This would remove the dependency on any storage API.

```window.location.search``` seems good to me! 

I guess the other thing to be aware of is that the next thing happening on the goals & objectives page is checkboxes to manually select goals for printing, which I guess **might** not impact anything, but will replace whatever is selected via filters so it might make sense to pass the filters in a way that it can be either filters that are passed or a list of ids

nvms:
> pass the filters in a way that it can be either filters that are passed or a list of ids

For goal selection, if we're able to pass a list of ids along with filters in the URL (e.g. `&status.in[]=Closed&id.in[]=1111&id.in[]=2222` (which after talking with @GarrettEHill seems OK), this simple `location.search` solution should *just work* for both filters and goal selection, although I think the handler for this particular endpoint in its current state may require a small change -- it seems to ignore the `id` bits of the above query.

0f784ad: ditched sessionStorage and now using location.search - test will be updated shortly

thewatermethod (approved):
Works great!

PR #1004, opened 2022-08-17
[TTAHUB-758] account management / user menu / profile page

nvms:
## Description of change

This PR introduces a dropdown user menu/avatar in the header. It replaces the links to admin & logout in the sidebar and additionally includes a link to a new account management view where email preferences can be managed.

I want to merge this into an integration branch which is branched off of main.

Once this PR is approved & merged into the integration branch, the diff for the followup #1012 should be much cleaner and easier to review.

## How to test

- The login and admin links in the sidebar have been removed.
- There is a user menu dropdown in the header.
- The user menu dropdown contains a link to a new route: `/account`.
- On the `/account` page you should be able to modify your email preferences.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-758

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


src/migrations/20220819014844-user-settings-table.js
```diff
@@ -0,0 +1,85 @@
...
+    await queryInterface.dropTable('UserSettings', { transaction });
+
+    // Remove ZALUserSettings and functions.
+    await queryInterface.dropTable('ZALUserSettings', { transaction });
+    await queryInterface.sequelize.query('DROP FUNCTION IF EXISTS "public"."ZALNoUpdateFUserSettings" ()', { transaction });
+    await queryInterface.sequelize.query('DROP FUNCTION IF EXISTS "public"."ZALNoTruncateFUserSettings" ()', { transaction });
+    await queryInterface.sequelize.query('DROP FUNCTION IF EXISTS "public"."ZALNoDeleteFUserSettings" ()', { transaction });
+
+    // Remove ZALUserSettingOverrides and functions.
+    await queryInterface.dropTable('ZALUserSettingOverrides', { transaction });
```
GarrettEHill, line 80: Does this work, I thought I blocked this.

src/migrations/20220819014844-user-settings-table.js
```diff
@@ -0,0 +1,85 @@
...
+        id: {
+          type: Sequelize.INTEGER,
+          primaryKey: true,
+          autoIncrement: true,
+          allowNull: false,
+        },
+        // 'email', etc.
+        class: { type: Sequelize.STRING, allowNull: false },
+        key: { type: Sequelize.STRING, allowNull: false },
+        default: { type: Sequelize.STRING, allowNull: false },
```
GarrettEHill, line 15: can we change this to a jsonb datatype, this would allow complex settings to be saved, Like saving filters

src/migrations/20220819014844-user-settings-table.js
```diff
@@ -0,0 +1,85 @@
...
+          allowNull: false,
+          type: Sequelize.INTEGER,
+          references: { model: { tableName: 'Users' }, key: 'id' },
+        },
+        userSettingId: {
+          allowNull: false,
+          type: Sequelize.INTEGER,
+          references: { model: { tableName: 'UserSettings' }, key: 'id' },
+        },
+        value: { allowNull: false, type: Sequelize.STRING },
```
GarrettEHill, line 55: can we change this to a jsonb datatype, this would allow complex settings to be saved, Like saving filters

src/migrations/20220819014844-user-settings-table.js
nvms: yeah, it seemed to work. I only had this in here because I was constantly `undo`ing and `up`ing the migration as I was fine tuning. meant to pull this out before committing. will remove

src/migrations/20220819014844-user-settings-table.js
nvms: that's a great idea - changed here: 5fbf645

src/migrations/20220819014844-user-settings-table.js
nvms: 5fbf645

src/migrations/20220819014844-user-settings-table.js
```diff
@@ -0,0 +1,88 @@
...
+        id: {
+          type: Sequelize.INTEGER,
+          primaryKey: true,
+          autoIncrement: true,
+          allowNull: false,
+        },
+        // 'email', etc.
+        class: { type: Sequelize.STRING, allowNull: false },
+        key: { type: Sequelize.STRING, allowNull: false },
+        default: { type: Sequelize.JSONB, allowNull: false },
```
GarrettEHill, line 15: I think null needs to be allowed both as the default and the override.

kryswisnaskas (commented):
Looks great - I was able to navigate to the profile, follow links, and set the email preferences! 👍 

A couple of comments in the code and also noticed the error message doesn't clear after a successful save. Steps to recreate:

- undo the migration
- try to save the email preferences
- note the error message which is expected
- run the migration
- save the email preferences
- note both a success and error message are present

src/migrations/20220819014844-user-settings-table.js
```diff
@@ -0,0 +1,76 @@
...
+    await queryInterface.addConstraint('UserSettingOverrides', {
+      fields: ['userId', 'userSettingId'],
+      type: 'unique',
+      transaction,
+    });
+  }),
+  down: (queryInterface) => queryInterface.sequelize.transaction(async (transaction) => {
+    await queryInterface.dropTable('UserSettingOverrides', { transaction });
+    await queryInterface.dropTable('UserSettings', { transaction });
+  }),
```
kryswisnaskas, line 75: In this case, we do want the "down" to be functional, so we need to add the removal of the ZALUserSetting and ZALUserSettingOverrides and functions for production.

Previously, we made an exception to omit the "down" since it wasn't very feasible to write it, but it was in a v. lengthy and complex migration.

nvms: Got it! Restored this here: 7d4fe57

src/services/userSettings.js
```diff
@@ -0,0 +1,193 @@
...
+
+  return Promise.all([
+    UserSettingOverrides.destroy({ where: { userId, userSettingId: { [Op.in]: settingIds } } }),
+    UserSettingOverrides.bulkCreate(save.map(({ userSettingId, value }) => ({
+      userId,
+      userSettingId,
+      value: sequelize.cast(JSON.stringify(value), 'jsonb'),
+    }))),
+  ]);
+};
```
kryswisnaskas, line 71: I know we use this pattern in other places (implemented during an accelerated initial MVP schedule), but ideally, we would change it to updates, if possible. Additionally, with the implementation of the audit log, we end up with a less readable output.

nvms: I refactored this to `update` where possible, `insert` where necessary and `delete` just as it did before. I think this should result in more readable audit log entries. Does it look okay? c2107d60

kryswisnaskas: Thanks. The code needs to be modified to also include "key" in the `updateable` or the filtering can use `userSettingId `instead.

Sequelize also offers `findOrCreate `or `upsert `that could be helpful here.

nvms: thanks for catching this! I believe this is now fixed. I wrote some additional tests to verify this behavior works as intended.

frontend/src/components/NavLink.js
```diff
@@ -2,9 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { NavLink as Link } from 'react-router-dom';
 
-function NavLink({ children, to, exact }) {
+function NavLink({
+  children, to, exact, fn,
+}) {
   return (
-    <Link className="usa-nav__link" to={to} activeClassName="usa-current" exact={exact}>
+    <Link className="usa-nav__link" to={to} activeClassName="usa-current" exact={exact} onClick={fn}>
```
thewatermethod, line 9: Just bringing this comment from #1017 over here so it doesn't get lost:

For the "/logout" path, we should either make a route that handles logging out when navigated to or switch this to be a button. We don't want to have links that act like buttons for accessibility reasons, and I was able to get split second 404's this way so it might also mess up or at least befuddle our analytics in that way

kryswisnaskas (approved):
Looks good! 👍

PR #1002, opened 2022-08-16
[TTAHUB-853] New AR topics migration

nvms:
## After `up`

Old topics should correctly map to the new topic and have good `deletedAt` values:

Screen Shot 2022-08-17 at 10 42 33 AM

Screen Shot 2022-08-17 at 10 42 40 AM

The new topic name should be replaced in the `topics` column in `ActivityReports` table:

Screen Shot 2022-08-17 at 10 43 28 AM

## After `down`

`topics` column in `ActivityReports` table reverted to original values:

Screen Shot 2022-08-17 at 10 44 06 AM

`mapsTo` and `deletedAt` restored to original values in `ActivityReports` table:

Screen Shot 2022-08-17 at 10 44 20 AM

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-853

### Every PR

- [x] Meets issue criteria
- [ ] JIRA ticket status updated
- [ ] Code is meaningfully tested
- [ ] Meets accessibility standards (WCAG 2.1 Levels A, AA)
- [ ] API Documentation updated
- [ ] Boundary diagram updated
- [ ] Logical Data Model updated
- [ ] [Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions

### Production Deploy

- [ ] Staging smoke test completed

### After merge/deploy

- [ ] Update JIRA ticket status

PR #1007, opened 2022-08-18
[TTAHUB-853] New AR topics - frontend

nvms:
## Description of change

This is the frontend change to be paired with #1002 and should be merged along with it.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-853

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/Constants.js
+++ b/frontend/src/Constants.js
@@ -147,7 +147,7 @@ export const OTHER_ENTITY_TYPES = [
 // - src/constants.js
 export const TOPICS = [
   'Behavioral / Mental Health / Trauma',
-  'Child Assessment, Development, Screening',
+  'Child Screening and Assessment',
   'CLASS: Classroom Organization',
   'CLASS: Emotional Support',
   'CLASS: Instructional Support',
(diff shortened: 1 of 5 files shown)


frontend/src/Constants.js
```diff
@@ -147,7 +147,7 @@ export const OTHER_ENTITY_TYPES = [
 // - src/constants.js
 export const TOPICS = [
   'Behavioral / Mental Health / Trauma',
-  'Child Assessment, Development, Screening',
```
thewatermethod, line 150: these will also need to be updated in ```src/constants.js```(I don't see that change in the other PR either)

I don't want to expand the scope of this PR, but we have a topics endpoint now - perhaps we can write a Jira to switch out of these hardcoded thingamabobs and use those where appropriate.

nvms: thanks - somehow it looks like I also missed a whole bunch of other references to old topic names that I think should be updated, which I've done here: 6fbd748f34a5ec28e1eb7eb84262cbf97d7d0e0c

hopefully I caught them all in that commit ^^ tests appear to be passing, so hopefully that means 👍 

> perhaps we can write a Jira to switch out of these hardcoded thingamabobs and use the endpoint/info from DB

:100: agreed, good idea

thewatermethod (approved):

PR #1020, opened 2022-08-30
[TTAHUB-981] Fix OE Objectives Saving and Read Only View

AdamAdHocTeam:
## Description of change

This ticket fixes two issues:

1. Error when saving objectives on a other-entity report.
2. Display saved goals in a read only view.

**Update 0907:** To return to the report and see the objectives in read only mode you will need to comment out validation for the fields we currently do NOT save (roles, topics, and resources) - GoalValidator.js -unfinishedObjectives().

## How to test

1. Create an other entity report.
2. Saving objectives on the report should no longer error.
3. Once the objectives are saved you should be redirected to a read only view.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-981

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


thewatermethod:
I ran into a white screen of death

Steps to reproduce:
1) Create a report and add a couple of objectives. I had two other entity recipients and I added resources to one and not to the other.
2) Leave report (go back to activity reports landing page)
3) Go to edit report, white screen of death

I tried twice, the second time clearing local storage before trying to edit.

src/services/activityReports.js
```diff
@@ -382,6 +382,16 @@ export async function activityReportAndRecipientsById(activityReportId) {
       {
         model: Objective,
         as: 'objectivesWithoutGoals',
+        include: [
```
thewatermethod, line 385: Will test shortly, but shouldn't this also include files & resources?

thewatermethod:
Found a possible bug
- Create a new other entity report
- Add an objective, save, you are in supporting attachments
- Return to the /activity-reports landing page
- Return to the activity report, go to the goals & objectives tab, the objective is empty (nothing in the title, tta provided, etc.) I would expect to see no topics and that stuff, which is fine, but I would expect to see a title. The objectivesWithoutGoals array returned from the API is empty.

AdamAdHocTeam:
> * jective is empty (nothing in the title, tta provided, etc.) I would expect to see no topics and that stuff, which is fine, but I would expect to see a title. The objectivesWithoutGoals array returned from the API is empty.

I cannot reproduce this issue. I might be doing something wrong. Also please try clearing you local storage. I had a weird issue yesterday were I think it was using the local storage data.

thewatermethod (approved):

PR #1029, opened 2022-09-09
[TTAHUB-996] Prevent Duplicate Goals when Saving Draft

AdamAdHocTeam:
## Description of change

We need to prevent creating duplicate goals and objectives if an auto save occurs when the user is still editing.

We also need to add saving of draft other-entity objectives while preventing the same case.

## How to test

Create goals and objectives for both **recipient** and **other-entity** reports. Allow auto save to occur at various steps (CRUD) to ensure we are not creating duplicate goals and objectives in the database. You should be able to finalize goal and objective creation and submit the report as before.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-996

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/Navigator/__tests__/index.js
+++ b/frontend/src/components/Navigator/__tests__/index.js
@@ -282,4 +282,34 @@ describe('Navigator', () => {
     fetchMock.post('/api/activity-reports/goals', []);
     expect(fetchMock.called('/api/activity-reports/goals')).toBe(false);
   });
+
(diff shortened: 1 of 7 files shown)


thewatermethod:
Seems to work really well, only thing I think we need to add is that the "autosaved on" message on the left hand nav bar should be updated when the autosave fires on goals & objectives

thewatermethod (approved):

PR #1035, opened 2022-09-16
TTAHUB-992 readonly bug / CTA button order / "add next step"

nvms:
## How to test

- The spacing around "Add next step" buttons should match the figma prototype:

Screen Shot 2022-09-19 at 11 33 47 AM

- The `TTA objective` textarea should be readonly when:
  - The objective's parent goal status is 'Closed'.
  - The objective's initial status when loaded is 'Completed' or 'Suspended'.
  - When the objective is on an approved report.

Possible readonly/writable combinations are tested:

https://github.com/HHS/Head-Start-TTADP/blob/5cc9c442012d36ec868d3179ac6d3a4d6ed9981b/frontend/src/pages/ActivityReport/Pages/components/__tests__/ObjectiveTitle.js#L55-L73

- The CTA buttons on the review and submit stage should be swapped:

Screen Shot 2022-09-16 at 2 52 45 PM

.. is now ..

Screen Shot 2022-09-16 at 2 54 23 PM

## Issue(s)

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.scss
+++ b/frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.scss
@@ -6,8 +6,7 @@
 }
 
 .usa-button.ttahub-next-steps__add-step-button {
-  margin-top: 2px;
-  margin-bottom: 12px;
+  margin-top: 0;
 }
 
 /* Outline a blank step onBlur or form Validation */
--- a/frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.scss
+++ b/frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.scss
@@ -6,7 +6,8 @@
 }
 
 .usa-button.ttahub-next-steps__add-step-button {
-  margin-top: 0;
+  margin-top: 2px;
+  margin-bottom: 12px;
 }
 
 /* Outline a blank step onBlur or form Validation */
--- a/frontend/src/pages/ActivityReport/Pages/Review/Submitter/Draft.js
+++ b/frontend/src/pages/ActivityReport/Pages/Review/Submitter/Draft.js
@@ -136,16 +136,7 @@ const Draft = ({
         <div className="margin-top-3">
           <ApproverStatusList approverStatus={approverStatusList} />
         </div>
-        <Button
-          outline
-          type="button"
-          onClick={async () => {
-            await onSaveForm(false);
-            updateShowSavedDraft(true);
-          }}
-        >
-          Save Draft
-        </Button>
+        <Button disabled={!connectionActive} type="submit">Submit for approval</Button>
         { !connectionActive && (
         <Alert type="warning" noIcon>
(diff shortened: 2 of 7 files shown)


thewatermethod (commented):
Everything looks great! Just a couple of comments.

> Next Steps - "add next step" should push the page down when displayed, keeping the space between the sections consistent.

I see why you went with smooth scrolling here (the jira says "push the page down"), but I believe what this is requesting is that the "add next step" button should have a margin on it so that the spacing rules are preserved. I checked [the prototype](https://www.figma.com/proto/rJGMygyqMz8jjCcoEpzcnK/Activity-Reports?node-id=3261%3A120890&starting-point-node-id=3261%3A120890&show-proto-sidebar=1) but it looks like they aren't conditionally hiding and showing the button there. Kelly might have some clarification, she made the prototype and wrote this ticket.

On the smooth scrolling, I would consider maybe canning that for now, unless that is exactly what the UX folks want and they can't be dissuaded. Here's some issues I see with it.

- It doesn't work properly on Firefox (pretty minor issue, although it will forever annoy me)
- Right now, it will jump past empty date fields if you leave the page and come back
- It will partially obscure the page title on page load, at least when I tried it
- It's not doing anything to manage focus, so someone using a keyboard to navigate will likely find it disorienting, especially when they are pulled back up to the top
- On that subject, we should be "opt in" for animations. [In Javascript, we can test to see if the user has set prefer reduced motion to off](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) and only do the scrolling then. It's a pretty minor thing in this case, since the scroll effect is so small, but I feel like its a pretty easy best practice to observe.

thewatermethod:
Sorry, I missed something before!

>(Bug) When creating an AR, the objective text area is changing to editable when the status changes from in progress to complete

I think what this bug on the Jira is saying is that we *don't* want someone in the middle of creating an objective to set it to read only, I.E. the objective title is only read only if the objective is loaded from the API as "Completed," not if someone is creating it for the first time

frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.scss
```diff
@@ -6,11 +6,12 @@
 }
 
 .usa-button.ttahub-next-steps__add-step-button {
-  margin-top: 0;
+  margin-top: 2px;
+  margin-bottom: 12px;
 }
```
thewatermethod, line 11: is this not-utility classes because of a CSS specificity thing (just curious)? also, weird that that they aren't using the standard margin sizes, which are generally divisible by 8

nvms: no -- I think I was just trying to get it to be a pixel-perfect match with the figma prototype, but using `margin-bottom-2` on the button and reverting the styles shown above gets it close enough 🙃 06bb7cbd

thewatermethod (approved):
🧨

PR #1045, opened 2022-09-27
fix - don't show more than one alert at a time

nvms:
## Description of change

As mentioned in the 9/27 standup, we don't want to show more than one alert on the page at a time. In this PR, the verification result banner is shown until the email preferences form alert is shown, which is done by piggybacking onto the form's `onSubmit` method.

## Issue(s)

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


kryswisnaskas:
Looks good  - the banner now shows under "Email Preferences"

I run into an error though (locally)
`2022-09-27T18:19:43.552Z AUDIT error:   SERVICE:USER - UNEXPECTED ERROR - Error: Token already validated`

Steps to reproduce:
- login with an unverified user
- verify email
- the success banner shows up under the "Email Preferences" which is what we want 👍 
- immediately modify the preferences to subscribe to a weekly digest
- a success banner shows 👍 
- refresh the page
- note the error

nvms:
we talked on slack about this, but to recap here: the UI may end up requesting that the token be validated more than once and in the event that it does, and the token has already been validated, it is okay to reply with a 200 instead of a 4xx so that the UI shows a success message (again). it does not feel like an error case.

eccc9e43

kryswisnaskas (approved):
👍

PR #1083, opened 2022-10-14
Update import goals script

kryswisnaskas:
## Description of change
Updates the goal import script to reflect the schema changes. Topics and Roles are no longer looked at since they belong to individual objectives. The status of existing goals is updated, but only if the status from RTTAPA is more recent.

## How to test
Looks at the code, make sure the tests pass.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1045

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


src/tools/importPlanGoals.js
```diff
@@ -135,26 +100,50 @@ export default async function importGoals(fileKey, region) {
...
               // eslint-disable-next-line no-console
-              console.log(`Couldn't find grant: ${fullGrant.number}. Exiting...`);
+              logger.error(`Couldn't find grant: ${fullGrant.number}. Exiting...`);
               throw new Error('error');
             }
             const grantId = dbGrant.id;
-            await Goal.findOrCreate({
-              where: { grantId, name: goal.name, isFromSmartsheetTtaPlan: true },
-              defaults: goal,
+            const dbGoal = await Goal.findOne({
```
AdamAdHocTeam, line 107: What if we have multiple goals with the same grant and text but some are marked as complete? Do we need to take into account the goal status?

src/tools/importPlanGoals.js
```diff
@@ -84,39 +81,7 @@ export default async function importGoals(fileKey, region) {
...
-                      if (!cleanRoleTopics.some((e) => e.roleId === roleId
-                                                      && e.topicId === topicId)) {
-                        cleanRoleTopics.push({ roleId, topicId });
-                      }
-                    }
-                  }
-                }
-              }
-            } else {
+            if (column !== 'topics') { // ignore topics
```
AdamAdHocTeam, line 84: As we discussed there is no way to map Topics to objectives.

src/tools/importPlanGoals.js
AdamAdHocTeam: Also assuming we don't care about this field anymore 'isFromSmartsheetTtaPlan'.

src/tools/importPlanGoals.js
```diff
@@ -135,26 +130,46 @@ export default async function importGoals(fileKey, region) {
...
+                },
+                {
+                  // where: { id: dbgoal.id },
+                  individualHooks: true,
+                },
+              );
+              // determine if status needs to be updated
+              await updateStatus(goal, dbGoal);
+            } else {
+              const newGoal = await Goal.create({ grantId, ...goal, isFromSmartsheetTtaPlan: true });
```
thewatermethod, line 166: Could we set "createdVia" here to imported?

src/tools/importPlanGoals.js
kryswisnaskas: > What if we have multiple goals with the same grant and text but some are marked as complete? Do we need to take into account the goal status?

Yes. The status will only be updated if it is newer.

kryswisnaskas: > Also assuming we don't care about this field anymore 'isFromSmartsheetTtaPlan'.

We still would want this for historical reasons.

hardwarehuman (approved):

PR #1090, opened 2022-10-17
[TTAHUB-1029] only certain urls and other bugs

thewatermethod:
## Description of change
- Improved URL validation with a regex (see ticket for specs)
- Reorder objective status choices (moved suspended before closed)
- On the activity report, don't derive UI state from the status dropdown (fixes: "AR goals and objectives tab - TTA resources disappears when "suspend" status is chosen")
- Display validation on next steps when the fields are blank and the "Add next steps" button is clicked

## How to test
- Enter some URLs and make sure that the conditions are satisfied
- Confirm the objective statuses are in the right order
- Confirm that you can set an objective to suspended without losing sections of the UI
- Go the next steps page and try and add a next step without entering any data, you will see the validation messages

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1029

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


kryswisnaskas (commented):
I was able to verify the objective status order is now fixed as well as the resource fields not disappearing when the objective's status is changed to "Suspended". 👍 

The next steps validation now shows up when trying to create a new step without filling in values for the first one. 👍 

On the next steps page we still have a case when the first step is filled out correctly and then "Add next step" link clicked, the date from the previous step is now flagged as invalid.

The resource validation is better, but we still accept things like the following. Is that right?

image

thewatermethod:
@kryswisnaskas did you have to do anything weird to get that URL in there? It doesn't pass the validation for me

Screen Shot 2022-10-18 at 11 06 57 AM

AdamAdHocTeam (commented):
Do we want to prevent URL's that end with a period '.'
image

These also not sure if we care

image

Other items look good.

AdamAdHocTeam (approved):
If you have agreed with Angela on the acceptable URL's then I think we are good.

🔧

PR #1098, opened 2022-10-20
[TTAHUB-1070] Add a global try catch within the transaction with rollback

GarrettEHill:
## Description of change
We are concerned that there are edge cases where an uncaught error in the business logic is occurring, allowing allowing an incomplete operation on the database to occur. 

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1070

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy


src/routes/transactionWrapper.js
```diff
@@ -1,9 +1,22 @@
...
 export default function transactionWrapper(originalFunction) {
   return async function wrapper(req, res, next) {
-    return sequelize.transaction(async () => {
-      const result = await originalFunction(req, res, next);
+    return sequelize.transaction(async (t) => {
+      let result;
+      try {
+        result = await originalFunction(req, res, next);
+      } catch (err) {
+        t.rollback();
```
thewatermethod, line 17:
```suggestion
        await t.rollback();
```

thewatermethod: looks like as an alternative you can just throw an error 
https://sequelize.org/docs/v6/other-topics/transactions/#throw-errors-to-rollback

thewatermethod (dismissed):

AdamAdHocTeam (commented):
I think we first need to verify what happens if an unhandled exception occurs. 

1. Do database changes already get rolled back if an unhandled error occurs? 
2. Is the user made aware of the error? 
3. Consider a way for us to get ahold of the error message, call stack, user, function, etc.? It might make more sense to have these written to a table instead of having the e-mailed (to spam folder).

src/routes/transactionWrapper.js
```diff
@@ -1,9 +1,22 @@
...
 export default function transactionWrapper(originalFunction) {
   return async function wrapper(req, res, next) {
-    return sequelize.transaction(async () => {
-      const result = await originalFunction(req, res, next);
+    return sequelize.transaction(async (t) => {
+      let result;
+      try {
+        result = await originalFunction(req, res, next);
+      } catch (err) {
+        await t.rollback();
```
kryswisnaskas, line 17: This can be deleted since we are using managed transactions. At this point, the transaction is already rolled back.

kryswisnaskas (commented):
Looks good!  This will catch any previously uncaught errors. 👍 

Only one comment in the code.

kryswisnaskas:
> 1. Do database changes already get rolled back if an unhandled error occurs?
> 2. Is the user made aware of the error?
> 3. Consider a way for us to get ahold of the error message, call stack, user, function, etc.? It might make more sense to have these written to a table instead of having the e-mailed (to spam folder).

Valid questions.
1. Yes, since we have the `transactionWrapper` On error, the transaction should roll back.
2.  Users get a generic error.
3.  Since now we will be catching the previously uncaught errors, sequelize errors will be written to "RequestErrors" along with the body of the request. Perhaps we can plan for a separate PR to expand on that if needed.

GarrettEHill:
With the new changes, the error handler logs should now be out of band of the transaction and thus not rolled back on errors.

thewatermethod (approved):
No issues using the app & saving data, confirmed that sequelize errors are written to the request errors table and not rolled back.

src/routes/transactionWrapper.js
```diff
@@ -1,9 +1,22 @@
...
+
+const logContext = {
+  namespace,
+};
 
 export default function transactionWrapper(originalFunction) {
   return async function wrapper(req, res, next) {
     return sequelize.transaction(async () => {
-      const result = await originalFunction(req, res, next);
+      let result;
```
thewatermethod, line 13: knowing that the transaction functionally handles all the errors as a try/catch, why do we want to change this file? What purpose is this serving? I'm concerned that adding that "Wrapper" namespace would make it harder to track down an error in the future

GarrettEHill: So your suggestion is to remove line 11 and line 22, correct?

GarrettEHill: Yes to get the error into the handleErrors function

src/routes/transactionWrapper.js
```diff
@@ -1,9 +1,22 @@
...
+      try {
+        result = await originalFunction(req, res, next);
+      } catch (err) {
+        await handleErrors(req, res, err, logContext);
+        throw err;
```
kryswisnaskas, line 18:
```suggestion
```

src/routes/transactionWrapper.js
```diff
@@ -1,9 +1,22 @@
...
+      }
```
kryswisnaskas, line 19: We don't want to throw here, since the error handler already takes care of any error and use the passed in `namespace ` in the log.

kryswisnaskas:
Looks good. Only one comment.

src/routes/transactionWrapper.js
GarrettEHill: we need it to throw to force the transaction to fail.

src/routes/transactionWrapper.js
kryswisnaskas: At this point, the transaction should have already been rolled back.

src/routes/transactionWrapper.js
kryswisnaskas: > knowing that the transaction functionally handles all the errors as a try/catch, why do we want to change this file? What purpose is this serving? I'm concerned that adding that "Wrapper" namespace would make it harder to track down an error in the future

We have a mixture of cases in the code where we already handle any errors and the ones that don't. Where we handle them correctly, the namespace will be correct (non-wrapper). It'd be best to refactor the individual functions that are missing try/catch, but that's a bigger change.

src/routes/transactionWrapper.js
kryswisnaskas: If we were to leave it, in case the body of the try/catch throws an unexpected exception, we would crash the server.

AdamAdHocTeam: I think I have adjusted this.

src/routes/transactionWrapper.js
```diff
@@ -15,7 +15,6 @@ export default function transactionWrapper(originalFunction) {
         result = await originalFunction(req, res, next);
       } catch (err) {
         await handleErrors(req, res, err, logContext);
-        throw err;
```
GarrettEHill: This will not work as without throwing the transaction will not be rolled back.

src/routes/transactionWrapper.js
```diff
@@ -1,13 +1,28 @@
...
-    });
+    let error;
+    try {
+      return sequelize.transaction(async () => {
+        let result;
+        try {
+          result = await originalFunction(req, res, next);
+        } catch (err) {
+          error = err;
+          throw err;
```
kryswisnaskas, line 20: This will not crash the server, wouldn't the inner try/catch simply pass the caught error to the outer catch without providing any other functionality?  (the transaction gets rolled back regardless)

GarrettEHill: correct, but it does save off the original error so it can be correctly logged and not just logging a failed transaction.

kryswisnaskas: We really don't need it, since the err would be caught by the outer catch. It would make more sense if there was some other logic in the inner catch, e.g. a different logging message. 
 Since we need this PR quickly, I am fine with leaving it and refactoring later.

kryswisnaskas (approved):
👍

src/routes/transactionWrapper.js
hardwarehuman: I think the idea here is that the error coming out of the transaction failure wouldn't necessarily look the same as the error coming out of the function that caused the error within the transaction.

hardwarehuman (approved):

src/routes/transactionWrapper.js
thewatermethod: Perhaps given the confusion it would be nice to add a comment explaining the intent

PR #1104, opened 2022-10-26
[TTAHUB-1076] Show correct buttons on form when editing existing goals

thewatermethod:
## Description of change
- Editing an existing goal on the RTR now shows different buttons (save and cancel). Save simply saves and returns the user to the RTR, cancel does the same without saving first. 
- The tooltips on the Activity Reports table now open down if they are in the first five rows, the bottom five rows open in the top direction. This prevents tall tooltips being cut off at the top/bottom

## How to test
- Ensure the above works as expected

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1076

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/ActivityReportsTable/ReportRow.js
+++ b/frontend/src/components/ActivityReportsTable/ReportRow.js
@@ -119,7 +119,7 @@ function ReportRow({
         </Link>
       </th>
       <td>
-        <TooltipWithCollection collection={recipients} collectionTitle={`recipients for ${displayId}`} />
+        <TooltipWithCollection collection={recipients} collectionTitle={`recipients for ${displayId}`} position={openMenuUp ? 'top' : 'bottom'} />
       </td>
       <td>{startDate}</td>
       <td>
--- a/frontend/src/components/ActivityReportsTable/index.js
+++ b/frontend/src/components/ActivityReportsTable/index.js
@@ -288,7 +288,7 @@ function ActivityReportsTable({
                   report={report}
                   handleReportSelect={handleReportSelect}
                   isChecked={reportCheckboxes[report.id] || false}
-                  openMenuUp={index > displayReports.length - 1}
+                  openMenuUp={index > displayReports.length - 5}
                   numberOfSelectedReports={numberOfSelectedReports}
                   exportSelected={handleDownloadClick}
                 />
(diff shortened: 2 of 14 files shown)


kryswisnaskas (commented):
Verified that the buttons change now when editing "In Progress" and "Not Started" goals 👍 

Tooltip also looks good! 🎉 

Found a few things:
1. certain goals don't save. To reproduce:
- select RTR's first recipient in region one (Abbot, Ondricka & Bracus)
- edit goal G-61363
- click Save (note the error)

2. There is no space between "Region" and it's number on the goal edit form.
3. When attempting to edit a Closed goal, topic input field shows up as editable and required, but there are no buttons. Also, there is "Invaliddate" under date. (This might be a separate PR if it's too much since it is present on Dev as well)
4. We should prevent a scenario where someone with read permissions to multiple regions attempts to edit goals where they don't have write permissions in. We could bring this up at the standup.

thewatermethod:
>  certain goals don't save. To reproduce:

I don't have that recipient in my local database. Is the error "Enter a valid date"? I would think if a user is opening up a goal to edit, they should have to fill out all the fields, even if the goal already existed (unless it's closed, etc)

thewatermethod:
> We should prevent a scenario where someone with read permissions to multiple regions attempts to edit goals where they don't have write permissions in. We could bring this up at the standup.

We have a region id in the URL, I will open a separate PR to handle permissions on the frontend based on that.

> There is no space between "Region" and it's number on the goal edit form.
When attempting to edit a Closed goal, topic input field shows up as editable and required, but there are no buttons. Also, there is "Invaliddate" under date. (This might be a separate PR if it's too much since it is present on Dev as well)

These have been addressed. Just let me know if my understanding of the error in my previous comment is correct when you get a chance, @kryswisnaskas

thewatermethod:
[This is the PR](https://github.com/HHS/Head-Start-TTADP/pull/1108) to handle the permissions more explicitly

thewatermethod:
[This PR makes goal end date no longer a required field](https://github.com/HHS/Head-Start-TTADP/pull/1114)

AdamAdHocTeam (commented):
This might have already been pointed out in Krys's review. If I edit and in progress goal add some topics and end date then save I get the below:

image

AdamAdHocTeam (approved):
Looks good.

PR #1111, opened 2022-10-28
[TTAHUB-1074] fix delete button functionality from edit goals page

thewatermethod:
## Description of change
In the distant past, when the manage goals page was constructed, I was asked to disable deleting of goals from the manage goals page so that things couldn't get too wild while we were testing.

It was just a few nights ago that I awoke with the realization that I had never rewired things. I mopped the sweat from my brow, dusted off this code and discovered it was largely still functional, except for needing a few small tweaks. I removed references to the GrantGoals table. I refactored it to remove two goals at once where appropriate. I believe it now functions as intended. 

## How to test
The only way to access it is by creating a goal from the RTR and then immediately deleting it. There is no "read only" screen for editing goals any longer. This change also effects (mildly) the delete goals function from the AR, although only on the frontend.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1074

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/goals.js
+++ b/src/services/goals.js
@@ -25,6 +25,13 @@ import {
   cacheGoalMetadata,
   destroyActivityReportObjectiveMetadata,
 } from './reportCache';
+import { auditLogger } from '../logger';
+
+const namespace = 'SERVICE:GOALS';
+
+const logContext = {
+  namespace,
+};
 
 const OPTIONS_FOR_GOAL_FORM_QUERY = (id, recipientId) => ({
   attributes: [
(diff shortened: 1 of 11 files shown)


src/services/goals.js
```diff
@@ -1553,95 +1560,70 @@ export async function createOrUpdateGoalsForActivityReport(goals, reportId) {
...
+          where: {
+            id: goalIds,
+          },
+          as: 'goals',
+        },
+      ],
+    });
+
+    const isOnReport = reportsWithGoal.length;
+    if (isOnReport) {
```
AdamAdHocTeam, line 1581: This would be all or nothing?

AdamAdHocTeam (commented):
If I create a new goal from RTR then try to remove it I get the following error:

image

src/services/goals.js
AdamAdHocTeam: I'm not saying this is wrong just food for thought.

thewatermethod: Yeah I think that makes sense, it is only triggered on the button press in the RTR, and I don't think we want to handle only deleting one yet since we treat those goals as one goal everywhere in the UI. Maybe if it turns out to be a problem in the real world.

thewatermethod:
> If I create a new goal from RTR then try to remove it I get the following error:

I have fixed this error

AdamAdHocTeam (commented):
From RTR if I create two goals when I attempt to save the second goal I get the below error:

image

thewatermethod:
> From RTR if I create two goals when I attempt to save the second goal I get the below error

I have fixed this

AdamAdHocTeam (commented):
I'm able to save a draft goal from RTR without text:

image

AdamAdHocTeam (approved):
All my issues have been fixed.

PR #1118, opened 2022-11-01
[TTAHUB-857] Add new paging to goals and update select

AdamAdHocTeam:
## Description of change

- Persist 'select all' check box until they unselect it. 
- Always show select all pages alert if 'select all' is checked and all pages aren't selected.
- Paging now has 'per page' selector in 10, 25, 50, and all.
- Changing the rows per page brings you back to the first page.

## How to test

Make sure the design is matched and the above are working as expected.

https://www.figma.com/proto/WoAULVDXnKuVx5t1y9jgzW/RTTAPA-Goals-Plan?node-id=367%3A38184&scaling=min-zoom&page-id=210%3A18964&starting-point-node-id=367%3A36637&show-proto-sidebar=1

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-857

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/SelectPagination.js
+++ b/frontend/src/components/SelectPagination.js
@@ -3,19 +3,7 @@ import PropTypes from 'prop-types';
 import Pagination from 'react-js-pagination';
 import { Dropdown } from '@trussworks/react-uswds';
 
-export function renderTotal(title, offset, perPage, activePage, count) {
-  const from = offset >= count ? 0 : offset + 1;
-  const offsetTo = perPage * activePage;
-  let to;
-  if (offsetTo > count) {
-    to = count;
-  } else {
-    to = offsetTo;
-  }
-  return `Showing ${from}-${to} of ${count} ${title.toLowerCase()}`;
-}
-
-export const SelectPagination = ({
+function SelectPagination({
   title,
   offset,
   activePage,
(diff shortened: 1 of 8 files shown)


frontend/src/components/SelectPagination.js
```diff
@@ -0,0 +1,87 @@
...
+  let to;
+  if (offsetTo > count) {
+    to = count;
+  } else {
+    to = offsetTo;
+  }
+  return `Showing ${from}-${to} of ${count} ${title.toLowerCase()}`;
+}
+
+export const SelectPagination = ({
```
thewatermethod, line 18: no need to export twice (you are exporting as default below, which is our normal practice)

AdamAdHocTeam:
> It all looks pretty good. A couple of questions
> 
> * Should the "preview and print selected" button say that when there are no goals selected?
> * If you have no goals selected, should it only print the goals that are on that page or every goal (which is what it does now)?
> * Should we add a test on the page size changes from the "GoalDataController" level or above? Right now none of the stuff in there is being tested (like perPageChange, for example). If you want to leave it for later, that's fine, shouldn't block this PR.

I think all of these have been addressed.

PR #1124, opened 2022-11-03
[TTAHUB-1069] Add AR App Loader

AdamAdHocTeam:
## Description of change

When creating an AR we want to put a loader over the entire application.

## How to test

We should display a App loader when performing the following AR actions.
- 'Save and Continue'
- 'Save Draft'
- 'Save Goal' 
- Jumping between AR Pages via NAV

It should NOT occur during an auto save.

The AR creation process should work as it did before.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1069

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/Navigator/index.js
+++ b/frontend/src/components/Navigator/index.js
@@ -79,7 +79,7 @@ function Navigator({
   const selectedObjectivesWithoutGoals = watch('objectivesWithoutGoals');
 
   // App Loading Context.
-  const { isLoading, setLoadingText, setIsLoading } = useContext(AppLoadingContext);
+  const { isAppLoading, setIsAppLoading, setAppLoadingText } = useContext(AppLoadingContext);
   const [isAutoSave, setIsAutoSave] = useState(false);
   const [isGoalFormClosed, toggleGoalForm] = useState(selectedGoals.length > 0);
 
@@ -93,9 +93,9 @@ function Navigator({
   );
 
   const setSavingLoadScreen = () => {
-    if (!isAutoSave && !isLoading) {
-      setLoadingText('Saving');
-      setIsLoading(true);
+    if (!isAutoSave && !isAppLoading) {
+      setAppLoadingText('Saving');
+      setIsAppLoading(true);
     }
   };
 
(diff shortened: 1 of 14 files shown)


frontend/src/components/Navigator/index.js
```diff
@@ -420,81 +444,81 @@ function Navigator({
...
-                  )}
-                </DismissingComponentWrapper>
-              </Container>
-            )}
+                      <div className="display-flex">
+                        {showSaveGoalsAndObjButton
+                          ? (
+                            <>
+                              <Button className="margin-right-1" type="button" disabled={isAppLoading} onClick={onGoalFormNavigate}>{`Save ${isOtherEntityReport ? 'objectives' : 'goal'}`}</Button>
+                              <Button className="usa-button--outline" type="button" disabled={isAppLoading} onClick={isOtherEntityReport ? onSaveDraftOetObjectives : onSaveDraftGoal}>Save draft</Button>
```
nvms, line 491: I think the onClick handler should be more like:

```javascript
onClick={isOtherEntityReport ? () => onSaveDraftOetObjectives(false) : () => onSaveDraftGoal(false)}
```

because currently, `onSaveDraftGoal` will receive the click event for its `isAutoSave` argument which is truthy and so the overlay is not shown

Screen Shot 2022-11-04 at 9 19 02 AM

nvms (changes requested):
found a small bug, otherwise this looks great

frontend/src/components/Navigator/index.js
AdamAdHocTeam: Nice catch, will change.

AdamAdHocTeam: I have changed the code if you can have another look 👀

nvms:
we just talked on slack about this, so leaving it here as well:

it would be nice if there was logic to detect if a manual save was in progress in order to prevent an auto save from happening, and it seems like with these changes you have here this might be a small and worthwhile addition to make

AdamAdHocTeam:
> we just talked on slack about this, so leaving it here as well:
> 
> it would be nice if there was logic to detect if a manual save was in progress in order to prevent an auto save from happening, and it seems like with these changes you have here this might be a small and worthwhile addition to make

This should be fixed now.

nvms (approved):

PR #1126, opened 2022-11-04
[Ttahub 1113] emergent goals not rttapa

GarrettEHill:
## Description of change
set all goals with isFromSmartsheetTtaPlan as not true to isRttapa is false

## How to test
edit a non rttapa goal, the no radio `button should be selected for the is rttapa question.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1113

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20221103000000-emergent-goals-not-rttapa.js
+++ b/src/migrations/20221103000000-emergent-goals-not-rttapa.js
@@ -33,7 +33,7 @@ module.exports = {
         // update isRttapa status on ActivityReportGoals table
         await queryInterface.sequelize.query(
           `UPDATE "ActivityReportGoals" arg
-          SET "isRttapa" = g."isRttapa"::"enum_ActivityReportGoals_isRttapa"
+          SET "isRttapa" = (g."isRttapa"::TEXT)::"enum_ActivityReportGoals_isRttapa"
           FROM "Goals" g
           WHERE arg."goalId" = g.id
           AND arg."isRttapa" IS NULL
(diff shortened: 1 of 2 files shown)


src/migrations/20221103000000-emergent-goals-not-rttapa.js
```diff
@@ -0,0 +1,82 @@
...
+          SELECT "ZAFSetTriggerState"(null, null, null, 'DISABLE');
+          `,
+          { transaction },
+        );
+
+        // remove old column from users table
+        await queryInterface.sequelize.query(
+          `UPDATE "Goals" g
+          SET "isRttapa" = false
+          WHERE COALEASE(g."isFromSmartsheetTtaPlan",false) = false;`,
```
thewatermethod, line 29: isRttapa is an enum of "Yes", "No", not a bool

It's this way as a convenience for the way the UI consumes it - I mean, an enum instead of a boolean because, in the mockups, the field was not indicated as required (or at least that's how I remember it, that could be my mistake), and so the intention seemed to be it could be null, and in that case it makes checking false a little more complicated. 

I'm not sure if the designers were included on this decision (maybe it was their idea, I don't know) but if not, I would confirm that we want to auto set all non-RTTAPA existing goals to emergent and confirm that the field shouldn't be required with either them or Jenn.

GarrettEHill: OK, I will have Jenn confirm before we merge. I have updated to use the enum.

thewatermethod (approved):
Looks good. Any point in updating the ARGoals here? I think not, but just asking.

kryswisnaskas:
Works! 👍  I verified on new activity reports. On an existing report we still an unselected radio button, bu that should be rare.

One change:
Can we drop the enum on the rollback?

kryswisnaskas (approved):

PR #1165, opened 2022-11-15
[TTAHUB-1188] Fix issue where objectives with some data weren't being saved

thewatermethod:
## Description of change

Previously, an objective would need to have a title in order to be saved. With this change, if any part of the form is filled out, an objective will be saved.

## How to test
To reproduce bug on main ar redesign:

- create an AR with one recipient
- use an existing goal with an objective
- select the objective, but don't save
- click "Yes" under file upload
- change your mind and select "Create a new objective"
-  note uploading message appears which is correct
- enter TTa provided and "Save draft"
- note the TTA provided is saved leading someone to believe that they saved their TTA provided.
- click "Save draft" again
- the TTA provided is gone

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1188
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1197

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/goals.js
+++ b/src/services/goals.js
@@ -1281,7 +1281,11 @@ async function createObjectivesForGoal(goal, objectives, report) {
   }
 
   // we don't want to create objectives with blank titles
-  return Promise.all(objectives.filter((o) => o.title).map(async (objective) => {
+  return Promise.all(objectives.filter((o) => o.title
+    || o.ttaProvided
+    || o.topics.length
+    || o.resources
+    || o.files).map(async (objective) => {
     const {
       id,
       isNew,
--- a/src/services/goals.js
+++ b/src/services/goals.js
@@ -1284,8 +1284,8 @@ async function createObjectivesForGoal(goal, objectives, report) {
   return Promise.all(objectives.filter((o) => o.title
     || o.ttaProvided
     || o.topics.length
-    || o.resources
-    || o.files).map(async (objective) => {
+    || o.resources.length
+    || o.files.length).map(async (objective) => {
     const {
       id,
       isNew,
--- a/frontend/src/components/Navigator/index.js
+++ b/frontend/src/components/Navigator/index.js
@@ -311,10 +311,19 @@ function Navigator({
     setValue('goalName', '');
     setValue('goalEndDate', '');
     setValue('goalIsRttapa', '');
+    setValue('goalForEditing.objectives', []);
 
     // the form value is updated but the react state is not
     // so here we go (todo - why are there two sources of truth?)
-    updateFormData({ ...formData, goals: newGoals });
+    updateFormData({
+      ...formData,
+      goals: newGoals,
+      goalForEditing: null,
+      goalName: '',
+      goalEndDate: '',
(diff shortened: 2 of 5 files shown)


kryswisnaskas:
Some data gets saved, but I still run into a couple of scenarios:

Scenario 1
1. create an AR with one recipient
2. use an existing goal with an objective
3. select the objective, but don't save
4. click "Yes" under file upload
5. change your mind and select "Create a new objective"
6. note uploading message appears which is correct
7. enter TTa provided and "Save draft"
8. note the TTA provided is saved leading someone to believe that they saved their TTA provided.
9. click "Save draft" again
10. TTA provided is retained 👍 
11. add topics and an attachmet
12. "Save draft"
13. the objective selection box is empty, so select "Create new objective"
14. observe that topics and attachments get removed; TTA provided is still there
15. click "Save draft"
16. TTA provided will be gone

Scenario 2
steps 1-12 from before
13. Enter objective title 
14. Save goal
15. Read-only box shows
16. Try editing
17. Nothing happens

thewatermethod:
> the objective selection box is empty, so select "Create new objective"

What should the objective selection box say if there is no title? Why would we assume someone would select "create new objective" without intending to create a new objective? I think the problem is the price of doing business if we allow people to save objectives with no title.

I will look into why the TTA provided isn't being cleared out when "create a new objective" is selected

thewatermethod:
@kryswisnaskas I have addressed the issue where the TTA provided is being lost as described above, and the issue where you are unable to edit a goal. If concerns about the question I raised in my last comment remain, and I am understanding them correctly, I think they should involve the UX team first and be handled in a different issue/PR

src/services/goals.js
```diff
@@ -1281,7 +1281,11 @@ async function createObjectivesForGoal(goal, objectives, report) {
   }
 
   // we don't want to create objectives with blank titles
-  return Promise.all(objectives.filter((o) => o.title).map(async (objective) => {
+  return Promise.all(objectives.filter((o) => o.title
+    || o.ttaProvided
+    || o.topics.length
+    || o.resources
+    || o.files).map(async (objective) => {
```
GarrettEHill, line 1288: why are these not checked in the same way as topics?

thewatermethod: good catch, thanks

kryswisnaskas:
> @kryswisnaskas I have addressed the issue where the TTA provided is being lost as described above, and the issue where you are unable to edit a goal. If concerns about the question I raised in my last comment remain, and I am understanding them correctly, I think they should involve the UX team first and be handled in a different issue/PR

Yes, this is an issue for a larger audience. Perhaps we should disable the TTA provided and perhaps other fields if there is no objective's title. Otherwise, we are running into odd cases where toggling the "Select TTA objective" between different objectives, adjusts the topics and resources based on the selection, but the TTA provided stays the same.

Thank you for fixing the Edit issue. I was able to verify it worked! 👍

kryswisnaskas (approved):

PR #1169, opened 2022-11-16
[TTAHUB-1183] Prevent read only view on AR objectives also align all objective fields

AdamAdHocTeam:
## Description of change

This change aims to prevent ever entering a read only view while entering an objective from the AR. It also cleans up and aligns the topics, resources, and files for the various views.

## How to test

1. Create a Goal with and Objective from the RTR
2. On the objective add 3 topics, 3 resources, and 3 files
3. Create a AR using the goal you created
4. Switching the objective status on the AR should NOT go into a read only view
5. Set the objective to 'In Progress' on the AR and remove 1 topic, 1 resource, and 1 file
6. View the goal on the RTR you should see that only the removed items can be removed (leave them)
7. Approve your report and view the Goal from the RTR it should be in the same state as the goal is NOT complete
8. Complete the goal from the main RTR goals and objectives page
9. You should see that its in a read only view and the topics, resources, and objectives you did NOT use have a icon next to them

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1183

### Every PR

- [ ] Meets issue criteria

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/GoalForm/Form.js
+++ b/frontend/src/components/GoalForm/Form.js
@@ -188,7 +188,7 @@ export default function Form({
         />
       ))}
 
-      { (status !== 'Closed' || userCanEdit) && (
+      { (status !== 'Closed') && (
         <div className="margin-top-4">
           {errors[FORM_FIELD_INDEXES.OBJECTIVES_EMPTY]}
           <PlusButton onClick={onAddNewObjectiveClick} text="Add new objective" />
--- a/frontend/src/components/GoalForm/ObjectiveFiles.js
+++ b/frontend/src/components/GoalForm/ObjectiveFiles.js
@@ -14,7 +14,6 @@ export default function ObjectiveFiles({
   files,
   onChangeFiles,
   goalStatus,
-  status,
   isOnReport,
   onUploadFiles,
   index,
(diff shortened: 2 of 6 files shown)


frontend/src/components/GoalForm/Form.js
```diff
@@ -188,7 +188,7 @@ export default function Form({
         />
       ))}
 
-      { (status !== 'Closed' || userCanEdit) && (
```
thewatermethod, line 191: I think this needs to be adjusted so that users without editing privileges don't see the button. (View a goal in a region where you only have read permissions, not write, to reproduce)

Screen Shot 2022-11-17 at 9 01 58 AM

thewatermethod:
This is what I see when the goal and objective are used on a report, but the report is in draft. I don't think that is correct, right?

Screen Shot 2022-11-17 at 9 20 47 AM

frontend/src/components/GoalForm/Form.js
AdamAdHocTeam: This has been fixed to be an && (nice catch).

AdamAdHocTeam:
@thewatermethod I think this is ready for another look.

thewatermethod (approved):
nice looks good

PR #1175, opened 2022-11-17
[TTAHUB-1147] Order AR Objectives

AdamAdHocTeam:
## Description of change

This sets and orders by the new column 'arOrder' column on the ActivityReportObjectives table. 

This change should ensure that objective sorting stays in the order in they were added.

## How to test

For both a **Recipient** and **Other-Entity** reports: Create, Update, and Delete objectives. 

We should maintain the order in which they were added.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1147

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


src/services/createOrUpdateGoalsForActivityReport.test.js
```diff
@@ -0,0 +1,411 @@
...
+        isNew: true,
+      },
+    ];
+
+    await saveObjectivesForReport(
+      objectivesToCreate,
+      otherEntityReport,
+    );
+
+    const createdObjectives = await getObjectivesByReportId(otherEntityReport.id);
```
AdamAdHocTeam, line 234: It seems the function 'createOrUpdateGoalsForActivityReport()' doesn't support OE reports. So I added a CRUD test here with 'saveObjectivesForReport()'.

src/services/reportCache.js
```diff
@@ -95,7 +95,7 @@ const cacheTopics = async (activityReportObjectiveId, topics = []) => {
   ]);
 };
 
-const cacheObjectiveMetadata = async (objective, reportId, metadata) => {
+const cacheObjectiveMetadata = async (objective, reportId, metadata, index) => {
```
thewatermethod, line 98: I think it makes sense to include the index in the metadata parameter instead of adding another (also I would personally prefer if it was called "order")

src/services/goals.js
```diff
@@ -358,19 +358,33 @@ export function reduceObjectives(newObjectives, currentObjectives = []) {
...
       value: id,
       ids: [id],
       // Make sure we pass back a list of recipient ids for subsequent saves.
       recipientIds: [objective.getDataValue('otherEntityId')],
       isNew: false,
+      arOrder,
     }];
   }, currentObjectives);
+
+  const sortedObjectives = objectivesToSort.sort((o1, o2) => {
```
thewatermethod, line 377: ```sort``` sorts the array in place, so it might make sense to just sort and return without saving to a variable

thewatermethod:
This looks pretty good. I was unable to break it - left and came back a few times, edited data, removed some objectives, etc.

PR #1190, opened 2022-11-22
[TTAHUB-943] playwright

nvms:
## Description of change

This PR has:

- a new dependency: playwright
- playwright tests
- a new job in CI for running these tests

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-943

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


tests/account-management.spec.ts
```diff
@@ -0,0 +1,29 @@
+// import { test, expect } from '@playwright/test';
```
AdamAdHocTeam, line 1: Is this still a WIP?

nvms: this test runs fine locally when mailcatcher is also running, but fails in CI probably due to this: https://support.circleci.com/hc/en-us/articles/360007444314-Troubleshooting-sending-email-s-from-a-container

I think there's still value in being able to run this test locally, so will likely gate it behind a `process.env.IS_CI` check or something

tests/activity-report.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+    const url = await page.url();
+    const arNumber = url.split('/').find((part) => /^\d+$/.test(part));
+
+    // create the second goal
+    await page.getByRole('button', { name: 'Add new goal' }).click();
+    await page.getByTestId('textarea').click();
+    await page.getByTestId('textarea').fill('g2');
+    await page.getByRole('group', { name: 'Is this a Recipient TTA Plan Agreement (RTTAPA) goal?*' }).getByText('Yes').click();
+    await page.locator('.css-125guah-control > .css-g1d714-ValueContainer').click();
+    await page.locator('#react-select-25-option-0').click();
```
thewatermethod, line 79: I assume this test works consistently? I have (in the past) struggled to get the same output every time from react-select without using the props to specify ids, etc.

thewatermethod:
Overall this is really great, this will be a huge help

- Can we add instructions about using the ```codegen``` command to create tests? 
- Do we need to update ```.gitignore```? I ran the tests locally and have both ```playwright-report``` and ```test-results``` showing as new files.
- Mail test isn't passing for me locally but I haven't looked into it at all so perhaps it's to do with my particular setup (nothing shows up in mailcatcher)

tests/activity-report.spec.ts
nvms: it passes consistently as long as the styling for the selector stays the same. this particular selector (and others within the test that have similar hashes) is nested within the select component, so changing the padding/margin of the parent element should not change this hash here

nvms:
> Can we add instructions about using the codegen command to create tests?

writing this up right now and will push shortly

> Mail test isn't passing for me locally but I haven't looked into it at all so perhaps it's to do with my particular setup (nothing shows up in mailcatcher)

I think in `.env` you'll want to make sure that `SEND_NOTIFICATIONS` and `SEND_NON_PRODUCTION_NOTIFICATIONS` are both true

tests/activity-report.spec.ts
thewatermethod: Although I am mystified by the reference to margin and padding, I think what you are saying is that it seems to use the same hash every time. I accept! 

I think that it might have been in JestDOM where Emotion was generating different hashes to prefix the CSS each test run (css-_g1d714_ would become css-_125guah_ or whatever)... but I guess if that does happen happens, we'll know to adjust the code here, so it's a moot point

thewatermethod (approved):
I'm looking forward to getting this going ASAP

PR #1197, opened 2022-11-28
[TTAHUB-1127] Prevent showing "Add new objective" on initial select

AdamAdHocTeam:
## Description of change

If the user is creating a new recipient report  and are adding the first objective. They should only be able to do so via the objective select dropdown.

## How to test

Creating a new Recipient report should only allow a selection from the objective drop down. Every subsequent objective should show the "Add new objective" button.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1127

### Every PR

- [x] Meets issue criteria

### Production Deploy

### After merge/deploy


thewatermethod (approved):

thewatermethod:
What might be nice is a test asserting that the button is not there when there are no objectives so that way we have a chance of catching it if we break it

AdamAdHocTeam:
> What might be nice is a test asserting that the button is not there when there are no objectives so that way we have a chance of catching it if we break it

Test has been added.

PR #1209, opened 2022-12-01
[TTAHUB-1254] Validate URLS on the backend and before saving drafts

thewatermethod:
## Description of change
On save draft, bad URLs could be sent to the backend. This change
- Prevents bad urls from being sent
- And if they are, we now make sure to validate them on the backend (no invalid URLs are saved)

## How to test
To duplicate bug - save draft with invalid URL in the resource field of the AR or RTR.
To test fix, enter invalid URL text in an AR or the RTR form. Attempt to save or save draft. Note that your invalid text is not saved.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1254

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/Navigator/index.js
+++ b/frontend/src/components/Navigator/index.js
@@ -187,13 +187,15 @@ function Navigator({
     let invalidResources = false;
     const invalidResourceIndices = [];
 
+    if (objectives) {
     // refire the objective resource validation
-    objectives.forEach((objective, index) => {
-      if (!validateListOfResources(objective.resources)) {
-        invalidResources = true;
-        invalidResourceIndices.push(index);
-      }
-    });
+      objectives.forEach((objective, index) => {
+        if (!validateListOfResources(objective.resources)) {
+          invalidResources = true;
+          invalidResourceIndices.push(index);
+        }
+      });
+    }
 
     if (!isAutoSave && invalidResources) {
       // make an attempt to focus on the first invalid resource
--- a/tests/activity-report.spec.ts
+++ b/tests/activity-report.spec.ts
@@ -48,9 +48,12 @@ test.describe("Activity Report", () => {
     await page.getByLabel('Number of participants involved *').fill('5');
     await page.getByRole('button', { name: 'Save and continue' }).click();
 
+    await page.getByRole('button', { name: 'Supporting attachments not started' }).click(); 
+    await page.getByRole('button', { name: 'Goals and objectives not started' }).click(); 
+
     // create the first goal
     await page.getByTestId('label').locator('div').filter({ hasText: '- Select -' }).nth(2).click();
-    await page.locator('#react-select-15-option-0').getByText('Create new goal').click();
+    await page.locator('#react-select-17-option-0').getByText('Create new goal').click();
     await page.getByTestId('textarea').click();
     await page.getByTestId('textarea').fill('g1');
     await page.getByText('Yes').click();
@@ -60,7 +63,7 @@ test.describe("Activity Report", () => {
     await page.getByLabel('TTA objective *').click();
     await page.getByLabel('TTA objective *').fill('g1o1');
     await page.locator('.css-125guah-control > .css-g1d714-ValueContainer').click();
-    await page.locator('#react-select-21-option-0').click();
+    await page.locator('#react-select-23-option-0').click();
     await blur(page);
 
     // save draft doesn't work with invalid resources
@@ -92,7 +95,7 @@ test.describe("Activity Report", () => {
     await page.getByTestId('textarea').fill('g2');
     await page.getByRole('group', { name: 'Is this a Recipient TTA Plan Agreement (RTTAPA) goal?*' }).getByText('Yes').click();
     await page.locator('.css-125guah-control > .css-g1d714-ValueContainer').click();
-    await page.locator('#react-select-25-option-0').click();
+    await page.locator('#react-select-27-option-0').click();
     await page.getByLabel('TTA objective *').click();
     await page.getByLabel('TTA objective *').fill('g2o1');
     await page.locator('.css-125guah-control > .css-g1d714-ValueContainer').click();


src/lib/urlUtils.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable import/prefer-default-export */
+// regex to match a valid url, it must start with http:// or https://, have at least one dot, and not end with a dot or a space
+const VALID_URL_REGEX = /^https?:\/\/.*\.[^ |^.]/;
+
```
thewatermethod, line 4: sharing code between the front and the backend continues to be a pain point

src/services/createOrUpdateGoalsForActivityReport.test.js
```diff
@@ -353,7 +353,7 @@ describe('createOrUpdateGoalsForActivityReport', () => {
     expect(createdGoals.length).toBe(1);
     expect(createdGoals[0].id).not.toBeNull();
     expect(createdGoals[0].name).toBe('Test create goal for activity reports');
-    expect(createdGoals[0].grantIds).toStrictEqual(grants.map((g) => g.id));
+    expect(createdGoals[0].grantIds.sort()).toStrictEqual(grants.map((g) => g.id).sort());
     expect(createdGoals[0].objectives.length).toBe(3);
```
thewatermethod, line 357: this is just for deflaking reasons, nothing to do with the larger change

src/lib/urlUtils.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable import/prefer-default-export */
+// regex to match a valid url, it must start with http:// or https://, have at least one dot, and not end with a dot or a space
+const VALID_URL_REGEX = /^https?:\/\/.*\.[^ |^.]/;
```
GarrettEHill, line 3: This is the regex i needed to cleanly pull urls out of the legacy data.
```suggestion
const VALID_URL_REGEX = '(?:(?:http|ftp|https|file):\\/\\/)?(?:[\\w%_-]+(?:(?:\\.[\\w%_-]+)+)|(?:\\/[\\w][:]))(?:[\\w\\\\\'\'.,@?^=%&:\\/~+#()-]*[\\w@?^=%&\\/~+#-])';
```

frontend/src/components/GoalForm/constants.js
```diff
@@ -4,7 +4,7 @@ import { v4 as uuidv4 } from 'uuid';
 // regex to match a valid url, it must start with http:// or https://, have at least one dot, and not end with a dot or a space
 const VALID_URL_REGEX = /^https?:\/\/.*\.[^ |^.]/;
```
GarrettEHill, line 5:
```suggestion
const VALID_URL_REGEX = '(?:(?:http|ftp|https|file):\\/\\/)?(?:[\\w%_-]+(?:(?:\\.[\\w%_-]+)+)|(?:\\/[\\w][:]))(?:[\\w\\\\\'\'.,@?^=%&:\\/~+#()-]*[\\w@?^=%&\\/~+#-])';
```

thewatermethod: We don't want to allow file or ftp. This also looks a little complicated compared to what we actually need according to Angela, which is written out in the comment above. Can you explain why we should make this change?

thewatermethod: tried adding this locally, seems to have some syntax errors with the slashes in there so I'm going to hold off committing this suggestion for now

kryswisnaskas (commented):
Works as far as not saving multiple url entries in one field!

Shouldn't we mention that these are going to be discarded?

After all, if not carefully read, our tooltip could be interpreted as inviting users to paste multiple urls in the "Link" (singular) field.
image

nvms:
> ...if not carefully read, our tooltip could be interpreted as inviting users to paste multiple urls in the "Link" (singular) field.

piggybacking on this, I could see something like the following occurring, which passes validation. not sure about the first with the \n and \r, but the second one seems plausible

Screen Shot 2022-12-05 at 8 33 49 AM

thewatermethod:
@nvms 
> to be fair I can't actually reproduce that behavior of pasting from somewhere WITHOUT it sticking spaces between the urls - but just thought I'd point it out. maybe not something to worry about?

Initially, I got everyone to agree that if it was a valid URL in the loosest sense (what passes an input type="url"), then we'd be good to go. However, it's apparent that this is just not good enough in practice, so I've added a check for multiple occurrences of "http", among the other things it needs to check for

@kryswisnaskas 
> Shouldn't we mention that these are going to be discarded?

Kelly, Adam, and I updated the messaging. There are expanded criteria enumerated in the error message, and there is no helper text explicitly asking the users to enter one per line. We don't mention the discarding but hopefully this will help clarify. If there is still confusion/reaction to this, I will ask the UX team to revisit

package.json
```diff
@@ -15,9 +15,9 @@
     "start:ci": "cross-env POSTGRES_USERNAME=postgres POSTGRES_DB=ttasmarthub TTA_SMART_HUB_URI=http://localhost:3000 concurrently \"yarn start:web\" \"yarn client\" \"yarn start:worker\"",
     "start:web": "node ./build/server/index.js",
     "start:worker": "node ./build/server/worker.js",
-    "server": "nodemon src/index.js --exec babel-node --ignore ./tests/ --ignore ./playwright-report/ --ignore ./test-results/",
-    "worker": "nodemon src/worker.js --exec babel-node --ignore ./tests/ --ignore ./playwright-report/ --ignore ./test-results/",
-    "server:debug": "nodemon --inspect=0.0.0.0:9229 src/index.js --exec babel-node --ignore ./tests/ --ignore ./playwright-report/ --ignore ./test-results/",
+    "server": "nodemon src/index.js --exec babel-node --ignore ./tests/ --ignore ./playwright-report/ --ignore ./test-results/ --ignore ./frontend/",
```
AdamAdHocTeam, line 18: Is this supposed to be here?

nvms: whoops! PR to fix this is here: https://github.com/HHS/Head-Start-TTADP/pull/1220

AdamAdHocTeam (commented):
One minor thing I noticed is that if a bad URL (resource) is entered on the RTR. You can 'Save draft' and it will automatically remove the bad resource. However 'Save and continue' prevents the save unlike from the AR. Were the bad resource is removed.

thewatermethod:
> One minor thing I noticed is that if a bad URL (resource) is entered on the RTR. You can 'Save draft' and it will automatically remove the bad resource. However 'Save and continue' prevents the save unlike from the AR. Were the bad resource is removed.

@AdamAdHocTeam -> I have adjusted the behaviors to match

nvms (approved):

kryswisnaskas:
> @AdamAdHocTeam -> I have adjusted the behaviors to match
Would it be hard to extend the same behavior to auto save as well?

thewatermethod:
> Would it be hard to extend the same behavior to auto save as well?

In what way? Prevent auto save?

kryswisnaskas:
> * https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1254



> > Would it be hard to extend the same behavior to auto save as well?
> 
> In what way? Prevent auto save?

Flagging an invalid entry and keeping the values. The concern is that we would wipe out the potentially carefully entered urls on auto save.

thewatermethod:
> Flagging an invalid entry and keeping the values. The concern is that we would wipe out the potentially carefully entered urls on auto save. 

Ah, I see. Right now we do the opposite in all cases (RTR & AR). I am looking into the autosave behavior and will try to come up with a way to preserve the data at least until they navigate away

PR #1225, opened 2022-12-07
[TTAHUB-1254] Fix bug with other entity reports

thewatermethod:
## Description of change
Fixes two bugs with other entity reports, one of which was introduced by #1209, the other I'm not sure. Add an e2e test to validate the fix. 

## How to test
You can see these bugs on main
- First bug: Start an other entity report. On the activity summary, select only the other entities, then proceed to cycle through the links in the activity report side navigation. You should find yourself stuck by a syntax error.
- Second bug: Complete an other entity report. You'll find that you cannot get the "goals and objectives" side nav label to show complete if you proceed through the report in order (it shows complete when you save and continue onto "supporting attachments," but if you proceed to next steps it reverts to "Not started")

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1254

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/pages/ActivityReport/index.js
+++ b/frontend/src/pages/ActivityReport/index.js
@@ -178,36 +178,6 @@ export const unflattenResourcesUsed = (array) => {
   return array.map((value) => ({ value }));
 };
 
-/**
- * Goals created are editable until the report is loaded again. The report used to
- * not update freshly created goals with their DB id once saved, but this caused
- * any additional updates to create a brand new goal instead of updating the old goal.
- * We now use the goal created in the DB. However this means we no longer know if the
- * goal should be editable or not, since it was loaded from the DB. This method takes
- * the list of newly created goals and grabs their names, placed in the `editableGoals`
- * variable. Then all goals returned form the API (the report object passed into this
- * method) have their name compared against the list of fresh goals. The UI then uses
- * the `new` property to determine if a goal should be editable or not.
- * @param {*} report the freshly updated report
- * @returns {function} function that can be used by `setState` to update
- * formData
- */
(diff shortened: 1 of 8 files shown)


frontend/src/pages/RecipientRecord/pages/GoalsObjectives.js
```diff
@@ -27,10 +27,8 @@ export default function GoalsObjectives({
   const onRemoveFilter = (id) => {
     const newFilters = [...filters];
     const index = newFilters.findIndex((item) => item.id === id);
-    if (index !== -1) {
```
thewatermethod, line 30: these were removed to raise branch coverage; it may look a little weird, but when I set about writing a test for it, there was no way to do it through the UI, so not sure it makes sense to check here

frontend/src/pages/ActivityReport/index.js
```diff
@@ -178,36 +178,6 @@ export const unflattenResourcesUsed = (array) => {
   return array.map((value) => ({ value }));
 };
 
-/**
```
thewatermethod, line 181: removed as we haven't used this since the switchover, that way we don't have to test this

frontend/src/pages/RecipientRecord/pages/GoalsObjectives.js
AdamAdHocTeam, line 30: Was this check hurting anything?

AdamAdHocTeam (approved):
Looks good.

- I was able to create an OE report and switch pages via the nav.
- I was able to get the objectives step to 'complete'.

PR #1228, opened 2022-12-08
playwright improvements: ability to advance timers

nvms:
## Description of change

It would be nice to be able to write e2e tests for things that depend on the order of certain time-based operations, like the autosave.

Since playwright doesn't support advancing timers (yet), this PR adds `sinon` as a devDependency and a helper in `tests/common.ts` that injects a `__clock` method into the `window` of our tests so that we can do this:

```typescript
import { test, expect } from '@playwright/test';
import { useClock } from './common';

useClock(test);

test('advance time', async ({ page }) => {
  await page.evaluate(() => (window as any).__clock.tick(120_000));
});
```

## Issue(s)

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,8 @@
     "worker": "nodemon src/worker.js --exec babel-node",
     "server:debug": "nodemon --inspect=0.0.0.0:9229 src/index.js --exec babel-node",
     "client": "yarn --cwd frontend start",
-    "test": "jest src --runInBand", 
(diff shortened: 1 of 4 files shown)


thewatermethod (approved):
Let's give it a try! But also we should try to keep an eye on that issue so we can remove the dependency at some point

PR #1239, opened 2022-12-15
Update import goals script

kryswisnaskas:
## Description of change

Updates the goal import script to reflect the schema changes. Topics and Roles are no longer looked at since they belong to individual objectives. The status of existing goals is updated, but only if the status from RTTAPA is more recent.

## How to test
Unit tests pass

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1045

### Every PR

- [x] Meets issue criteria
- [n/a] UI review complete

### Production Deploy

- [x] Staging smoke test completed

### After merge/deploy


src/tools/importPlanGoals.js
```diff
@@ -135,26 +145,53 @@ export default async function importGoals(fileKey, region) {
             const dbGrant = await Grant.findOne({ where: { ...fullGrant }, attributes: ['id', 'recipientId'] });
             if (!dbGrant) {
               // eslint-disable-next-line no-console
-              console.log(`Couldn't find grant: ${fullGrant.number}. Exiting...`);
+              logger.error(`Couldn't find grant: ${fullGrant.number}. Exiting...`);
```
GarrettEHill, line 148: Do many of these occur? Does this indicate insufficiencies in the import of HSES? Are we missing a type of grant?

src/tools/importPlanGoals.js
```diff
@@ -135,26 +145,53 @@ export default async function importGoals(fileKey, region) {
...
+                  timeframe: goal.timeframe,
+                  isRttapa: 'Yes',
+                },
+                {
+                  // where: { id: dbgoal.id },
+                  individualHooks: true,
+                },
+              );
+              // determine if status needs to be updated
+              await updateStatus(goal, dbGoal);
```
GarrettEHill, line 180: Instead of running two updates, could the status be conditionally set? This way we don't unnecessarily fill the audit log and protentional make undesired updates to the timeframe and isRttapa. As if the status is ignored, should the other updates maybe be ignored too.

kryswisnaskas: I see the audit log consideration, but the status updates should not be extensive, so the audit log will only get minimal updates. We also want to bring timeframe for the data consistency. Ensuring isRttapa is set in the same update will go along with what we discussed previously and agreed that all goals from RTTAPA will be marked so in the hub.

PatricePascual-ACF (approved):

PR #1242, opened 2022-12-16
[TTAHUB-1329] Format enddate on backend

thewatermethod:
## Description of change
This change formats the goal end date on the backend. This comes out of a user reported bug where a goal end date saved on the activity report would immediately disappear on editing it.

## How to test
Confirm that you can save a goal's end date, edit it, leave and return and do the same on the activity report and the RTR create goals page.

To repro existing bug on main:
1) on a new AR, add an existing goal without a date. 
2) Add a date
3) Save draft
4) Leave report (go to landing page)
5) Return to report
6) Date should be gone

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1329

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/services/goals.js
```diff
@@ -545,6 +545,7 @@ function reduceGoals(goals, forReport = false) {
         currentValue.objectives,
       ),
       isNew: false,
+      endDate: currentValue.endDate,
```
thewatermethod, line 548: @AdamAdHocTeam 

[From sequelize docs](https://sequelize.org/api/v6/class/src/model.js~model):

> ...if getters and/or setters are defined for field they will be invoked, instead of returning the value from dataValues

Before we were grabbing endDate as part of ```...currentValue.dataValues``` in this reducer, which meant the custom getter wasn't being called. Turns out all we have to do is access it as above and that get function I wrote on the model is invoked. Good call!

AdamAdHocTeam (approved):
Nice.

GarrettEHill (approved):

PR #1248, opened 2022-12-20
[TTAHUB-1288] Ensure required fields are identified as such by a screenreader

thewatermethod:
## Description of change
Handled the various cases where a field was required and wasn't announced as such by the screenreader. ```react-select``` needed to be updated to access a required attribute added in a more recent version. Other fields were handled on a case by case basis. 

Also added some focus management for a keyboard user adding a new objective on the RTR

## How to test
Access the site via a screenreader (I used VoiceOver on Mac)
Make sure all required fields in the AR & RTR are announced as required

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1288

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -46,7 +46,7 @@
     "react-router-dom": "^5.2.0",
     "react-router-hash-link": "^2.3.1",
     "react-router-prop-types": "^1.0.5",
-    "react-select": "^4.1.0",
+    "react-select": "^5.7.0",
     "react-stickynode": "^4.0.0",
     "react-with-direction": "^1.3.1",
     "sass": "^1.51.0",
--- a/frontend/src/components/AutomaticResizingTextarea.js
+++ b/frontend/src/components/AutomaticResizingTextarea.js
@@ -7,7 +7,7 @@ import {
 const DEFAULT_TEXTAREA_HEIGHT = 160;
 
 export default function AutomaticResizingTextarea({
-  onUpdateText, onBlur, inputName, disabled, value,
+  onUpdateText, onBlur, inputName, disabled, value, required,
 }) {
   const [height, setHeight] = useState(`${DEFAULT_TEXTAREA_HEIGHT}px`);
   const onChange = (e) => {
@@ -24,7 +24,7 @@ export default function AutomaticResizingTextarea({
       name={inputName}
       value={value}
       onChange={onChange}
-      required
+      required={required}
       disabled={disabled}
       style={{ height }}
     />
@@ -37,8 +37,10 @@ AutomaticResizingTextarea.propTypes = {
   inputName: PropTypes.string.isRequired,
   disabled: PropTypes.bool,
   value: PropTypes.string.isRequired,
+  required: PropTypes.bool,
 };
 
 AutomaticResizingTextarea.defaultProps = {
   disabled: false,
+  required: true,
 };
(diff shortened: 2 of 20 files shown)


frontend/package.json
```diff
@@ -52,7 +52,7 @@
     "react-router-dom": "^5.2.0",
     "react-router-hash-link": "^2.3.1",
     "react-router-prop-types": "^1.0.5",
-    "react-select": "^4.1.0",
+    "react-select": "^5.7.0",
     "react-stickynode": "^4.0.0",
```
thewatermethod, line 56: we can make this update given we aren't supporting IE any longer

AdamAdHocTeam (commented):
Some things I found testing with Microsoft Narrator:

- On AR summary page "What type of TTA was provided?" doesn't say required but it has a asterisk.
- On goal creation "Is this a Recipient TTA Plan Agreement?" doesn't say required by has asterisk.
- On goal creation Objective Topics doesn't say required.
- When tabbing between Recipient and Specialist Next steps we should announce what we are currently focused on (Recipient vs Specialist step).

thewatermethod:
>  On AR summary page "What type of TTA was provided?" doesn't say required but it has a asterisk.
    On goal creation "Is this a Recipient TTA Plan Agreement?" doesn't say required by has asterisk.
    On goal creation Objective Topics doesn't say required.
    When tabbing between Recipient and Specialist Next steps we should announce what we are currently focused on (Recipient vs Specialist step).

Think this is ready for another look.

I was largely unable to recreate these but I made some changes (explicitly marking certain inputs with an attribute). The recipient/specialist inputs have different labels that function as announcements (enter recipient's next step, etc). I don't have Microsoft Narrator at hand to check on this MacBook but if this round of changes doesn't solve it I can perhaps set my GFE up to test with that.

Also, on the Objective Topics, just checking to make sure that you installed the updated react-select before testing, as the announcement depends on using >v5.

AdamAdHocTeam (commented):
I did another pass of testing some things seem fixed while others still and issue (use chrome screen reader extension):

- Still no prompt for "What type of TTA was provided?" maybe do the same thing as duration.
- On goal "Is this a Recipient TTA Plan Agreement?" only reads when I directly click on it (not tab).
- Objective topics seems fixed.
- Steps seem fixed.

thewatermethod:
Of the screenreaders you mentioned, I don't have easy access to NVDA or Microsoft Narrator (Windows only). I did test with ChromeVox and confirmed the following:

When the fieldset is focused, **required is announced** on "What type of TTA was provided?". It's difficult to get this focus behavior with the keyboard/ChromeVox combo, but it is basically the same problem you describe with the RTTAPA question, clicking in the field to get the readout.

Screen Shot 2023-01-19 at 9 31 15 AM

ChromeVox doesn't read off the fieldset/legend on the radio button or checkbox groups the way that VoiceOver does, but I did confirm that we have marked it up in [the manner recommended by the W3C](https://www.w3.org/WAI/tutorials/forms/grouping/). The way it is advertised to work (and does on VoiceOver) is [the way described in this article](https://accessibility.blog.gov.uk/2016/07/22/using-the-fieldset-and-legend-elements/).

Regardless, I am not overly concerned with either ChromeVox or Microsoft Narrator. They have comparatively tiny shares of the screen reader market, presumably at least partially because they don't always work as expected. NVDA is another matter. I think the way forward is for me to write another ticket to test this functionality exhaustively with NVDA and to do that, I have to get a Windows environment up and running.

In the meantime, I'd like to get these changes merged in (updates to a vendor library, improvements to the semantics of the HTML).

GarrettEHill:
Can this be resynced with main to allow us to review it and get it pushed out.

PR #1251, opened 2022-12-21
[TTAHUB-1324] User impersonation

nvms:
### how it works

instead of reaching for `req.session.userId`, the backend should get the `userId` by calling `currentUserId(req, res)`. previously, `currentUserId` would just check `req.session` and `req.locals` for a `userId`. if it existed, it would be returned. now, it first checks if `req.headers['auth-impersonation-id']` exists. if it does, and the user defined by `req.session.userId` (or `req.locals.userId`) is an admin, then impersonation is allowed and `currentUserId` returns the user id in the `auth-impersonation-id` header.

since `req.session.userId` and `req.locals.userId` is never touched, we still know who the impersonator is, and we can verify their administrative permissions with every request.

on the frontend side of things, `get`, `post`, `put` and `delete` automatically apply this header to requests if it sees an `auth-impersonation-id` key in localStorage. that's it. to impersonate somebody, we set this key. to stop impersonating, we destroy this key.

### things we can do with this

- write playwright tests for multi-user scenario behaviors (websocket things, approvals, other permission-based views)
- develop locally without ever having to modify `CURRENT_USER_ID` in `.env`
- better support our customers by seeing the exact UI that they see, although there are probably more discussions that need to be had about how best to do this, and what type of audit logging we will need in order to ensure we're fully compliant with any privacy-related issues. until this is decided, we can just disable this in production and still benefit from the developer experience of this change

### things to still figure out before this is ready for production

- audit logging issues: if I'm impersonating a user and I modify a report, it will appear in the logs as though the user I am impersonating made this change, which is not exactly true. it seems like maybe we should be able to differentiate a "normal" session action from an "impersonated" session action, while tracking who was doing the impersonating.
- should all admins be able to impersonate any user? probably not. maybe an additional permission will need to be introduced?

for now this is disabled in prod:

on the frontend:
https://github.com/HHS/Head-Start-TTADP/blob/271f8215183851efbe15042f99f3df9b10faf1ab/frontend/src/pages/Admin/UserSection.js#L117-L119

and the backend:
https://github.com/HHS/Head-Start-TTADP/blob/271f8215183851efbe15042f99f3df9b10faf1ab/src/services/currentUser.js#L43-L45

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1324

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


frontend/src/components/HeaderUserMenu.js
```diff
@@ -33,8 +34,17 @@ UserMenuNav.propTypes = {
 };
 
 function HeaderUserMenu() {
+  let hasImpersonationKey;
+  try {
+    hasImpersonationKey = window.localStorage.getItem(LOCAL_STORAGE_IMPERSONATION_KEY) !== null;
```
thewatermethod, line 39: this is probably fine, but we do have a ```storageAvailable``` function in ```/hooks/helpers``` as well (just as an FYI)

Is it possible that session storage is a better choice for this? Impersonation for the life of a browser tab, that way you could be accessing the site as two users at once? Not sure how we'll use this 100% so not sure how to properly weigh costs/benefits to both

thewatermethod: the other weird corner case is thing is for users, only people on our team as far as I can think of, who use the site in multiple environments, I think session storage is better in that scenario as well

src/routes/users/handlers.js
```diff
@@ -37,8 +38,8 @@ export async function getPossibleStateCodes(req, res) {
 
 export async function requestVerificationEmail(req, res) {
```
thewatermethod, line 39: This seems like another thing we have to think about before pushing this to prod - are there certain things a user impersonator _shouldn't_ be able to do, like request an email verification?

thewatermethod:
This is very cool, and I think it would be pretty helpful. Why not bring up enabling this in production with the group at a standup (unless you did, and I missed it)? I totally support erring on the side of caution, but I'd like to know if there's actually any reason not to enable this, and if there is, what we can do to mitigate or overcome those issues.

frontend/src/components/HeaderUserMenu.js
nvms: this is a great idea! I made this change here: bc9cae8

src/routes/users/handlers.js
nvms: yeah, I raised this concern and a few others at yesterday's standup. I think that along with this specific case, there may also be other things we don't want to allow impersonators to do. I don't have a well-defined list of what those things are yet though, which probably means there are more group conversations to be had

my current view on this is: we allow basically anything if `NODE_ENV` is `development` (because having these powers is very useful to us while building stuff), otherwise we disallow things like this

nvms:
> This is very cool, and I think it would be pretty helpful. Why not bring up enabling this in production with the group at a standup (unless you did, and I missed it)? I totally support erring on the side of caution, but I'd like to know if there's actually any reason not to enable this, and if there is, what we can do to mitigate or overcome those issues.

thanks! I did bring it up yesterday but Patrice and a few others who might be interested in this feature weren't able to make it. I think I'll probably bring it up on Monday as a discussion point

what I'm thinking is: if this looks okay in its current state, then we get it in and when the group decides what limitations should be in place for prod, I can open a PR with those changes. or maybe we just hold out for a decision and make those changes to this PR? I could go either way on this. what do you think?

thewatermethod:
> what I'm thinking is: if this looks okay in its current state, then we get it in and when the group decides what limitations should be in place for prod, I can open a PR with those changes. or maybe we just hold out for a decision and make those changes to this PR? I could go either way on this. what do you think?

I think it totally makes sense to break this into two if that is the way to keep things moving forward. I just want to have a definite plan or understanding of what the next steps are (do we need to have the "support engineer" permission, for example?), and I also would like to double check that there are actual security issues with this.

thewatermethod:
So where did we leave this at standup? Conversation went so fast I got lost. Is this ready to move forward as is?

thewatermethod:
(I am ready to approve but want to make sure its all done being worked on)

nvms:
> (I am ready to approve but want to make sure its all done being worked on)

there are no more changes needed for this PR IMO, at standup we confirmed there are only two requirements for this to be enabled in production:

- [x] admins should not be allowed to impersonate other admins
- [ ] the audit log needs to track actions of impersonated users

I'm planning a follow-up PR to handle the audit log stuff as well as removing this bits that prevent this from working in production

thewatermethod (approved):

PR #1253, opened 2022-12-21
[TTAHUB-1300] Fix display of grants for multi grant goals on the RTR

thewatermethod:
## Description of change
This PR fixes a bug where the wrong API field was accessed when editing or viewing an existing goal with multiple grants on the RTR.( All grants are returned but only one was populating the select or read-only field)

## How to test
Create a multi grant goal on the RTR. Save, submit, and edit. You will see that the proper number of grants is displayed throughout. Leave the RTR. Return to that goal or another multi-grant goal. You will still see the proper number of grants, whether the goal is editable or read-only.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1300

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


kryswisnaskas:
👍 Verified that multiple grants are shown when editing for multi-grant goals!

Could the trailing dashes and spaces be removed to make it look like the grant numbers listed under objectives? That way the UI will be more consistent and the Playwright tests will appreciated it as well.
(this could be a separate ticket/PR if it's more complicated)

kryswisnaskas:
I was expecting to see something similar to what there is on the objective:
image

Which also highlights a need to add 's' to the 'Grant number'.

I looked at the production data and it appears that we don't have any active grants without the corresponding program type data, so it sounds like we might be good there (at least for now). In this case it is a matter of adjusting our test seeders as it seems like the current tests are not catching the "empty" program types (below). I can update the seeders, but as a side note, shouldn't
 `    numberWithProgramTypes: {
      type: DataTypes.VIRTUAL,`

already remove the dashes?
image

thewatermethod:
@kryswisnaskas I switched to using the derived "numberWithProgramTypes", which should correctly display any grants without program types, added an "s" to the Grant number row in the Objective Card, and added programs to the seeders for grants. I ran through the common use cases and don't see any weirdness.

src/seeders/20201211172017-grants.js
```diff
@@ -85,12 +87,28 @@ const grants = [
...
+  status: 'Active',
+  endDate: '2032-09-01',
+};
+
+const programs = grants.map((grant) => ({
+  ...baseGrant,
+  name: faker.company.companyName(),
+  id: grant.id,
+  grantId: grant.id,
+}));
```
kryswisnaskas, line 102: Should we also add the program types since that will be more aligned with what is in prod?

kryswisnaskas (approved):
Looks good! 👍 (only one comment on the program types)

frontend/src/components/GoalForm/index.js
```diff
@@ -146,7 +143,7 @@ export default function GoalForm({
         setDatePickerKey(goal.endDate ? `DPK-${goal.endDate}` : '00');
         setIsRttapa(goal.isRttapa);
         initialRttapa.current = goal.isRttapa;
-        setSelectedGrants(formatGrantsFromApi([goal.grant]));
+        setSelectedGrants(formatGrantsFromApi(goal.grants ? goal.grants : [goal.grant]));
```
AdamAdHocTeam, line 146: This fixes the issue of only one objective being updated also.

AdamAdHocTeam (approved):
Works.

PR #1255, opened 2022-12-22
[TTAHUB-1311] fix resources not clearing when switching between editing goals

thewatermethod:
## Description of change

There was a bug that essentially boiled down to: on the AR, if you opened a goal for editing, then opened a different goal for editing _without_  saving the goal you initially opened, the resource text field would not be cleared. The Jira lists steps to reproduce the bug. I found that writing a test in Jest would be kind of gnarly given, so I opted to add a Playwright test instead.

## How to test
Create two goals on an activity report. The first should have an objective with a URL resource, the second should have none. Save them both so that they are in read-only mode. Edit the first via the ellipsis menu. The resource should appear, as expected. Edit the second without saving the first goal. The resource should disappear, which makes sense, since the second goal had no objectives with a resource.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1311

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


kryswisnaskas (approved):
Looks good! 👍

frontend/src/pages/ActivityReport/Pages/goalsObjectives.js
```diff
@@ -186,14 +188,11 @@ const GoalsObjectives = ({
   };
 
   // the read only component expects things a little differently
-  const goalsForReview = selectedGoals.map((goal, index) => {
-    const fieldArrayName = `goals[${index}].objectives`;
-    const objectives = getValues(fieldArrayName) || [];
+  const goalsForReview = selectedGoals.map((goal) => {
```
AdamAdHocTeam, line 191: Why were we returning the objectives here before?

PR #1264, opened 2023-01-03
[TTAHUB-1317] Only find matching objectives based on trim

AdamAdHocTeam:
## Description of change

This solves a bug that I introduced as part of the fix for [TTAHUB-1240](https://github.com/HHS/Head-Start-TTADP/pull/1204/files).

In attempt to prevent showing duplicate objectives with leading or trailing spaces we perform a trim on the returned objective title. We cannot do this as it flags the title as changed in the Sequelize instance. This causes a validation error to be thrown if the goal is on an approved report (user case).

For now we will see if a non-trailing version of the objective title exists. Else we will return the title as is with spaces.

## How to test

You can edit recipient 'Blueprints' for G-20778, G-12658 and attempt to change all statuses to complete and save.

OR

Create a goal with a few objectives on an approved report. Make sure one of the Objective titles has trailing spaces. You should still be able to edit the goal and make certain changes to the Objective and save without error.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1317

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/models/hooks/activityReport.js
+++ b/src/models/hooks/activityReport.js
@@ -230,7 +230,7 @@ const determineObjectiveStatus = async (activityReportId, sequelize, isUnlocked)
       // Get latest report by end date.
       const latestAR = relevantARs.reduce((r, a) => {
         if (r && r.endDate) {
-          return r.endDate > a.endDate ? r : a;
+          return new Date(r.endDate) > new Date(a.endDate) ? r : a;
         }
         return a;
       }, {
--- a/src/models/tests/objectiveStatus.test.js
+++ b/src/models/tests/objectiveStatus.test.js
@@ -169,12 +169,12 @@ describe('Objective status update hook', () => {
       grantFour = await Grant.findOne({ where: { id: 476470 } });
 
       // Reports.
-      reportOne = await ActivityReport.create({ ...sampleReport, endDate: '2022-09-30T12:00:00Z' });
+      reportOne = await ActivityReport.create({ ...sampleReport, endDate: '2023-01-01T12:00:00Z' });
       reportTwo = await ActivityReport.create({
         ...sampleReport,
         submissionStatus: REPORT_STATUSES.SUBMITTED,
         calculatedStatus: REPORT_STATUSES.APPROVED,
-        endDate: '2022-09-29T12:00:00Z',
+        endDate: '2022-09-30T12:00:00Z',
       });
       reportThree = await ActivityReport.create({ ...sampleReport });
 
--- a/src/services/goals.js
+++ b/src/services/goals.js
@@ -363,7 +363,7 @@ export function reduceObjectives(newObjectives, currentObjectives = []) {
 
     return [...objectives, {
       ...objective.dataValues,
-      title: objective.title.trim(),
+      title: objective.title,
       value: id,
       ids: [id],
       // Make sure we pass back a list of recipient ids for subsequent saves.
@@ -448,7 +448,7 @@ export function reduceObjectivesForActivityReport(newObjectives, currentObjectiv
 
     return [...objectives, {
       ...objective.dataValues,
-      title: objective.title.trim(),
+      title: objective.title,
       value: id,
       ids: [id],
       ttaProvided,
--- a/src/services/goals.js
+++ b/src/services/goals.js
@@ -344,7 +344,7 @@ export function reduceObjectives(newObjectives, currentObjectives = []) {
   // we pass in the existing objectives as the accumulator
   const objectivesToSort = newObjectives.reduce((objectives, objective) => {
     const exists = objectives.find((o) => (
-      o.title === objective.title.trim() && o.status === objective.status
+      o.title.trim() === objective.title.trim() && o.status === objective.status
     ));
 
     if (exists) {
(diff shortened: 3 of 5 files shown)


thewatermethod:
With this change, objectives are not de-duplicated on the "edit goals" page (when you click "edit" from the RTR) or in the objective select dropdown in the Activity Report. I confirmed that they were correctly de-duplicated on main

Example: /recipient-tta-records/706/region/1/goals?id[]=27525,27528

AdamAdHocTeam:
> 27528

The issue here is that both objectives titles have trailing spaces. So when we attempt to find the existing objective, we need to trim both titles.

This should be ready for another look.

thewatermethod:
Found a couple of bugs. I know your code isn't responsible for introducing them but I wonder if they should be fixed before we merge this fix in since it will open us up to putting data in a weird state.

Editing the same objectives I linked to before
- updating the status from the RTR only updates _one_ of the objective statuses
- updating the status from the AR does not seem to update either objective at all, only the ActivityReportObjective

thewatermethod:
Regardless, up to you - if you want me to approve this to unblock a user, maybe you could write up some jira tickets to document these bugs.

AdamAdHocTeam:
> Found a couple of bugs. I know your code isn't responsible for introducing them but I wonder if they should be fixed before we merge this fix in since it will open us up to putting data in a weird state.
> 
> Editing the same objectives I linked to before
> 
> * updating the status from the RTR only updates _one_ of the objective statuses
> * updating the status from the AR does not seem to update either objective at all, only the ActivityReportObjective (I added the objective to an AR, changed the status, and approved the report)

Lets go over these once your back. While testing I'm getting some mixed results (maybe I'm using the wrong data).

src/models/hooks/activityReport.js
```diff
@@ -230,7 +230,7 @@ const determineObjectiveStatus = async (activityReportId, sequelize, isUnlocked)
       // Get latest report by end date.
       const latestAR = relevantARs.reduce((r, a) => {
         if (r && r.endDate) {
-          return r.endDate > a.endDate ? r : a;
+          return new Date(r.endDate) > new Date(a.endDate) ? r : a;
```
GarrettEHill, line 233: is this happening anywhere else?

GarrettEHill: Looks like maybe a few:
./src/services/recipient.js:204:        endDate: report.endDate > a.endDate ? report.endDate : a.endDate,
./src/services/recipient.js:253:    a.endDate === b.endDate ? a.id < b.id
./src/services/recipient.js:254:      : a.endDate < b.endDate) ? 1 : -1));
./src/widgets/totalHrsAndRecipientGraph.js:129:      if (!arDates.find((cache) => cache.id === r.id && cache.date === r.startDate)) {

AdamAdHocTeam: This was the test I ran.

// Check === with Dates
const dateTestE = '01/05/2023';
const dateELess = '01/05/2023';
const checkE = new Date(dateTestE) === new Date(dateELess);
console.log('ResultE: ', checkE);

// Check === with Strings
const dateTestF = '01/05/2023';
const dateFLess = '01/05/2023';
const checkF = dateTestF === dateFLess;
console.log('ResultF: ', checkF);

AdamAdHocTeam:
> > Found a couple of bugs. I know your code isn't responsible for introducing them but I wonder if they should be fixed before we merge this fix in since it will open us up to putting data in a weird state.
> > Editing the same objectives I linked to before
> > 
> > * updating the status from the RTR only updates _one_ of the objective statuses
> > * updating the status from the AR does not seem to update either objective at all, only the ActivityReportObjective (I added the objective to an AR, changed the status, and approved the report)
> 
> Lets go over these once your back. While testing I'm getting some mixed results (maybe I'm using the wrong data).

@thewatermethod I have fixed the two additional issues you mentioned in your comments.
1. The first issue will be fixed in ttahub-1320-update-all-grants.
2. The second issue is included in this PR (approved reports).

thewatermethod:
@AdamAdHocTeam Is there any way we could change our tests to catch some of the cases we've turned up during the review? Specifically, can we assert via a test somewhere:

- that it's possible to change the status of a "rolled-up" objective (the original bug) 
- that only one objective will be returned by the backend even if one has a trailing space and one doesn't
- that approving an AR will change an objective's status, even if that objective is on another AR

AdamAdHocTeam:
> * that it's possible to change the status of a "rolled-up" objective (the original bug)

"that it's possible to change the status of a "rolled-up" objective (the original bug)"

This bug is fixed by another PR.

AdamAdHocTeam:
> @AdamAdHocTeam Is there any way we could change our tests to catch some of the cases we've turned up during the review? Specifically, can we assert via a test somewhere:
> 
> * that it's possible to change the status of a "rolled-up" objective (the original bug)
> * that only one objective will be returned by the backend even if one has a trailing space and one doesn't
> * that approving an AR will change an objective's status, even if that objective is on another AR

@thewatermethod add requested test changes.

thewatermethod (approved):
Nice work, thanks for adding those tests

PR #1279, opened 2023-01-10
[TTAHUB-1182] Change "Is this RTTAPA" to "Goal type"

nvms:
## Description of change

|before label|now label|
|---|---|
|`Is this a Recipient TTA Plan Agreement (RTTAPA) goal?`|`Goal type`|
|`Yes`|`RTTAPA`|
|`No`|`Non-RTTAPA`|

Screen Shot 2023-02-01 at 10 00 45 AM

clicking `View goal type guidance` opens a drawer with some help docs:

Screen Shot 2023-02-01 at 10 00 53 AM

## Issue(s)

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


frontend/src/components/Drawer.js
```diff
@@ -0,0 +1,133 @@
...
+  stickyHeader,
+  stickyFooter,
+  children,
+  triggerRef,
+}) {
+  const [isOpen, setIsOpen] = useState(false);
+  const elementRef = useRef(null);
+  const closeButtonRef = useRef(null);
+
+  const headerHeight = useMemo(() => {
```
AdamAdHocTeam, line 26: Any big diff between **memo** with empty dependency array vs **useEffect** with an empty array?

frontend/src/components/GoalForm/GoalRttapa.js
```diff
@@ -1,7 +1,8 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useRef } from 'react';
 import PropTypes from 'prop-types';
 import { FormGroup, Fieldset, Radio } from '@trussworks/react-uswds';
 import Req from '../Req';
+import Drawer from '../Drawer';
```
AdamAdHocTeam, line 5: I almost wonder if we want to move this up to site NAV level so we can reuse it for messages on other pages.

AdamAdHocTeam: Just pass in the content it displays as a child.

AdamAdHocTeam (commented):
I have an AR with Non-RTTAPA select but the read only shows RTTAPA.

image

Also once a goal is approved do we want to show the type of goal when viewing the report.

image

frontend/src/components/Drawer.js
nvms: with `useMemo` I don't need to also manager header height state in some `useState` state container. instead, I can just use the return value, simplifying things a bit.

frontend/src/components/GoalForm/GoalRttapa.js
nvms: right now it's at `src/components/Drawer.js`. I'm fine with moving it somewhere else - where were you thinking?

AdamAdHocTeam (approved):
Looks good!

thewatermethod:
Everything looks good! Couple of thoughts, nothing blocking this PR IMO:
- Do we want to use a subtle slide-in animation for the drawer?
- Most or at least some of the CSS, I would think, could be converted to USWDS utility-classes. If the mockup looks different, like for the box-shadow, I think we should still defer [to the closest USWDS implementation](https://designsystem.digital.gov/utilities/shadow/)
- When I click on the drawer it triggers the form validation for the RTTAPA question, probably better if it didn't.

kryswisnaskas (approved):
Nice! Looks good! 👍

nvms:
> Everything looks good! Couple of thoughts, nothing blocking this PR IMO:
> 
> * Do we want to use a subtle slide-in animation for the drawer?
> * Most or at least some of the CSS, I would think, could be converted to USWDS utility-classes. If the mockup looks different, like for the box-shadow, I think we should still defer [to the closest USWDS implementation](https://designsystem.digital.gov/utilities/shadow/)
> * When I click on the drawer it triggers the form validation for the RTTAPA question, probably better if it didn't.

thanks - I made all of these changes

- [x] don't validate on blur
- [x] use utility classes as much as possible
- [x] subtle slide-in animation (video demo below)

https://user-images.githubusercontent.com/17074357/216650802-0d6d5cb7-03ec-4d93-82ab-ffeeebdc6d7d.mov

thewatermethod (approved):
I actually didn't mean for you to disable on blur validation BUT thinking about it, I think that's fine in this case because its required and doesn't let you save without a value, and you'd have to click a radio button (therefore passing validation) to get a blur anyways.

thewatermethod:
also the animation is cool

PR #1280, opened 2023-01-10
[TTAHUB-1258] Fix potential flaky tests

nvms:
## Description of change

looks like this may be the cause for much of the flakiness:

https://github.com/HHS/Head-Start-TTADP/blob/df00e8d01ebe563a7f8440186debde2ddfd207e1/src/testUtils.js#L81

this test util is called quite often, and given that there are only 13 possible return values for `animal.type()`, we're not providing enough variation for this to be reliably used as the value for a unique column

## How to test

run a test suite locally which makes heavy use of the above test util, such as `src/scopes/activityReport/index.test.js`. you should not get any validation errors

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1258

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


thewatermethod (approved):
🔥

PR #1295, opened 2023-01-19
Await each bulk before starting a new one

AdamAdHocTeam:
## Description of change

It seems that the issue we are having might have to do with overloading the ES queue.

This attempts to fix it by awaiting each batch call in the loop.

```
Write rejections

A 429 error message as a write rejection indicates a bulk queue error. The es_rejected_execution_exception[bulk] indicates that your queue is full and that any new requests are rejected. This bulk queue error occurs when the number of requests to the cluster exceeds the bulk queue size (threadpool.bulk.queue_size). A bulk queue on each node can hold between 50 and 200 requests, depending on which Elasticsearch version you are using.
```

https://aws.amazon.com/premiumsupport/knowledge-center/opensearch-resolve-429-error/

## Issue(s)

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/tools/createAwsElasticSearchIndexes.js
+++ b/src/tools/createAwsElasticSearchIndexes.js
@@ -136,22 +136,19 @@ export default async function createAwsElasticSearchIndexes(batchSize = 100) {
       : wholeIterations;
 
     // Bulk import.
-    const toBulkImport = [];
-
     for (let i = 1; i <= iterations; i += 1) {
       // If we are on the last loop iteration get from 0-remaining.
       const addToProcess = documents.splice(0, i === iterations ? documents.length : batchSize * 2);
       totalCount += addToProcess.length;
       // Add to bulk import batch.
-      toBulkImport.push(bulkIndex(
+      // eslint-disable-next-line no-await-in-loop
+      await bulkIndex(
         addToProcess,
         AWS_ELASTIC_SEARCH_INDEXES.ACTIVITY_REPORTS,
         client,
-      ));
+      );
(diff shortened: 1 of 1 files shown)


thewatermethod:
Just as an FYI - You likely don't need to go down to 100/per. It's not the number of reports per batch, it's the number of batches running concurrently, right?

My computer is running a job that'll take 20-30 minutes at least so I am going to step out now for a bit. I likely won't approve this before you leave but I'll take a look a bit later

AdamAdHocTeam:
> Just as an FYI - You likely don't need to go down to 100/per. It's not the number of reports per batch, it's the number of batches running concurrently, right?
> 
> My computer is running a job that'll take 20-30 minutes at least so I am going to step out now for a bit. I likely won't approve this before you leave but I'll take a look a bit later

Not 100% sure what the cause of this issue is TBH.

- I've tried running a single bulk update with all reports that gives the error.
- I've tried limiting the number reports in each bulk update that also gives the error.

My thinking with this PR is to process each bulk update (regardless of number 1 or all) one at a time and not concurrently.

Please let me know if you think this wont work or if you have any other suggestions.

thewatermethod:
>  I've tried running a single bulk update with all reports that gives the error.
    I've tried limiting the number reports in each bulk update that also gives the error.
    
Ah ok - I thought all reports gave you an out of memory/137 type error, but bulk update in chunks gives you the 429. If they both give you a 429, then I am a bit confused. either way, this worked when I tried it locally, so makes sense to give it a try on the server

thewatermethod (approved):

PR #1306, opened 2023-01-24
[TTAHUB-1386] ElasticSearch limits results to 10

AdamAdHocTeam:
## Description of change

By default ES limits the number of results to 10. This is because its designed to be used with paging. Because our current hybrid implementation of ES we need to get all our ids up front. The max supported results is 10k but this makes search very slow. With **our export of reports is limited to 2k** we will match this for search results.

## How to test

Searching for something like 'tta' on the AR landing using 'Report text' should yield 2k results.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1386

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


thewatermethod (approved):

thewatermethod:
I think the "2001" fix is a stopgap, and we should have separate messaging that says _the result set_ is incomplete, not just that it contains more than an exportable amount, but let's get this in so that it returns more than 10

AdamAdHocTeam:
> I think the "2001" fix is a stopgap, and we should have separate messaging that says _the result set_ is incomplete, not just that it contains more than an exportable amount, but let's get this in so that it returns more than 10

agree

PR #1323, opened 2023-02-03
Lets make sure we are waiting for search results

AdamAdHocTeam:
## Description of change

Switch search to use a for loop instead of a while.

## How to test

All tests should pass.
Make sure we get back the correct number of results from 'Report text' search.

## Issue(s)

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/lib/awsElasticSearch/index.js
+++ b/src/lib/awsElasticSearch/index.js
@@ -156,10 +156,7 @@ const search = async (indexName, fields, query, passedClient, overrideBatchSize)
     let totalHits = [];
 
     // Loop vars.
-    const maxLoopIterations = 9;
-    let retrieveAgain = true;
     const batchSize = overrideBatchSize || 10000; // Default batch size to 10k.
-    let loopIterations = 0;
     let res;
     let searchAfter;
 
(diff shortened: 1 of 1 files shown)


GarrettEHill:
I don't see how this is effectively different other then changing the max retry from 9 to 5

AdamAdHocTeam:
> I don't see how this is effectively different other then changing the max retry from 9 to 5

In JavaScript a WHILE loop will not wait for the async call to come back. This might lead to some strange overlapping requests. The FOR loop ensures the response comes back before it requests the next batch or results (if there are any).

src/lib/awsElasticSearch/index.js
```diff
@@ -180,8 +177,9 @@ const search = async (indexName, fields, query, passedClient, overrideBatchSize)
         },
       ],
     };
-
-    while (retrieveAgain && loopIterations <= maxLoopIterations) {
+    // Use a FOR loop to force synchronous requests.
```
kryswisnaskas, line 180:
```suggestion
```

src/lib/awsElasticSearch/index.js
```diff
@@ -180,8 +177,9 @@ const search = async (indexName, fields, query, passedClient, overrideBatchSize)
...
+    // This will help us keep the ES queue down.
```
kryswisnaskas, line 181:
```suggestion
```

kryswisnaskas (approved):
Works locally!

AdamAdHocTeam:
This is not needed, as long as a the while statement is in a async function it will work as expected.

PR #1329, opened 2023-02-05
Switch ES to phrase matching

AdamAdHocTeam:
## Description of change

Users want to be able to search for phrases like 'lead water' and ONLY return reports that contain that exact phrase. They don't want to return reports that contain the words 'lead' or 'water'.

Decided to try two different approaches here:
1. If they search a single word we use wild card matching %query% (this will find parts or URL's like 'ECLKC' or 'tta').
2. If they search for two or more words we use phrase matching (this will find matches for 'lead water' in the exact order).

## How to test

I'm hoping this fixes two issues.
1. Users get exact phrase matching reports.
3. It limits the number of matches speeding up most searches.

## Issue(s)

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/lib/awsElasticSearch/index.js
+++ b/src/lib/awsElasticSearch/index.js
@@ -167,9 +167,18 @@ const search = async (indexName, fields, query, passedClient, overrideBatchSize)
     let body = {
       size: batchSize,
       query: {
-        query_string: {
-          query: `*${query}*`,
-          fields,
+        bool: {
+          should: [
+            { match_phrase: { context: { slop: 0, query } } },
+            { match_phrase: { nonECLKCResources: { slop: 0, query } } },
+            { match_phrase: { ECLKCResources: { slop: 0, query } } },
+            { match_phrase: { recipientNextSteps: { slop: 0, query } } },
+            { match_phrase: { specialistNextSteps: { slop: 0, query } } },
+            { match_phrase: { activityReportGoals: { slop: 0, query } } },
+            { match_phrase: { activityReportObjectives: { slop: 0, query } } },
+            { match_phrase: { activityReportObjectivesTTA: { slop: 0, query } } },
+            { match_phrase: { activityReportObjectiveResources: { slop: 0, query } } },
+          ],
         },
       },
       sort: [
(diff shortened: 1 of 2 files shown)


thewatermethod:
We should definitely confirm that this is the behavior Patrice expects before spending any more time on this. I think we need to make clear that the "slop: 0" means that searching _water lead_ means you won't get any _lead water_ results. 

Not that we should necessarily be on the hook for a "google-like" experience, but that is how people expect search engines to work.

AdamAdHocTeam:
> We should definitely confirm that this is the behavior Patrice expects before spending any more time on this. I think we need to make clear that the "slop: 0" means that searching _water lead_ means you won't get any _lead water_ results.
> 
> Not that we should necessarily be on the hook for a "google-like" experience, but that is how people expect search engines to work.

[TTAHUB-1423](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1423)

src/lib/awsElasticSearch/index.js
```diff
@@ -163,15 +163,38 @@ const search = async (indexName, fields, query, passedClient, overrideBatchSize)
...
+          should: [
+            { match_phrase: { context: { slop: 0, query } } },
+            { match_phrase: { nonECLKCResources: { slop: 0, query } } },
+            { match_phrase: { ECLKCResources: { slop: 0, query } } },
+            { match_phrase: { recipientNextSteps: { slop: 0, query } } },
+            { match_phrase: { specialistNextSteps: { slop: 0, query } } },
+            { match_phrase: { activityReportGoals: { slop: 0, query } } },
+            { match_phrase: { activityReportObjectives: { slop: 0, query } } },
+            { match_phrase: { activityReportObjectivesTTA: { slop: 0, query } } },
+            { match_phrase: { activityReportObjectiveResources: { slop: 0, query } } },
```
GarrettEHill, line 189: these should be added
activityReport.additionalNotes
goal/activityReportGoal.timeframe

AdamAdHocTeam: Yea we will add them as we go.

FYI this PR is being replaced by put-back-search-after.

AdamAdHocTeam:
Part of this has already been merged.

The other half (more than 2k results) will be handled in https://github.com/HHS/Head-Start-TTADP/pull/1337.

AdamAdHocTeam:
New PR

PR #1341, opened 2023-02-08
[TTAHUB-1385] Prevent extra objective when changing recipients

thewatermethod:
## Description of change
This changes fixes the bug [described in the jira](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1385), essentially, that objectives weren't being cleaned up properly when recipients were switched, as well as providing a long test that validates all of this behavior.

## How to test
Follow the steps detailed in the jira and note that the bug referenced in the ticket is removed.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1385

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/services/activityReports.js
```diff
@@ -1088,7 +1088,7 @@ export async function createOrUpdate(newActivityReport, report) {
   if (activityRecipientType === 'other-entity' && objectivesWithoutGoals) {
     await saveObjectivesForReport(objectivesWithoutGoals, savedReport);
   } else if (activityRecipientType === 'recipient' && goals) {
-    await saveGoalsForReport(goals, savedReport, recipientsWhoHaveGoalsThatShouldBeRemoved);
+    await saveGoalsForReport(goals, savedReport);
```
thewatermethod, line 1091: the third parameter was not being used by ```saveGoalsForReport```

src/routes/activityReports/saveReport.test.js
```diff
@@ -0,0 +1,614 @@
...
+      name: goalName,
+      createdVia: 'rtr',
+      endDate: new Date(),
+      isRttapa: 'Yes',
+      grantId: secondGrantId,
+      status: GOAL_STATUS.DRAFT,
+    });
+  });
+
+  afterAll(async () => {
```
GarrettEHill, line 122: independentHooks: true needs to be on all updates and destroy to make sure any hooks get run appropriately

GarrettEHill (approved):

kryswisnaskas (commented):
It works for the scenario in the ticket! 👍 

However, when switching things up a bit, we end up with a lost goal. I was able to consistently reproduce the following scenario:

Create a common goal for a recipient with 2 grants on the RTR page
Create a draft report for the first grant and use the common goal with an objective 1
Create a different draft report for the second (not first this time) grant and use the common goal with an objective 2
Realize that a mistake has been made and a wrong grant was selected
Navigate to the Activity Summary and select the correct first (not second this time) grant
Save the report
Navigate to the RTR page
Note that on the RTR page, there is only one goal and activity reports for the second grant don't show the goal as an option (since not in db).

thewatermethod:
> However, when switching things up a bit, we end up with a lost goal. I was able to consistently reproduce the following scenario:

@kryswisnaskas I don't think this bug was introduced here but I've added a fix and test for it to this PR

kryswisnaskas:
Looks good! 👍 

I verified that both use cases work correctly. Yes, I doubt this bug was introduced here, but thank you so much for fixing it. There is still an edge case scenario where the TTA can be wiped out if the auto save kicks in after removing the first recipient and  before adding the second, but that's likely rare and we can either consider it a known bug or handle in a different ticket.

kryswisnaskas (approved):

PR #1346, opened 2023-02-10
Fix legacy ttaType values

hardwarehuman:
## Description of change

Takes legacy values like empty strings and the word "Both" and converts them to the modern equivalnt

## How to test

Run the migration, then check if `SELECT "ttaType", COUNT(*) FROM "ActivityReports" GROUP BY 1;` returns only the four allowed states e.g.:
```
             ttaType             | count
---------------------------------+-------
 {training,technical-assistance} |  2319
 {technical-assistance}          | 20374
                                 |   516
 {training}                      |  4624
 ```

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1275

### Every PR

- [ X ] Meets issue criteria
- [ X ] Code is meaningfully tested (run locally)

### Production Deploy

### After merge/deploy

### CODE REVIEW

File: src/migrations/20230209000000-correct_legacy_ttatype.js [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+              WHEN '{"Both"}' THEN ARRAY['training','technical-assistance']
+              WHEN '{}' THEN NULL
+              ELSE "ttaType"
+            END
+          WHERE "ttaType" NOT IN ('{training}','{training,technical-assistance}','{technical-assistance}');
+          `,
+          { transaction },
+        );
+      } catch (err) {
+        console.error(err); // eslint-disable-line no-console
```
kryswisnaskas, line 38: In this case we don't need the try/catch since the only extra functionality added here is logging, which is already provided by default and it's adequate (same for the above).

kryswisnaskas (approved):
Looks good! 👍 

I verified the numbers are correct after running the migration.

One comment.

src/migrations/20230209000000-correct_legacy_ttatype.js
```diff
@@ -0,0 +1,43 @@
...
+              WHEN '{"Training"}' THEN ARRAY['training']
+              WHEN '{"Both"}' THEN ARRAY['training','technical-assistance']
+              WHEN '{}' THEN NULL
+              ELSE "ttaType"
+            END
+          WHERE "ttaType" NOT IN ('{training}','{training,technical-assistance}','{technical-assistance}');
+          `,
+          { transaction },
+        );
+      } catch (err) {
```
GarrettEHill, line 37: after fixing the values, should the type of the column be changed from varchar[] to enum[] to aid in preventing bad data going forward

PR #1368, opened 2023-02-22
Admin - download active users

kryswisnaskas:
## Description of change
This PR provides a button in the admin UI to enable downloading of all active users. Active users are defined as users that don't have any records in the Permissions table. The downloaded users are in a CSV format.

The implementation for the download includes taking advantage of the streaming api. This is meant as a proof of concept with the idea of reducing the amount of memory that the CSV download needs. Eventually, I'd like to re-write the current implementation of the AR CSV download to potentially remove the limit.

The sequelize/queryGenerator currently has issues correctly mapping the returned data to a model with associations. To avoid the need for a raw query, the plan is to revisit the implementation once this is fixed (hopefully in the next version of sequelize).

## How to test

This change is deployed to Dev. It can be tested there or locally by going to the admin users section.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1414

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/routes/users/handlers.js
```diff
@@ -68,3 +69,23 @@ export async function verifyEmailToken(req, res) {
...
+ * @returns {*} - active users in a CSV format
+ */
+export async function getActiveUsers(req, res) {
+  try {
+    const usersStream = await activeUsers();
+
+    res.writeHead(200, { 'Content-Type': 'text/csv; charset=utf-8' });
+
+    usersStream.on('end', () => res.end());
+    usersStream.pipe(res);
```
GarrettEHill, line 87: I think there should be a check that the user is an admin

src/services/activeUsers.js
```diff
@@ -0,0 +1,124 @@
...
+    stream.push(transform ? transform(r) : r);
+  });
+  query.once('error', (err) => {
+    stream.emit('error', err);
+    end();
+  });
+  query.once('end', end);
+
+  return stream;
+};
```
GarrettEHill, line 66: Is this needed? the result set is less then 600 rows and the columns are small. why not just load all the data at once and process as normal. the "group by"'s are not too complicated  that this could have been done within a normal sequelize model findAll.

src/services/activeUsers.js
```diff
@@ -0,0 +1,124 @@
...
+  FROM
+      "ActiveUsers"
+  GROUP BY
+      "ActiveUsers"."name",
+      "ActiveUsers"."email",
+      "ActiveUsers"."homeRegionId",
+      "ActiveUsers"."lastLogin",
+      "ActiveUsers"."flags";
+      `;
+  const src = streamable(sql, convertToCSV);
```
GarrettEHill, line 122: are you missing an await?

src/services/activeUsers.js
GarrettEHill: I was wrong findAll will not work as the filter expressions are not supported yet https://github.com/sequelize/sequelize/issues/5732

src/services/activeUsers.js
```diff
@@ -0,0 +1,124 @@
...
+      MIN(CASE WHEN "ActiveUsers"."scope" = 'SITE_ACCESS' THEN "ActiveUsers"."scope" END) AS "SITE_ACCESS",
+      MIN(CASE WHEN "ActiveUsers"."scope" = 'ADMIN' THEN "ActiveUsers"."scope" END) AS "ADMIN",
+      MIN(CASE WHEN "ActiveUsers"."scope" = 'UNLOCK_APPROVED_REPORTS' THEN "ActiveUsers"."scope" END) AS "UNLOCK_APPROVED_REPORTS",
+      MIN(CASE WHEN "ActiveUsers"."scope" = 'READ_WRITE_REPORTS' THEN "ActiveUsers"."scope" END) AS "READ_WRITE_REPORTS",
+      array_agg(DISTINCT "ActiveUsers"."regionId") filter (WHERE "ActiveUsers"."scope" = 'READ_WRITE_REPORTS') AS "READ_WRITE_REPORTS regions",
+      MIN(CASE WHEN "ActiveUsers"."scope" = 'READ_REPORTS' THEN "ActiveUsers"."scope" END) AS "READ_REPORTS",
+      array_agg(DISTINCT "ActiveUsers"."regionId") filter (WHERE "ActiveUsers"."scope" = 'READ_REPORTS') AS "READ_REPORTS regions",
+      MIN(CASE WHEN "ActiveUsers"."scope" = 'APPROVE_REPORTS' THEN "ActiveUsers"."scope" END) AS "APPROVE_REPORTS",
+      array_agg(DISTINCT "ActiveUsers"."regionId") filter (WHERE "ActiveUsers"."scope" = 'APPROVE_REPORTS') AS "APPROVE_REPORTS regions",
+      "ActiveUsers"."flags" AS "Flags"
```
GarrettEHill, line 112: https://blog.jooq.org/the-performance-impact-of-sqls-filter-clause/
CASE is slower then filter, so the MIN(CASE columns would be better as string_agg(DISTINCT) filter

thewatermethod (approved):
very cool implementation, I would just double check that the formatting is as expected (abbreviations for roles, each permission as a column, both seem a little weird to me)

src/services/activeUsers.js
```diff
@@ -0,0 +1,124 @@
...
+
+  const result = Object.values(mappedUser).toString();
+
+  return `${result}\n`;
+}
+
+/**
+ * Transform function to convert the data to the right format.
+ *
+ * @param {*} sql - raw sql
```
thewatermethod, line 36: minor, no need to prioritize it, but its a freebie to add the types to the jsdoc
```js
/*
 * @param {string} sql - raw sql
 */
```
etc

thewatermethod: at least the simpler ones, I wouldn't personally bother with the objects, like the user one in convertToCSV

src/routes/users/handlers.js
kryswisnaskas: Added.

src/services/activeUsers.js
kryswisnaskas: This is intentional. The caller of this function uses `await`

src/services/activeUsers.js
kryswisnaskas: Interesting. Initially, I changed to using string agg and filter. The case ended up staying because the requirements changed slightly.

GarrettEHill: In your implementation they are effectively the same output, just the min case is slower.

kryswisnaskas: I actually didn't see any consistent performance improvement on our small dataset when running on my local machine. Perhaps this would be more visible with larger results, though.  In any case, since the requirements changed slightly, we are returning different data now, a string "Yes" rather than the permission name.

PR #1377, opened 2023-02-23
[TTAHUB-1528] Hide site alerts from print view

thewatermethod:
## Description of change
Users should not see site alerts when printing. This change hides the site alerts from print view

## How to test
Print any page of the site, including the approved reports. You will notice that the site alert is not there.

## Issue(s)
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1528

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/pages/ApprovedActivityReport/index.scss
+++ b/frontend/src/pages/ApprovedActivityReport/index.scss
@@ -20,7 +20,8 @@
 
 @media print {
   .smart-hub-offset-nav {
-    max-width: none;
+    margin-top: 0 !important;
+    margin-right: 0;
   }
 
   .smart-hub-offset-nav, .ttahub-activity-report-view {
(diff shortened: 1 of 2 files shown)


frontend/src/pages/ApprovedActivityReport/index.scss
```diff
@@ -20,7 +20,8 @@
 
 @media print {
   .smart-hub-offset-nav {
-    max-width: none;
+    margin-top: 0 !important;
```
AdamAdHocTeam, line 23: It is my understanding that we should avoid !important at all costs. Is there any way we can get around using it?

AdamAdHocTeam (approved):
Looks good.

Not sure if its worth investigating a way around the 'important!'.

frontend/src/pages/ApprovedActivityReport/index.scss
thewatermethod: unfortunately, no - the margin-top is _inlined_ dynamically based on the height of the header + alerts, and the only way to override inline styles is with important. they are both not recommended but I think they are the best thing to do in this case. luckily, using important in a print style does not usually lead the the same problems as using it for regular application stuff

PR #1397, opened 2023-03-07
[TTAHUB-1214] Add and update topics

AdamAdHocTeam:
## Description of change

- Add new topic 'Fatherhood / Male Caregiving'
- Update topic from 'Ongoing Monitoring Management System' to 'Ongoing Monitoring and Continuous Improvement'

## How to test

Create a new report with the new topic changes.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1214

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/migrations/20230307124028-create-and-update-topics.js
```diff
@@ -0,0 +1,87 @@
...
+      );
+
+      // Change existing topic.
+      await queryInterface.sequelize.query(
+        `UPDATE "Topics"
+         SET
+            "name" = 'Ongoing Monitoring and Continuous Improvement',
+            "updatedAt" = current_timestamp
+        WHERE "name" = 'Ongoing Monitoring Management System'; `,
+        { transaction },
```
GarrettEHill, line 35: Do not do this, add the new one, and mark the old one as maps to the new one

hardwarehuman: Is Angela's guidance on this to rewrite _all_ history? If so we also need to update `ActivityReports.topics` and `ActivityReportObjectiveTopics.topicId`. I'm not wild about this as it would mean that the only traces of the original state would be in the audit log, but if Angela says we're supposed to respond to FOIA requests with the revised topic text, then these are the updates to make sure we always do.

AdamAdHocTeam: Confirmed from Angela that Patrice wants historical data changed as well.

thewatermethod: following this discussion and the one on the topic PR only tangentially - should we update the AROTopics table to include topic.name as a string? since we're concerned about preserving the data on the activity reports, we would then have a record that doesn't change after the activity report is approved that we would use as the source of truth on the CSVs and the approved activity report page. I think this would also indicate a clearer report to topic relationship than the mapsTo field. 

Such a change might or might not be part of this PR, but concern over the FOIA'bility of these reports does seem to be something that comes up continuously

GarrettEHill (approved):

PR #1399, opened 2023-03-07
[TTAHUB-1543] Include objective topics in topics frequency graph

AdamAdHocTeam:
## Description of change

On the regional dashboard the topics and frequency graph only shows topics from AR's. We want to include topics from objectives on the AR.

## How to test

Topic counts should reflect both AR's and AR objective topics.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1543

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/widgets/topicFrequencyGraph.js
+++ b/src/widgets/topicFrequencyGraph.js
@@ -4,7 +4,7 @@ import {
   Topic,
   ActivityReportObjective,
 } from '../models';
-import { REPORT_STATUSES, TOPICS } from '../constants';
+import { REPORT_STATUSES } from '../constants';
 
 export default async function topicFrequencyGraph(scopes) {
   const topicsAndParticipants = await ActivityReport.findAll({
@@ -28,7 +28,13 @@ export default async function topicFrequencyGraph(scopes) {
     }],
   });
 
-  const topicsResponse = TOPICS.map((topic) => ({
+  // Get all DB topics.
+  const dbTopics = await Topic.findAll({
+    attributes: ['id', 'name', 'deletedAt'],
+    order: [['name', 'ASC']],
+  });
+  const topics = dbTopics.map((t) => t.name);
+  const topicsResponse = topics.map((topic) => ({
     topic,
     count: 0,
   }));
--- a/src/widgets/topicFrequencyGraph.js
+++ b/src/widgets/topicFrequencyGraph.js
@@ -1,8 +1,9 @@
-import { Op } from 'sequelize';
+import { Op, QueryTypes } from 'sequelize';
 import {
   ActivityReport,
   Topic,
   ActivityReportObjective,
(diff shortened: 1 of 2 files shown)


GarrettEHill:
This is still not using all the available data, some of the topics on the AR's are not the most updated version of the topic value. The topics table has a translation for some. Looks like a few more should be added.

SELECT
	t.topic,
	count(*),
	tt2."name"
FROM "ActivityReports" a
CROSS JOIN UNNEST(a.topics) t(topic)
LEFT JOIN "Topics" tt
ON t.topic = tt.name
LEFT JOIN "Topics" tt2
ON tt."mapsTo" = tt2.id
where t.topic != coalesce(tt2.name,tt.name,'')
group by 1,3
order by 1

src/widgets/topicFrequencyGraph.js
```diff
@@ -13,23 +15,39 @@ export default async function topicFrequencyGraph(scopes) {
...
+      include: [
+        {
+          model: Topic,
+          as: 'topics',
+        },
+      ],
+    }],
   });
 
   const topicsResponse = TOPICS.map((topic) => ({
```
GarrettEHill, line 31: can we use the Topics table to get this instead of the constants list, as it might not be maintained as well as the DB

AdamAdHocTeam:
> 

Not sure I follow what you are saying here. We shouldn't need to jump through any hoops to reflect our database data?

Lets have a quick call.

AdamAdHocTeam:
> This is still not using all the available data, some of the topics on the AR's are not the most updated version of the topic value. The topics table has a translation for some. Looks like a few more should be added.
> 
> SELECT t.topic, count(*), tt2."name" FROM "ActivityReports" a CROSS JOIN UNNEST(a.topics) t(topic) LEFT JOIN "Topics" tt ON t.topic = tt.name LEFT JOIN "Topics" tt2 ON tt."mapsTo" = tt2.id where t.topic != coalesce(tt2.name,tt.name,'') group by 1,3 order by 1

Not sure I follow what you are saying here. We shouldn't need to jump through any hoops to reflect our database data?

Lets have a quick call.

GarrettEHill:
@AdamAdHocTeam 
The data in the topics column on the ActivityReports table has values that are not in the topics table either as current values or as legacy values. We should add the remaining values to the topics table as legacy values mapped to the correct current values.

Additionally, I will again voice my concern that we should not be changing values that are used on approved reports, as they must be FOIA'able in the state on which they were approved. We can mark the deprecated value as deleted and populate the mapsTo field to indicate the value that should be used going forward. This makes for a small amount of work on selection, but will result in valid and consistent data going forward.

GarrettEHill (approved):

PR #1400, opened 2023-03-08
[DEMO] RSS feed fetching and surfacing

thewatermethod:
## Description of change
This is a technical demo created in advance of more work done to surface RSS feeds on our website. This demonstrates the basic pieces that I would suggest we use for implementing this feature.

- Feeds are fetched using our backend
- Stored in redis - we'd likely have to finesse this as far as cache expiration times, but the idea is that a random failed API call doesn't mean a bunch of blank elements on the page
- sent to the frontend for parsing (I would assume the frontend is better at parsing a document, although if the API exists in Node, it may be worth returning the result as JSON)
- it may also be worth using one of the browser storage APIs as a third level of caching. Those don't have the elegant cache expiration management that redis does, but it would one way of ensuring no one ever sees a blank element.

### we still need:
- we'd need a feed stylesheet to handle the HTML that comes back from confluence
- to separate concerns on the backend in keeping with our other services
- API parameters to differentiate RSS endpoints

## How to test
No test required

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1546

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


frontend/src/pages/FeedTest/index.js
```diff
@@ -0,0 +1,50 @@
...
+            key={entry.querySelector('id').textContent}
+            className="padding-1"
+          >
+            <h2 className="margin-0 padding-2">
+              {entry.querySelector('title').innerHTML}
+            </h2>
+            <div
+              className="usa-prose padding-2"
+              // eslint-disable-next-line react/no-danger
+              dangerouslySetInnerHTML={{ __html: entry.querySelector('summary').textContent }}
```
thewatermethod, line 43: this should be using the `<ReadOnlyEditor />` component

src/routes/feeds/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import transactionWrapper from '../transactionWrapper';
+import authMiddleware from '../../middleware/authMiddleware';
+import { generateRedisConfig } from '../../lib/queue';
+
+const {
+  uri: redisUrl,
+  tlsEnabled,
+} = generateRedisConfig();
+
+const feedUrl = 'https://acf-ohs.atlassian.net/wiki/createrssfeed.action?types=page&pageSubTypes=comment&spaces=conf_all&title=Confluence+RSS+Feed&labelString=releasenotes&excludedSpaceKeys%3D&sort=modified&maxResults=10&timeSpan=5&showContent=true&confirm=Create+RSS+Feed&os_authType=basic';
```
thewatermethod, line 13: I think it's totally fine to hardcode this in there but we will obviously need to expand this API to accommodate different content labels in Confluence

src/routes/feeds/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const getFeeds = async (_req: Request, res: Response) => {
+  const redisClient = createClient({
```
thewatermethod, line 16: I think it makes sense to abstract this redis call into something like 
```js
getCachedResponse(key, callback)
```
so we can use it for other things, like the resource dashboard for example

GarrettEHill: Do you want to do that as part of this PR or wait until its needed in more places?

thewatermethod: I'm waiting to actually see how they are using the RSS feed to do anymore work on this, but I expect that to be fairly soon.

thewatermethod: https://github.com/HHS/Head-Start-TTADP/pull/1407/files

PR #1407, opened 2023-03-10
[NO-TICKET] create caching function

thewatermethod:
## Description of change
This is a helper function for caching expensive calculations or exteranl API results in a redis queue.

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/lib/cache.ts
+++ b/src/lib/cache.ts
@@ -1,0 +1,35 @@
+import { createClient } from 'redis';
+import { generateRedisConfig } from './queue';
+
+interface CacheOptions {
+  EX?: number;
+}
+
+const {
+  uri: redisUrl,
+  tlsEnabled,
+} = generateRedisConfig();
+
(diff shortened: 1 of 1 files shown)


src/lib/cache.ts
```diff
@@ -0,0 +1,35 @@
...
+  tlsEnabled,
+} = generateRedisConfig();
+
+const redisClient = createClient({
+  url: redisUrl,
+  socket: {
+    tls: tlsEnabled,
+  },
+});
+
```
thewatermethod, line 19: to implement this function, you need to pass it a key and a callback.
The key is the lookup value. The callback is whatever function you need to call to refresh the cache. It needs to do whatever it's doing but return a string, because that is what can be stored in the redis cache. So functions that call this would probably need to call JSON.stringify / JSON.decode on their own

PR #1411, opened 2023-03-13
[TTHUB-1490] goals missing from csv

thewatermethod:
## Description of change
A bug was discovered where goals without objectives don't appear on the CSV export. This PR fixes that.

## How to test
Export selected reports including goals where there are no objectives and confirm that the goals are present. The Jira has a spreadsheet containing many, (23470 is a good option) although note that some of those are legacy goals, which this PR does not attempt to fix.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1490

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/lib/transform.js
+++ b/src/lib/transform.js
@@ -170,6 +170,68 @@ function sortObjectives(a, b) {
   return -1;
 }
 
+/**
+ *
+ * @param {Object[]} goalRecords
+ * @returns {Object} { goals: [], objectives: []}
+ */
+function makeGoalsAndObjectivesObjectFromActivityReportGoals(goalRecords) {
+  // the number of the goal as it appears in the CSV
+  const goalCsvRecordNumber = 1;
+
+  return goalRecords.reduce((acc, goal) => {
+    const goalId = goal ? goal.id : null;
+    const goalName = goal ? goal.name : null;
+
+    const goalNameIndex = goalName ? Object.values(acc).findIndex(goalName) : -1;
+
+    // if the goal is new we iterate over the properties and update the accumulator
+    if (goalNameIndex !== -1) {
+    // goal id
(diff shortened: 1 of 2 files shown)


src/lib/transform.js
```diff
@@ -170,6 +170,68 @@ function sortObjectives(a, b) {
...
+        writable: true,
+        enumerable: true,
+      });
+    }
+
+    // iterate over the accumulator
+    return acc;
+  }, {});
+}
+
```
nvms, line 234: I refactored this because I don't think we need to use `reduce` or `defineProperty` here, which (in my opinion) simplifies things - otherwise the functions are identical. if I missed something obvious please ignore this, but the tests pass locally for me with this change so I think it's safe

```suggestion
function makeGoalsObjectFromActivityReportGoals(goalRecords) {
  let goalCsvRecordNumber = 1;
  const goals = {};

  goalRecords.forEach((goal) => {
    const {
      id = null,
      name = null,
      status = null,
      createdVia = null,
    } = goal || {};
    const goalNameIndex = Object.values(goals).findIndex((n) => n === name);

    if (goalNameIndex === -1) {
      goals[`goal-${goalCsvRecordNumber}-id`] = `${id}`;
      goals[`goal-${goalCsvRecordNumber}`] = name;
      goals[`goal-${goalCsvRecordNumber}-status`] = status;
      goals[`goal-${goalCsvRecordNumber}-created-from`] = createdVia;
      goalCsvRecordNumber += 1;
      return;
    }

    const goalNameKey = Object.keys(goals)[goalNameIndex];
    const goalNumber = goalNameKey.match(/goal-(\d+)/)[1];
    const field = `goal-${goalNumber}-id`;

    goals[field] = `${goals[field]}\n${id}`;
  });

  return goals;
}
```

PR #1413, opened 2023-03-13
Increase topic frequency performance

AdamAdHocTeam:
## Description of change

The topics frequency was taking 9+ seconds to get the data. This change aims to increase the performance while using the scopes.

## How to test

Time the topics using the old code the new code should be significantly faster (the query also smaller).

## Issue(s)

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


nvms (approved):

src/widgets/topicFrequencyGraph.js
```diff
@@ -8,7 +8,7 @@ import {
 import { REPORT_STATUSES } from '../constants';
 
 export default async function topicFrequencyGraph(scopes) {
-  const topicsAndParticipants = await ActivityReport.findAll({
+  const topicsAndParticipants = await ActivityReport.scope().findAll({
```
thewatermethod, line 11: Not sure that ```.scope()``` is doing anything here or in the other places you are using it. I compared the SQL generated with and without and I saw close to no changes. Looks like any performance improvements must come from bringing in limited attributes.

I'd recommend double checking and removing "scope" if possible so we can reduce confusion as to what it is doing next time someone has to open this file.

Feels like a real performance improvement would be: reduce the number of DB queries, since we are doing 3 here

thewatermethod: other than that, looks good, saw a reduction of several seconds, which is excellent

thewatermethod: I only say that because in the past it's been the time of connection from sequelize to postgres that was the source of slowdown. None of these queries seem particularly onerous so my assumption was that that was the case here

AdamAdHocTeam: @GarrettEHill did a rewrite on this running all queries in parallel along with some postgres magic to get all topics in a single result.

I think this is good to go assuming all tests pass.

src/widgets/topicFrequencyGraph.js
```diff
@@ -71,4 +91,7 @@ export default async function topicFrequencyGraph(scopes) {
 
     return acc;
   }, topicsResponse);
+
+  // console.timeEnd('overallTime2');
```
thewatermethod, line 95: remove console log?

src/widgets/topicFrequencyGraph.js
thewatermethod: ultimately I'm seeing extremely varied results here from the same query, anything from around 2 seconds to around 10. Still I'm presuming this is a performance enhancement of some kind so I'll approve

thewatermethod (approved):

src/widgets/topicFrequencyGraph.js
AdamAdHocTeam: I cleaned this up to remove the timings. I also added back the approved status in the where.

PR #1417, opened 2023-03-14
Add an ability to turn on/off the anniversary statistics flag for all users

kryswisnaskas:
## Description of change
Adds an ability to turn on/off the anniversary statistics flag for all users in the admin UI. Implemented buttons to turn on or off the anv_statistics feature flag for all users since we will be turning it off soon, but perhaps re-use again next year.

## How to test
- navigate to the admin -> feature flags page
- made sure the buttons turn the flag on or off for all users as expected

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1576

### Every PR

- [x] Meets issue criteria
- [n/a] UI review complete

### Production Deploy

### After merge/deploy


frontend/src/pages/Admin/Flags.js
```diff
@@ -47,6 +56,28 @@ export default function Flags() {
                 {' '}
                 &nbsp; &nbsp;
                 <Link to={`/admin/users?flag=${feature}`} aria-label={`view users with the ${feature} feature flag`}>View active users</Link>
+                {' '}
+                &nbsp; &nbsp;
+                {feature === 'anv_statistics' && (
```
thewatermethod, line 61: any reason not to just have this button for every flag?

thewatermethod (approved):
works as advertised, I think we should probably build a full UI for adding and managing feature flags rather than using the enum as we have been

frontend/src/pages/Admin/Flags.js
kryswisnaskas: The plan was to provide a "low-cost" solution that would allow Patrice to better synchronize sending emails to users and turning on the anv statistics flag. There was some concern that if we provided buttons for other flags, admins might "play" with them and accidentally lose a subset of users who had flags enabled. Perhaps we shouldn't be concerned about that, in which case there aren't any other reasons. We may still remove feature flags once they are no longer required.

src/services/users.js
```diff
@@ -1,22 +1,18 @@
...
+import SCOPES from '../middleware/scopeConstants';
 
 import {
   User,
   Permission,
   Role,
   sequelize,
   UserValidationStatus,
-  ActivityReport as ActivityReportModel,
-  ActivityReportCollaborator,
```
AdamAdHocTeam, line 12: Thanks for removing these must have missed it.

AdamAdHocTeam (approved):
Overall looks good and works!

Somethings you might want to consider:

- There is no easy way of knowing the current state of this flag given that users turning it on/off will be an admin and always see the statistics. We might want to hide of show one of the buttons depending on the current state in the DB.

- Has there been any thought given to just leaving it on the landing page with a different title. Given the time that was spent on it (also the landing might look empty).

kryswisnaskas:
> * There is no easy way of knowing the current state of this flag given that users turning it on/off will be an admin and always see the statistics. We might want to hide of show one of the buttons depending on the current state in the DB.

Yes, hiding/showing the current state might be nice. I didn't want to spend too much time at first considering that most flags will go away. Managing state seemed like an overkill at this point.



> * Has there been any thought given to just leaving it on the landing page with a different title. Given the time that was spent on it (also the landing might look empty).

There are plans to incorporate the statistics in a different spot, but good point about the home page becoming empty.

PR #1458, opened 2023-03-30
[TTAHUB-1595] Duplicate Goal Cleanup

kcirtapfromspace:
Remove button to show topic guidance for now

## Description of change
Delete duplicate goals based on trimmed_hashes consolidating into a single goal for a recipient. 

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1595

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


thewatermethod:
> This PR only addresses current duplicates, the path(s) for creating duplicates still exists.

Wouldn't it just be adding ```trim``` to the 2 or 3 places we create a goal? Why not add that with this PR? Otherwise, we will be running this migration again.

kcirtapfromspace:
> > This PR only addresses current duplicates, the path(s) for creating duplicates still exists.
> 
> Wouldn't it just be adding `trim` to the 2 or 3 places we create a goal? Why not add that with this PR? Otherwise, we will be running this migration again.

I was under the assumption this already exists. (as most of the dup goals are from the goal work that went at the tail end of last year.) The only extra dup goal creation that we discovered would be related to TTAHUB-1721

thewatermethod:
Ok. Thanks for the background, @kcirtapfromspace. 

I think my real question is divorced from any solution but more a response to @GarrettEHill's comment
> This PR only addresses current duplicates, the path(s) for creating duplicates still exists.

Is it indeed still a problem? It sounds like you are saying it is not. 
If it is still a problem, is there a Jira for fixing this path, and in that case, wouldn't it be better to run the migration after fixing the problem?

(As an aside, I'm not sure that it isn't still possible to create goals where the only difference from an existing goal is leading or trailing whitespace. I don't see anywhere that we account for it, but if this PR isn't attempting to address that, that's totally fine with me).

kcirtapfromspace:
> Tested the script by arbitrarily creating a few goals and objectives with extra space and then running the migration, original data was untouched but dupes were removed 👍

@thewatermethod Have PR https://github.com/HHS/Head-Start-TTADP/pull/1539 to trim data upon update to DB if you've got a chance to take a look

thewatermethod:
> Have PR https://github.com/HHS/Head-Start-TTADP/pull/1539 to trim data upon update to DB if you've got a chance to take a look

@kcirtapfromspace added a comment there

src/migrations/20230330130554-goal-duplicate-cleanup.js
```diff
@@ -0,0 +1,2281 @@
...
+        )
+        SELECT *, 
+            pre_count - post_count AS diff,
+            post_count - (pre_count - "Deletes" + "Inserts") AS adjusted_diff
+
+        FROM "CollectStats"
+        ORDER BY id;
+        DROP TABLE "PreCountStatsByRegion" ;
+          `, { transaction });
+      } catch (err) {
```
GarrettEHill, line 2273: We need to drop all the temp tables in the script either right before they are created or at the end. for some reason temp tables stick around during back to back migration reseeds and will cause issues.

kcirtapfromspace: Are you thinking that RegionDiffs need also have a temp table create/delete step?
 ```sql
line 24: 
DROP TABLE IF EXISTS "PreCountStatsByRegion";
CREATE TEMP TABLE "PreCountStatsByRegion" AS (
...

line 2155
DROP TABLE IF EXISTS "PostCountStatsByRegion";
CREATE TEMP TABLE "PostCountStatsByRegion" AS (

extra drop:
line 2271  
DROP TABLE "PreCountStatsByRegion" ;
```

GarrettEHill: One of the other PR's inflight will allow reseeding the DB. the result of this is that any migration file using temp tables needs to guard for the temp table already existing. So please add drop if exists before each creation of temp tables.

kcirtapfromspace:
@GarrettEHill  poke for approval

GarrettEHill (approved):

PR #1461, opened 2023-04-04
[TTAHUB-1582] use ID instead of name for group filter

thewatermethod:
## Description of change
The groups filter as originally acted like all the other filters in that the value the user selected was the same as what appears in the filter pills and what appeared in the URL. However, given that the user selects by name and the names are not necessarily unique, this is asking for trouble as far as the future of the groups filter.

This change keeps the user interface and the pill value the same but modifies what appears in the URL to use the group ID, which should be distinct.

## How to test
Create a group and filter on it as before (the recipient record search, the regional dashboard, and the AR landing page all have the filter)

## Issue(s)
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1582

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/__tests__/App.js
+++ b/frontend/src/__tests__/App.js
@@ -62,9 +62,14 @@ describe('App', () => {
 
       const alertsUrl = join('api', 'alerts');
       fetchMock.get(alertsUrl, []);
+      const groupsUrl = join('api', 'groups');
+      fetchMock.get(groupsUrl, []);
 
       const renderApp = () => render(<App />);
-      act(renderApp);
+
+      act(() => {
+        renderApp();
+      });
 
       await waitFor(() => expect(fetchMock.called(alertsUrl)).toBe(true));
     });
--- a/frontend/src/components/MyGroupsProvider.js
+++ b/frontend/src/components/MyGroupsProvider.js
@@ -1,0 +1,40 @@
+import React, {
+  createContext,
+  useEffect,
+  useState,
+} from 'react';
+import PropTypes from 'prop-types';
+import { fetchGroups } from '../fetchers/groups';
+
+export const MyGroupsContext = createContext({});
+
+export default function MyGroupsProvider({ children, authenticated }) {
+  const [myGroups, setMyGroups] = useState([]);
+
+  useEffect(() => {
+    if (authenticated) {
+      try {
+        fetchGroups().then((groups) => {
+          setMyGroups(groups);
+        });
+      } catch (error) {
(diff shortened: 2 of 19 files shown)


frontend/src/__tests__/App.js
```diff
@@ -62,9 +62,14 @@ describe('App', () => {
...
       fetchMock.get(alertsUrl, []);
+      const groupsUrl = join('api', 'groups');
+      fetchMock.get(groupsUrl, []);
 
       const renderApp = () => render(<App />);
-      act(renderApp);
+
+      act(() => {
+        renderApp();
+      });
```
AdamAdHocTeam, line 72: Should you assert the groups call also?

frontend/src/components/MyGroupsProvider.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const MyGroupsContext = createContext({});
+
+export default function MyGroupsProvider({ children, authenticated }) {
+  const [myGroups, setMyGroups] = useState([]);
+
+  useEffect(() => {
+    if (authenticated) {
+      try {
+        fetchGroups().then((groups) => {
```
AdamAdHocTeam, line 17: Do we need an await here?

thewatermethod: nope

frontend/src/components/MyGroupsProvider.js
```diff
@@ -0,0 +1,40 @@
...
+        });
+      } catch (error) {
+        setMyGroups([]);
+      }
+    }
+  }, [authenticated]);
+
+  return (
+    <MyGroupsContext.Provider value={{ myGroups, setMyGroups }}>
+      {children}
```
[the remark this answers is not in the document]
thewatermethod: yeah, it shouldn't

AdamAdHocTeam (commented):
Found a bug if you set a group filter on the AR landing then switch to regional dashboard then back to AR landing you get the below error.

image

thewatermethod:
> Found a bug if you set a group filter on the AR landing then switch to regional dashboard then back to AR landing you get the below error.

@AdamAdHocTeam I am unable to recreate - can you try restarting front & backend?

AdamAdHocTeam (approved):
After restarting my server I could no longer repro the error.
 
I was able to match my group filter counts from main to this branch :-)

PR #1473, opened 2023-04-07
[TTAHUB-1521] - Display regions in AR Landing header/SiteNav

Spanglemonkey:
## Description of change
User's regions are displayed below their name in the SiteNav. If user has access to multiple regions, the AR landing page title will be "Activity reports - your regions" and if they only have access to one region, the title will be "Activity reports - your region". In site nav, there are 3 versions under the user's name: Region #, Regions #, #, #, and All Regions (see ticket with mockups). Also corrected the alignment of the "New Activity Report" button, and spacing between that button and filter button at smaller device widths.

## How to test
- Change user permissions to set user's home ID to region 14, with read permissions to all regions. Should see "All Regions" in SiteNav and "- your regions" in page title.
- Change user permissions to set user's home ID to region 1, with read permissions to region 1 and 5. Should see "Regions 1,5" in SiteNav and "- your regions" in page title.
- Change user permissions to set user's home ID to region 5, with read permissions to region 5. Should see "Region 5" in SiteNav and "- your region" in page title.
- Reduce window size to where "New Activity Report" button moves below page title. There should be 24px between title and "New Activity Report" button and 16px between Report button and filter button. 

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1521

### Every PR

- [x] Meets issue criteria
- [x] UI review complete

### Production Deploy

### After merge/deploy


Spanglemonkey:
I could not do all the testing in my local env, because I could not change permissions on my user. Also, @thewatermethod I had to add a third scenario to the region names in the SiteNav (if user had home ID of Region 14), which meant I had to add another React hook, import and const, compared to what you saw originally.

frontend/src/components/SiteNav.js
```diff
@@ -62,18 +61,20 @@ const SiteNav = ({
...
+    if (defaultRegion === 14) {
+      return 'All Regions';
+    }
+
     if (hasMultipleRegions) {
       return `Regions ${regions}`;
     }
-    if (defaultRegion === 14) {
-      return 'All Regions;';
-    }
```
thewatermethod: Flipped the order here since putting "hasMultipleRegions" first means you'd basically never hit that second clause. My understanding is that the "defaultRegion if" should take precendence.

Also removed a semicolon typo

frontend/src/pages/Landing/index.js
```diff
@@ -204,9 +204,6 @@ function Landing() {
     if (defaultRegion === 14 || hasMultipleRegions) {
       return 'your regions';
     }
-    if (defaultRegion === 14) {
-      return 'your regions';
-    }
```
thewatermethod, line 209: this is just a bit simpler

thewatermethod:
There's a logic error here that I missed before.

```js
  const regions = allRegionsUserHasPermissionTo(user).join(', ');
  const defaultRegion = user.homeRegionId || regions[0] || 0;
  const hasMultipleRegions = regions && regions.length > 1;
```  
Essentially, the ```join``` makes regions a string. So in our example, the ```allRegionsUserHasPermissionTo``` returns an array ```[11]``` and then we immediately turn it into a string, ```"11"```. So defaultRegion could access the first character of the string with ```regions[0]``` and ```hasMultipleRegions``` is true if the string is longer than 1 character.

See below for the debugger's output:  
Screenshot 2023-04-12 at 10 37 02 AM

We should store the joined value in a separate variable after we've evaluated ```defaultRegion``` and ```hasMultipleRegions```. We should probably also revise the test to catch this case.

frontend/src/components/SiteNav.js
```diff
@@ -56,6 +60,24 @@ const SiteNav = ({
     }
   }, [hasAlerts]);
 
+  // Determine Default Region.
+  const regions = allRegionsUserHasPermissionTo(user).join(', ');
+  const defaultRegion = user.homeRegionId || regions[0] || 0;
```
nvms, line 65: `regions` ends up being a string here.

if the user has permissions to regions 11 and 12, `regions` will be the string `11, 12`. this means that if `user.homeRegionId` is falsy, then `defaultRegion` will end up being `1` (because `regions[0] accesses the character at position `0` in the string `11, 12`, which is `1`). what we probably want is for it to be 11.

maybe it makes sense to return the string to an array before accessing the first element, like:

```typescript
const defaultRegion = user.homeRegionId || regions.split(', ')[0] || 0;
```

frontend/src/components/SiteNav.js
```diff
@@ -56,6 +60,24 @@ const SiteNav = ({
...
+  const hasMultipleRegions = regions && regions.length > 1;
```
nvms, line 66: because of the behavior mentioned above, this is always true if the string is longer than one character. this means that if I only have one region, region `10`, `hasMultipleRegions` will be true because the length of the string is longer than 1.

we need to make sure this is an array before checking its length.

```typescript
const regionsArray = regions.split(', ');
const hasMultipleRegions = regionsArray && regionsArray.length > 1;
```

a test to verify both of the things i've mentioned so far might look like this:

```typescript
    test('has no home region', async () => {
      const user = {
        // since I have no home region id, the `defaultRegion` falls back
        // to `regions.split(', ')[0] (or however else you want to implement this fix)
        homeRegionId: null,
        permissions: [
          {
            scopeId: SCOPE_IDS.READ_WRITE_ACTIVITY_REPORTS,
            regionId: 10,
          },
        ],
      };

      act(() => {
        renderSiteNav(user);
      });

      // this verifies `hasMultipleRegions` works as expected
      // because the heading is `Region` and not `Regions`, as we
      // only have one region.
      expect(await screen.findByText('Region 10')).toBeVisible();
    });
```

nvms (dismissed):
looks good so far, but some minor changes are needed in `SiteNav`

I've provided a test that should pass once the changes are made, and should also boost coverage since the `if (hasMultipleRegions)` branch isn't currently covered

edit: I just realized that @thewatermethod pointed out this exact issue already. hah! I should have read before reviewing.

thewatermethod (approved):

AdamAdHocTeam (approved):
Overall looks good.

One thing I noticed, but it might be by design. Is that if I have access to all regions but my home region is 1 (even though my role is central office) I still see the below. Setting the home region to 'central office' fixes it.

image

GarrettEHill:
If this is approved, why is it still waiting to be merged?

thewatermethod:
@Spanglemonkey FYI - main is merged in, everything is passing. Tested locally as well, still seems to work as advertised.

kryswisnaskas:
> @Spanglemonkey FYI - main is merged in, everything is passing. Tested locally as well, still seems to work as advertised.

Looks like we can merge this PR.

PR #1475, opened 2023-04-10
[TTAHUB-44] add foreign key and some predicate indices

hardwarehuman:
## Description of change

Adds indices in places where it's most likely to help database performance more than it hurts it.

## How to test

- deploy and watch the indexes get created
- make sure there are no unexplained slowdowns when writing new data. adding indexes adds a small amount of overhead to database write operations that shouldn't be noticeable, but could be in combination with other issues.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-44

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/migrations/20230408000000-add-foreign-key-indices.js
```diff
@@ -0,0 +1,78 @@
...
+    // Skipping creating a userProvidedUrl index because it's just a text field
+    queryInterface.addIndex('ObjectiveTemplateResources', ['objectiveTemplateId'], { transaction }),
+    queryInterface.addIndex('ObjectiveTemplateTopics', ['objectiveTemplateId'], { transaction }),
+    queryInterface.addIndex('ObjectiveTemplateRoles', ['objectiveTemplateId'], { transaction }),
+    queryInterface.addIndex('Goals', ['grantId'], { transaction }),
+    queryInterface.addIndex('Goals', ['goalTemplateId'], { transaction }),
+    queryInterface.addIndex('GoalTemplateObjectiveTemplates', ['goalTemplateId', 'objectiveTemplateId'], { transaction }),
+    queryInterface.addIndex('GoalTemplateObjectiveTemplates', ['objectiveTemplateId'], { transaction }),
+    queryInterface.addIndex('GroupGrants', ['groupId', 'grantId'], { transaction }),
+    queryInterface.addIndex('GroupGrants', ['grantId'], { transaction }),
```
GarrettEHill, line 34: Could this list be generated via making indexes off all present foreign keys per table, factorial combination,  populate it as via a function. make a trigger for the function on table create or alter to automatically be added as the db changes

hardwarehuman: I think we want to be more judicious in our addition of indexes off foreign keys, because sometimes they add more write overhead than it helps with reading. If the db was being used purely transactionally and never for charts, even fewer foreign keys would be useful. As it was, I actually removed half a dozen indices after reconsidering whether they would really help any actual query the system might perform.

kryswisnaskas (approved):
Looks good! 👍 

We just need to merge main into this branch to remove the errors and we can move it forward.

GarrettEHill:
I think we might want to look at using the "CONCURRENTLY" key word for the indexes to remove the potential for performance issues. we might even need to add this to all the current indexes too.

hardwarehuman:
@GarrettEHill @kryswisnaskas We don't need to worry about the CONCURRENTLY option, which is an index building approach that avoid table locking during index creation and (as far as I know) is never used thereafter. Because we're running in a migration while no one else is using the db it's not necessary.

GarrettEHill:
@hardwarehuman after more research, I agree. It will not provide any value add CONCURRENTLY.

kryswisnaskas:
> Because we're running in a migration while no one else is using the db it's not necessary.

Sounds like we need to coordinate with Patrice on the deploy timing so she can send notifications. The migration run quickly in the local environment with a production snapshot (3.8 s). Does this align with what you are seeing and expecting as well?

GarrettEHill:
@hardwarehuman I think you just need to merge in main to fix the current issue

PR #1479, opened 2023-04-10
[TTAHUB-1398] update ARs and link tables to use current topics

hardwarehuman:
## Description of change

We're basically "rewriting history" to update old topic names to use the current texts. This is evidently FOIA-acceptable.

## How to test

The main visible difference is the topic heatmap will no longer show an entry for `Child Assessment, Development, Screening`. Other statistics will also shift to show the current topic names (`Child Screening and Assessment` and `Teaching / Caregiving Practices`) to be slightly more heavily used by adding in older records.

```
             tablename             | updates
-----------------------------------+---------
 updated_activity_reports          |       2
 updated_objective_template_topics |    3404
 updated_objective_topics          |     820
 updated_role_topics               |       0
 updated_ar_objective_topics       |     806
 ```

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1398

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20230408000001-rewrite_to_use_current_topics.js
+++ b/src/migrations/20230408000001-rewrite_to_use_current_topics.js
@@ -1,0 +1,120 @@
+/* eslint-disable max-len */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const loggedUser = '0';
+      const sessionSig = __filename;
+      const auditDescriptor = 'RUN MIGRATIONS';
+      await queryInterface.sequelize.query(
+        `SELECT
+                set_config('audit.loggedUser', '${loggedUser}', TRUE) as "loggedUser",
+                set_config('audit.transactionId', NULL, TRUE) as "transactionId",
+                set_config('audit.sessionSig', '${sessionSig}', TRUE) as "sessionSig",
+                set_config('audit.auditDescriptor', '${auditDescriptor}', TRUE) as "auditDescriptor";`,
+        { transaction },
+      );
+      // Update all the existing link tables to use the current topics,
+      // then update ActivityReports.topics
+      await queryInterface.sequelize.query(
+        `
+        CREATE TEMP TABLE renamed_topics AS
+        SELECT DISTINCT
+          lt.id legacy_tid,
+          lt."mapsTo" current_tid,
+          lt.name legacy_name,
+          ct.name current_name
+        FROM "Topics" lt
+        JOIN "Topics" ct
+          ON lt."mapsTo" = ct.id
+        WHERE lt.name IN ('Teaching Practices / Teacher-Child Interactions', 'Child Assessment, Development, Screening')
+        ;
+        
+        CREATE TEMP TABLE updated_objective_topics AS
+        WITH updater AS (
+          UPDATE "ObjectiveTopics"
+          SET "topicId" = current_tid
+          FROM renamed_topics
+          WHERE "topicId" = legacy_tid
+          RETURNING
+            id updated_id,
+            legacy_tid
+        ) SELECT * FROM updater
+        ;
+
+        CREATE TEMP TABLE updated_ar_objective_topics AS
+        WITH updater AS (
+          UPDATE "ActivityReportObjectiveTopics"
+          SET "topicId" = current_tid
+          FROM renamed_topics
+          WHERE "topicId" = legacy_tid
+          RETURNING
+            id updated_id,
+            legacy_tid
+        ) SELECT * FROM updater
+        ;
+        
+        CREATE TEMP TABLE updated_objective_template_topics AS
+        WITH updater AS (
+          UPDATE "ObjectiveTemplateTopics"
+          SET "topicId" = current_tid
+          FROM renamed_topics
+          WHERE "topicId" = legacy_tid
+          RETURNING
+            id updated_id,
+            legacy_tid
+        ) SELECT * FROM updater
+        ;
+        
+        CREATE TEMP TABLE updated_role_topics AS
+        WITH updater AS (
+          UPDATE "RoleTopics"
+          SET "topicId" = current_tid
+          FROM renamed_topics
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/migrations/20230408000001-rewrite_to_use_current_topics.js [hunk 1]
```diff
@@ -0,0 +1,120 @@
...
+        CREATE TEMP TABLE renamed_topics AS
+        SELECT DISTINCT
+          lt.id legacy_tid,
+          lt."mapsTo" current_tid,
+          lt.name legacy_name,
+          ct.name current_name
+        FROM "Topics" lt
+        JOIN "Topics" ct
+          ON lt."mapsTo" = ct.id
+        WHERE lt.name IN ('Teaching Practices / Teacher-Child Interactions', 'Child Assessment, Development, Screening')
```
kcirtapfromspace, line 29: After all topics are updated to the correct current_names will there be a followup to remove legacy topic names?

hardwarehuman: My plan was to leave the historical topic names in the Topics table, which is the approach we've taken in the past.

GarrettEHill:
No, please don't update the ARs, we already have code inplace to map the old values to the new ones for all the statistical reporting

src/migrations/20230408000001-rewrite_to_use_current_topics.js
GarrettEHill: We dont need to do this, we do the mapping live when stats are collected

kryswisnaskas (commented):
Looks good! I verified that the migration updates the topics and links in almost all of the places.
Could we also rename the topics in the legacy field for some report so that the old topic doesn't show for reports like the following:

http://localhost:3000/activity-reports/legacy/R02-AR-000171

There is a number of these reports:
```
select id, imported->>'topics', * from "ActivityReports"
where imported is not NULL and imported->>'topics' like 'Child%'
```

hardwarehuman:
@kryswisnaskas The import topic strings are a bit messy in that they have multiple newline-separated topics, but even after newline separation there's sometimes target populations (e.g. ECS) in them after a pipe separation ( actually a space-pipe-space separation). This only matters if we are going to add another topic translator for legacy reports following the pattern @GarrettEHill and @nvms  already established; updating the text in place would be very easy. I updated the Jira with a question as to whether we need to update the legacy reports.

If not, then this PR doesn't even need to be merged, as the rest of the problem is fixed by a fix to translation in https://github.com/HHS/Head-Start-TTADP/pull/1483. If we *do* need to fix legacy reports, then we have to decide if we should add another translator or if we're okay with the easy path of switching this branch to a migration modifying the old import data.

kryswisnaskas:
In the past we've modified the topics in the imported field, so even if we wanted to achieve some sort of historical record, it wouldn't be accurate. Additionally, the customer indicated that this wasn't desired. Given this, if changing the `imported `field is easy, then please, let's go ahead with it. 

Also, without the changes in this PR we still get a couple of ARs that show the old values in the UI. Specifically, reports 23865, 23869. The CSV download for these reports is also affected.

hardwarehuman:
@kryswisnaskas It's now updated with just the import data update, but the other data updates commented out. After this runs there shouldn't be any remnants of the old data visible to Hub users. Probably it would be best to remove the other updates entirely instead of leaving in a large commented section, but I'm still thinking if there's anything else to test.

kryswisnaskas:
Looks good as far as the legacy reports display! 👍 Thank you!  However, I still see the 2 reports mentioned above showing the old values. I can provide screenshots if needed.

hardwarehuman:
> Looks good as far as the legacy reports display! 👍 Thank you! However, I still see the 2 reports mentioned above showing the old values. I can provide screenshots if needed.

@kryswisnaskas  I just merged in `main` to bring in https://github.com/HHS/Head-Start-TTADP/pull/1483 which I should have done  before to address that. Sorry!

kryswisnaskas:
For some reason, these 2 reports still show the old value:

image

image

hardwarehuman:
@kryswisnaskas I'm seeing that too. @GarrettEHill would you expect this to be addressed? It would be easy to just uncomment the rest of the data changes to push this through and clean everything up, but maybe we're missing something

kryswisnaskas:
It's interesting there are only 2 reports like that. I think in this case opting for the easy solution makes sense.  Alternatively, we could investigate why these are not converting, but the effort spent would not be worth it, wouldn't you agree?

kryswisnaskas (approved):
Looks good! Thank you.

GarrettEHill:
> For some reason, these 2 reports still show the old value:

Approved Reports should show the value they were written with.

GarrettEHill:
We can not keep changing FOIA documents after they are approved.

kryswisnaskas:
> We can not keep changing FOIA documents after they are approved.

I understand the concern. Given that this was requested by the customer, and with the audit log, we have a record of the change if a reconstruction of the original topic was ever requested.

GarrettEHill:
Did the customer ask for the reports to have only the new values or analytics/statistics? I think this is difference we keep running up against and we need to have a defined hard policy set by the customer. Who is responsible to maintain the legitimacy of the FOIA'able records? Who will be reprimanded if a variance is found? I think they really only want the analytics/statistics and all new reports going forward to use the new values.

hardwarehuman:
> Did the customer ask for the reports to have only the new values or analytics/statistics? I think this is difference we keep running up against and we need to have a defined hard policy set by the customer. Who is responsible to maintain the legitimacy of the FOIA'able records? Who will be reprimanded if a variance is found? I think they really only want the analytics/statistics and all new reports going forward to use the new values.

Yes, Angela had a specific discussion with Patrice about it and our guidance from the customer is that the data itself can be changed under FOIA rules because it's the "same" topic. We want to retain the ability to fix it if they ever change their minds, but for now (and for a while now) that is their actual guidance on topics.

PR #1503, opened 2023-05-01
[TTAHUB-1688] Allow editing of FEI goals on the RTR

thewatermethod:
## Description of change
Several needed improvements to the conditional prompts/FEI goal feature. Most pertinently, the ability to edit an existing curated goal's prompts has been added to the RTR's edit goal form.

## How to test
- Create a multi-**recipient** report. Select the FEI goal, and approve the report. You should be unable to select FEI root causes for the goal, since it is a multi-recipient report.
- Create a second report, using one of the two previous recipients. Update the goal's FEI root cause, submit, and approve the report.
- Go to the RTR and the goals and objectives tab for the first recipient. You should see the FEI goal, and when you edit it, you'll see the root cause as read only.
- Go to the RTR and the goals and objectives tab for the second recipient. Since you've never added root causes via an activity report, you should be able to edit the root causes on this recipient's FEI goal. Ensure you can do so without issue.

## Issue(s)
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1688

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


kryswisnaskas:
Nice! In the most part it works as advertised, but I did run into an error:

- create a report with 2 recipients, submit and approve
- create a second report with the first of the above recipients
- select the FEI goal and 2 root causes
- "Save goal"
- edit the newly created goal
- note that there is no way to edit the root causes and "Save draft" produces: `Error(`Response for '${promptRequirements.title}' is required.`

kryswisnaskas:
Looks good now with the fix! 🚀 

One outstanding question still and this is more of a business question. Did we decide that as soon as 2 root causes are entered and the goal saved, there is no way to edit the root causes?

This might become a source of frustraction in a case when someone just made a mistake and wants to correct it before going any further. Since it's no longer editable the user might try removing the goal and re-selecting it. In this case they might lose the tta, but not being able to edit the root causes anyway. They might also try deleting the activity report thinking that starting from scratch would make a difference (but it doesn't).

thewatermethod:
> One outstanding question still and this is more of a business question. Did we decide that as soon as 2 root causes are entered and the goal saved, there is no way to edit the root causes?

We did, since the goal is on an approved report (I originally demoed this with the ability to perpetually change the root causes and was told to adjust)

kryswisnaskas:
> st)

Then emphasizing the importance of user guidance makes sense. Maybe we can remind everyone. Users will understand that once the "Save goal" button is pressed, the root causes are locked.

kryswisnaskas (approved):

thewatermethod:
> Then emphasizing the importance of user guidance makes sense. Maybe we can remind everyone. Users will understand that once the "Save goal" button is pressed, the root causes are locked.

I am meeting with Patrice to demo some other changes later today, I will confirm this is the expected behavior with her then

thewatermethod:
>  I am meeting with Patrice to demo some other changes later today, I will confirm this is the expected behavior with her then

Confirmed that this is the expected behavior

PR #1523, opened 2023-05-09
[TTAHUB-955] Generic Report

GarrettEHill:
## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-955

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/migrations/20230807000000-genericReport-migrateTrainingReports.js
```diff
@@ -0,0 +1,463 @@
...
+          UNNEST(
+            string_to_array(jsonb_array_elements_text(data->'targetPopulations'),E'\n')
+          ) target_population
+        FROM "EventReportPilots" erp
+        )
+        SELECT
+          reports_id,
+          utp.id,
+          "createdAt",
+          "updatedAt"
```
GarrettEHill, line 283: @hardwarehuman: I know is not explicitly needed, but please include the table alias prefix to the selected columns so its easier to follow were each field comes from.

src/migrations/20230807000000-genericReport-migrateTrainingReports.js
```diff
@@ -0,0 +1,913 @@
...
+        -- Reused Collaborator tables
+        DROP TABLE IF EXISTS owner_status;
+        CREATE TEMP TABLE owner_status
+        AS
+        SELECT s.id sid
+        FROM "Statuses" s
+        JOIN "ValidFor" v
+          ON s."validForId" = v.id
+          AND v.name = 'collaborator'
+        WHERE s.name = 'approved' -- TODO: this is terminal, which could lock editing, but the others statuses make even less sense for events
```
GarrettEHill, line 376: @hardwarehuman The collaborator statusId should be left null for any non-approving user

src/constants.js
```diff
@@ -49,6 +49,13 @@ const OBJECTIVE_STATUS = {
   COMPLETE: 'Complete',
 };
 
+const OBJECTIVE_SUPPORT_TYPES = [
```
thewatermethod, line 52: this should be in the common package

package.json
```diff
@@ -305,6 +305,8 @@
     "csv-parse": "^4.14.1",
     "csv-stringify": "^5.6.2",
     "cucumber-html-reporter": "^5.2.0",
+    "deepmerge": "^4.3.1",
+    "dot-wild": "^3.0.1",
```
thewatermethod, line 309: ```dot-wild``` is a little worrying to me, there are security related issues on the repo and it hasn't been updated in 6 years

thewatermethod: A few more files down, I see it would be difficult to replace as it seems to be deeply intertwined with the approach we took here

src/lib/modelUtils.ts
```diff
@@ -0,0 +1,335 @@
...
+  } else if (Array.isArray(data)) {
+    remappedData = [];
+  } else {
+    remappedData = {};
+  }
+
+  // Iterate over each key in the remapping definition
+  Object.keys(remappingDefinition).forEach((key) => {
+    // Determine the source and target paths based on the reverse flag
+    // eslint-disable-next-line no-nested-ternary
```
thewatermethod, line 182: I'm taking eslint's side here I think

src/lib/modelUtils.ts
```diff
@@ -0,0 +1,335 @@
...
+  && !Array.isArray(value) // Check if the value is not an array
+);
+
+/**
+ * Checks if two values are deeply equal.
+ * @param value1 - The first value to compare.
+ * @param value2 - The second value to compare.
+ * @returns True if the values are deeply equal, false otherwise.
+ */
+const isDeepEqual = (value1: any, value2: any): boolean => {
```
thewatermethod, line 247: lodash has this I think if you want to just use theirs

src/lib/modelUtils.ts
```diff
@@ -0,0 +1,335 @@
...
+  return { mapped: remappedData, unmapped: unmappedData };
+};
+
+/**
+ * Checks if a value is an object.
+ *
+ * @param value - The value to be checked.
+ * @returns A boolean indicating whether the value is an object or not.
+ */
+const isObject = (value: any): boolean => (
```
thewatermethod, line 235: I believe this is equivalent to "isPlainObject" from lodash if we don't want to maintain our own (we're already using lodash all over the place, so it wouldn't be adding a dependency)

src/lib/modelUtils.ts
thewatermethod: "isEqual"

src/lib/modelUtils.ts
```diff
@@ -0,0 +1,335 @@
...
+  [DataTypes.INET.key]: 'string',
+  [DataTypes.MACADDR.key]: 'string',
+  [DataTypes.CITEXT.key]: undefined,
+  [DataTypes.TSVECTOR.key]: undefined,
+  // Add more mappings as needed
+};
+
+// Get the column information using the describe method
+const getColumnInformation = async (model: typeof Model) => {
+  const tableDetails = await model.describe();
```
thewatermethod, line 50: do you think this needs a try/catch

src/lib/modelUtils.ts
```diff
@@ -0,0 +1,335 @@
...
+    allowNull: columnDetails.allowNull,
+  }));
+  return columns;
+};
+
+const getColumnNamesFromModelForType = async (
+  model: typeof Model,
+  type,
+) => {
+  const modelData = await getColumnInformation(model);
```
thewatermethod, line 63: do you think this needs a try/catch?

thewatermethod: I won't iterate this question out for each of these functions, but same question for the rest of the stuff in this file, especially the ones that could have rejected promises in the middle, or ones with a bunch of second-level object property accessing. 

Once this builds successfully, I'm also curious about the test coverage with all these nested if statements. It's possible some of this could be simplified if we find paths of code that are so unlikely as to not need tests.

package.json
GarrettEHill: https://security.snyk.io/package/npm/dot-wild I don't see any vulnerabilities. What issues do you see?

thewatermethod:
Scanning ahead, I noticed a bunch of empty files - I'm assuming those aren't to be deleted (i.e. still a WIP)?

src/lib/modelUtils.ts
```diff
@@ -0,0 +1,335 @@
...
+        // Add the changed value to the changedValues object
+        changedValues[key] = value;
+      }
+    });
+
+  // Return the object containing the changed values
+  return changedValues;
+};
+
+const includeToFindAll = async (
```
thewatermethod, line 307: Brief comment explaining this function

src/lib/modelUtils.ts
GarrettEHill: do you mean "no-nested-ternary"?

src/lib/modelUtils.ts
GarrettEHill: Sure, what do you propose we do in the catch? Log, call the handleErrors, something else?

src/lib/modelUtils.ts
GarrettEHill: great thought. I don't know. I have tests for some if it. But I don't know how to see coverage when I am still failing lint as a whole.

package.json
GarrettEHill: It had not been updated in 6 years, agreed. But it works great.

src/migrations/20230802000000-genericReport.js
```diff
@@ -0,0 +1,2853 @@
...
+        name: {
+          type: Sequelize.TEXT,
+          allowNull: false,
+        },
+        isTerminal: {
+          type: Sequelize.BOOLEAN,
+          allowNull: false,
+          default: false,
+        },
+        ordinal: {
```
thewatermethod, line 206: what does ordinal mean in a case like "suspended"?

src/lib/modelUtils.ts
```diff
@@ -0,0 +1,335 @@
...
+  [DataTypes.DECIMAL.key]: 'number',
+  [DataTypes.BOOLEAN.key]: 'boolean',
+  [DataTypes.TIME.key]: 'string',
+  [DataTypes.DATE.key]: 'string',
+  [DataTypes.DATEONLY.key]: 'string',
+  [DataTypes.HSTORE.key]: undefined,
+  [DataTypes.JSON.key]: 'object',
+  [DataTypes.JSONB.key]: 'object',
+  [DataTypes.NOW.key]: undefined,
+  [DataTypes.BLOB.key]: undefined,
```
AdamAdHocTeam, line 30: So these undefined mappings simply don't have a corresponding value?

src/migrations/20230802000000-genericReport.js
GarrettEHill: for a set of statuses ( same valid for) ordinal defines the order of progression of the statuses.
example:
"draft" 1
"not started" 2
"in progress" 3
"suspended" 4
"complete" 5

src/lib/modelUtils.ts
GarrettEHill: correct

GarrettEHill: the list is all the types defined in the sequelize.DataTypes. but many are very rarely ever used in any context, and not in our context. The only one that is marked as unknow that we use on any table we use is DataType.ARRAY. To have a mapping for that I would also need to identify the type of the item in the array. So I think this is a not going to be as problem as all the tables in this PR don't use the array datatype.

src/models/helpers/associationsAndScopes.js
```diff
@@ -0,0 +1,424 @@
+/**
+ * Converts a PascalCase string to camelCase.
+ *
+ * @param {string} input - The PascalCase string to convert.
+ * @returns {string} - The converted camelCase string.
+ */
+const pascalToCamelCase = (input) => (input
```
thewatermethod, line 7: _.camelCase(input)

src/models/helpers/isFlagged.ts
```diff
@@ -0,0 +1,626 @@
...
+
+/**
+ * This function automatically populates a flag field in an instance if it is undefined or null.
+ * It also ensures that the flag field is included in the fields array of options.
+ * @param flagName - The name of the flag field.
+ * @param sequelize - The Sequelize instance.
+ * @param instance - The instance to populate the flag field for.
+ * @param options - The options object containing the fields array.
+ */
+const autoPopulateIsFlagged = (
```
thewatermethod, line 12: this doesn't seem to use the sequelize instance

src/models/helpers/isFlagged.ts
```diff
@@ -0,0 +1,626 @@
...
+ * This function is responsible for auto-populating a flag in an instance of a Sequelize model.
+ * It checks if the 'autoPopulateFlag' option is set to true, and if not, it sets the flag to
+ * false and adds it to the list of fields to be saved.
+ *
+ * @param sequelize - The Sequelize instance.
+ * @param instance - The instance of the Sequelize model.
+ * @param options - The options object.
+ * @param flagName - The name of the flag to be auto-populated.
+ */
+const autoPopulateFlag = (
```
thewatermethod, line 590: this is nearly identical to the function at the top of this file, autoPopulateIsFlagged, if we need the check at the top of this function, maybe condense so we aren't repeating so much

src/lib/modelUtils.ts
thewatermethod: Depends on whether you think it makes sense to return an empty array or blow up

src/models/helpers/isFlagged.ts
GarrettEHill: your right, I have been working on this for too long. Thanks I will reduce and reuse

src/migrations/20230802000000-genericReport.js
thewatermethod: suspended isn't part of the progression, that's why I asked. In the UI, suspended is the final option. Trying to get at how we're using "ordinal"

src/services/goalsAndObjectives/goalTemaplate.ts
```diff
@@ -0,0 +1,106 @@
+// TODO: everything
```
thewatermethod, line 1: template is mispelled in the filename

src/services/enums/reasons.ts
```diff
@@ -0,0 +1,42 @@
+import * as genericEnum from './generic';
+import db from '../../models';
+import { ENTITY_TYPE } from '../../constants';
+
+const { Reasons } = db;
+
+const findAll = async (
+  data: { validFor: typeof ENTITY_TYPE[keyof typeof ENTITY_TYPE] },
+) => genericEnum.findAll(Reasons, data);
```
thewatermethod, line 9: many of these functions could be made generic, IE, we only need a separate findAll for reasons and status if they are doing something besides passing a model through.

thewatermethod: Also, in this case, I believe the tablename should be Reason (singular)

GarrettEHill: I have a generic implementation, these functions call it. They pass into it the "unique" thing each needs. How would you change it?

package.json
GarrettEHill: @thewatermethod What concerns do you have other then it has not been touched in 6 years?

src/migrations/20230802000000-genericReport.js
```diff
@@ -0,0 +1,2853 @@
...
+       * - ObjectiveTemplateResources - isFoiaable
+       * - ObjectiveTemplateTopics - isFoiaable
+       * - Objectives - isFoiaable
+       * - ObjectiveFiles - isFoiaable
+       * - ObjectiveResources - isFoiaable
+       * - ObjectiveTopics - isFoiaable
+       *  */
+
+      //---------------------------------------------------------------------------------
+      await queryInterface.createTable('Foiaable', {
```
AdamAdHocTeam, line 102: Prob havent gotten there yet but how do we use Foiaable and validFor tables?

src/migrations/20230802000000-genericReport.js
GarrettEHill: Ok, so suspended needs to be last, We can make that change.

src/migrations/20230802000000-genericReport.js
GarrettEHill: Foiaable table identifies which columns for which tables are "read-only" after the isFoiaable flag is set. This allows some values to still be updated while maintaining the integrity of other columns.

ValidFor is used in conjunction with many of the enum-esk tables to allow similar enums to be located in the same table. Like "Statuses" hold the valid statuses for Training Events, Training Sessions, Goals, Objectives, and Collaborators. So they all reside in the same table and have the mapsTo construct to allow extensibility without each requiring their own table. 

Future-future plan would be to have an admin page to  manage these enum-esk values. Think Audience, national centers, participants, reasons, roles, statuses, support types, target populations, and topics. Let them manage the values. Then the UI and data just run off the admin maintained data. Again, future-future plan.

src/services/enums/reasons.ts
thewatermethod:
```js
const findAll = async (
  model: ModelType, data: { validFor: typeof ENTITY_TYPE[keyof typeof ENTITY_TYPE] },
) => genericEnum.findAll(model, data);
```

etc

package.json
thewatermethod: I was referring to this: https://github.com/wadackel/dot-wild/issues/18

src/services/enums/reasons.ts
thewatermethod: discussed this on a call - clarified that what I'd really be after is no middleman function at all, IE we just call genericEnum.findAll where appropriate. Also stressed that it's a personal preference thing rather than a "make this change right now" kind of suggestion.

PR #1531, opened 2023-05-10
[TTAHUB-1713] Better handle prompts on multirecipient reports, add resource bug workaround

thewatermethod:
## Description of change
There are three changes here, intended to resolve the user problem detailed in the Jira ticket.

1) As discussed in the original requirements, the front end should not be sending prompts to the back end if the report is a multi-recipient report. Likely, this particular bug is caused by switching the number of recipients after setting prompts, but we are now handling that case. The front end now formats the goals appropriately each time they are sent to the backend, to remove the prompts where necessary.

2) Upon reflection, there is no need at this time for the back end to validate the presence of the prompts. Given the disparate amount of states prompts & goals can be in, I've removed the validation from the backend that checks the required key. This is checked on the frontend. The max selections check remains in place.

3) There was an additional, unrelated bug due to resources. I've added a comment but I am fixing it with an if statement for now to unblock any users affected. If this PR is approved, I will add a Jira to dig into why that issue is occuring.

## How to test
- There is a db copy in the keybase, ```production.2023.05.08.sql```.
- With this database loaded, be (or impersonate) Ann Williams (id 250)
- visit http://localhost:3000/activity-reports/30857/goals-objectives
- save the report

It should save without issue.

Ensure goals & objectives still save correctly.

## Issue(s)
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1714

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/resource.js
+++ b/src/services/resource.js
@@ -587,8 +587,6 @@ const genericProcessEntityForResources = async (
   resourceIds,
   ignoreDestroy = false,
 ) => {
-  console.log({ resourceTableModel });
-
   // Either used the current resource data from the entity passed in or look it up.
   const currentResources = entity[resourceTableAs]
     ? entity[resourceTableAs]
--- a/src/services/resource.js
+++ b/src/services/resource.js
@@ -354,13 +354,24 @@ const filterResourcesForSync = (
         };
       }
 
+      const resourceSourceFields = (() => {
+        // I am seeing this occaisionally come through as a string, like a SQL array literal
+        // ex: '{field1,field2,field3}'
+        // todo - figure out why this is happening and fix it
+        if (resource.sourceFields && Array.isArray(resource.sourceFields)) {
+          return resource.sourceFields;
+        }
+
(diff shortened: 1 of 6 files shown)


src/services/resource.js
```diff
@@ -354,13 +354,24 @@ const filterResourcesForSync = (
         };
       }
 
+      const resourceSourceFields = (() => {
+        // I am seeing this occaisionally come through as a string, like a SQL array literal
```
GarrettEHill, line 358: This is going to end up with bad data in the db, I think the path with the error needs to be found. Just short circuiting it is not a good idea.

nvms (approved):
> @nvms I removed the workaround to start debugging and it seems we didn't need it to fix Ann's problem - can you just give me a sanity/double check when you have a free moment?

tested again and am no longer seeing the resources error. it looks like the AR correctly saves without error on autosave

PR #1533, opened 2023-05-11
[TTAHUB-1680] Single or multiple recipients filter

AdamAdHocTeam:
## Description of change

Add filter to allow viewing reports with single or multiple recipients.

**NOTE:** Patrice says we don't need to worry about OE reports for this filter. She also agreed to a name change for the filter of 'Single or multiple recipients'.

## How to test

**Test on:** AR Landing, Regional Dashboard, Resource Dashboard and Goals (not RTR).
- Create filter to show single reports and verify. 
- Create filter to show multiple and verify.

We should count recipients with the same UEI number but different names as the same recipient (single).

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1680

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/scopes/activityReport/singleOrMultiRecipient.js
```diff
@@ -0,0 +1,30 @@
...
+  ON "Grants"."recipientId" = "Recipients"."id"
+  INNER JOIN "ActivityRecipients" "ActivityRecipients"
+    ON "ActivityRecipients"."grantId" = "Grants"."id"
+  GROUP BY "ActivityRecipients"."activityReportId"
+  HAVING
+    COUNT(DISTINCT CASE WHEN "Recipients"."uei" IS NULL OR  "Recipients"."uei" = ''
+      THEN
+        "Recipients"."name"
+      ELSE "Recipients"."uei"
+      END)
```
GarrettEHill, line 17: This would be about 30% faster
```suggestion
    COUNT(DISTINCT COALESCE(NULLIF("Recipients"."uei",''),"Recipients"."name"))
```

src/scopes/activityReport/singleOrMultiRecipient.js
```diff
@@ -0,0 +1,30 @@
+import { Op } from 'sequelize';
+import { sequelize } from '../../models';
+
+const recipientSql = (singleOrMulti) => (`SELECT
```
GarrettEHill, line 4:
```suggestion
const recipientSql = (singleOrMulti) => (`SELECT DISTINCT
```

src/scopes/activityReport/singleOrMultiRecipient.js
AdamAdHocTeam: Ok this was my initial implementation but I couldn't get it to work guess I was missing the NULLIF.

Good to know.

Thanks.

frontend/src/components/filter/activityReportFilters.js
```diff
@@ -342,6 +344,23 @@ export const targetPopulationsFilter = {
   ),
 };
 
+export const singleOrMultiRecipientsFilter = {
+  id: 'singleOrMultiRecipients',
+  display: 'Single or multiple recipients',
+  conditions: SINGLE_OR_MULTI_RECIPIENT_CONDITIONS,
+  defaultValues: {
+    is: 'Single recipient',
```
nvms, line 352:
```suggestion
    is: 'single-recipient',
```

otherwise:

https://github.com/HHS/Head-Start-TTADP/blob/ab4a5b2792901f42457021bb1b24e5f035af41bf/src/scopes/activityReport/singleOrMultiRecipient.js#L18

`singleOrMulti` here will get the value ["Single recipient"]

nvms, line 352:
```suggestion
    is: 'single-recipient',
```

otherwise:

https://github.com/HHS/Head-Start-TTADP/blob/ab4a5b2792901f42457021bb1b24e5f035af41bf/src/scopes/activityReport/singleOrMultiRecipient.js#L18

`singleOrMulti` here will get the value ["Single recipient"]

nvms (approved):
lgtm

thewatermethod (changes requested):
Screenshot 2023-05-12 at 11 59 16 AM

Filter pill is empty on link sharing

Steps to repro:
1. visit page, add filters
2. paste link in slack message to yourself
3. click on link
4. profit.

the issue is that the query is parsed as an array in this case but is a string when creating a report the value in mapDisplayValue is Array [ "multi-recipients" ] in step 4 above, so mapDisplayValue will have to handle both cases. Probably should add a unit test for just  mapDisplayValue as well

thewatermethod (approved):

PR #1536, opened 2023-05-11
[no ticket] Don't dedupe on prompts via RTR, plug duplicate goal race condition

thewatermethod:
## Description of change
Two issues I turned up when working on a support case
- There was duplicative goal information coming from the frontend that created an observable race condition leading to an extra goal being created. I think this is now visible due to code pushed just a few hours ago, so it's unlikely to have created many extra goals.
- The RTR wasn't accounting for FEI goal responses when reducing the goals for display, so it was possible to trap an FEI goal in a non-editable state. 

## How to test
Follow these steps:
- Create a report for a recipient that has two grants, but only select one. 
- Select the FEI goal (the first time for this recipient). Fill out the rest of the form, including the root causes.
- Set yourself as an approver, go back to the activity summary and select the second grant for that same recipient.
- Approve the report, and go to the RTR for that recipient. There should be two entries on the RTR, one for each goal you just created (they aren't condensed because they have different prompt responses.) There should be no extra goals created.

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/services/recipient.js
```diff
@@ -410,6 +412,29 @@ export async function getGoalsByActivityRecipient(
...
+          'ordinal',
+          'title',
+          'prompt',
+          'hint',
+          'fieldType',
+          'options',
+          'validations',
+        ],
+        required: false,
+        include: [
```
GarrettEHill, line 429: Instead of including the GoalFieldResponce use the through attribute on the GoalTemplateFieldPrompt. This way it will only match a single value instead of the response for every goal.

GarrettEHill: In this method the whole included would be replaced with the following
through: { attributes:['response'] ],

The required is implied as to get to the prompt it already have to link through the response

thewatermethod: thank you

AdamAdHocTeam (commented):
I'm not sure if this is a bug but if I select a FEI goal used on another report the root causes are already selected (I tried clearing my local storage).

thewatermethod:
> I'm not sure if this is a bug but if I select a FEI goal used on another report the root causes are already selected (I tried clearing my local storage)

Sorry, I'm not following - what are the steps to repro? You create a report, approve it, and then you create another report with different grants and the root causes are there?

AdamAdHocTeam:
> > I'm not sure if this is a bug but if I select a FEI goal used on another report the root causes are already selected (I tried clearing my local storage)
> 
> Sorry, I'm not following - what are the steps to repro? You create a report, approve it, and then you create another report with different grants and the root causes are there?

I think this will get you the issue.

1. Create a report with the same recipient twice as above steps but DON'T approve.
2. Create another report with the same recipient you added first above.
3. Add the same FEI goal notice the two root causes are already selected.

AdamAdHocTeam (approved):
Overall looks good. Might want one more pair of eyes tho.

AdamAdHocTeam (approved):
Overall looks good. Might want one more pair of eyes tho.

thewatermethod:
@kryswisnaskas 

> In the test scenario above, the goal on the RTR for the second grant has the root causes editable. Assuming that's what we want.

The scenario in the PR description?

> Once a user adds a second grant to a report where the FEI goal had its root causes selected, the root causes are no longer visible in the read-only box. As a result, the approver can't see the root causes and is approving something they can't see. Is that planned?

It's understood by me that saving the goal saves the root causes. I suppose we could clear out the root causes for the first grant in this scenario. Or keep them on the AR goal model only until the report is approved, and then put them at the goal level. This could lead to weird situations as well though.

> In one scenario, I ended up with duplicated FEI goals on the "Goals & Objectives" page. Is this a bug?

Which scenario? One of the aims of this PR is to **NOT de-duplicate goals** with differing responses so that they are editable on the RTR if they don't have anything selected.

thewatermethod:
Also, @kryswisnaskas, its worth noting that all the questions you ask above are the case in prod right now, except for maybe the third which I am not clear on.

kryswisnaskas:
> > In the test scenario above, the goal on the RTR for the second grant has the root causes editable. Assuming that's what we want.
> 
> The scenario in the PR description?

Correct.



> > Once a user adds a second grant to a report where the FEI goal had its root causes selected, the root causes are no longer visible in the read-only box. As a result, the approver can't see the root causes and is approving something they can't see. Is that planned?
> 
> It's understood by me that saving the goal saves the root causes. I suppose we could clear out the root causes for the first grant in this scenario. Or keep them on the AR goal model only until the report is approved, and then put them at the goal level. This could lead to weird situations as well though.

The jist of this question is that the approver doesn't see the root causes that were entered for the first grant. If we limited the root causes to having them entered on RTR only, this wouldn't be a problem.  Alternatively, we could show them as read-only for the approver to know what they are.







> > In one scenario, I ended up with duplicated FEI goals on the "Goals & Objectives" page. Is this a bug?
> 
> Which scenario? One of the aims of this PR is to **NOT de-duplicate goals** with differing responses so that they are editable on the RTR if they don't have anything selected.

In one scenario, I ended up with duplicated FEI goals on the "Goals & Objectives" page. Is this a bug?

Which scenario? One of the aims of this PR is to NOT de-duplicate goals with differing responses so that they are editable on the RTR if they don't have anything selected.

Unfortunately, I don't have the exact steps for this. I was adding and removing grants and recipients from ARs that had a FEI goal with root causes. Ended up with 2 goals. Screenshot below:

image

thewatermethod:
Closing as the duplicate condition is plugged via #1525 and we can wait until asked to do the other thing, as well as needing it less given goal creation moves to the RTR.

PR #1546, opened 2023-05-17
[TTAHUB-1660] programmatic method to reset db for testing

GarrettEHill:
## Description of change
Some tests are adversely effected by the data left behind by other tests. In a perfect world all test should cleanup after them-selves, but that is not always the case or even feasible. For those cases, we can use this new api to reseed the db back to a clean state via either of the following methods:

-  `import { reseed } from '../../../tests/utils/dbUtils';
    await reseed();`

- `await request(app).get('/testing/reseed');`

Both of these api methods are locked down to only function in a non-production environment by requiring at process.env.NODE_ENV to be either "development" or "test" or process.env.CI to be defined and be called from a file where the stack trace includes /test/ or /testingOnly/ in the path or a file name ending in .test.js or .test.ts

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1660

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


nvms:
looks like the project is failing to start in CI:

```bash
[0] $ node ./build/server/src/index.js
[1] $ yarn --cwd frontend start
[1] $ cross-env DISABLE_ESLINT_PLUGIN=true react-scripts start
[0] 2023-05-17T19:27:08.346Z - error: 	Uncaught exception Cannot find module './testOnly'
[0] Require stack:
[0] - /home/circleci/project/build/server/tests/utils/dbUtils.js
[0] - /home/circleci/project/build/server/src/routes/testingOnly/handlers.js
[0] - /home/circleci/project/build/server/src/routes/testingOnly/index.js
[0] - /home/circleci/project/build/server/src/routes/apiDirectory.js
[0] - /home/circleci/project/build/server/src/app.js
[0] - /home/circleci/project/build/server/src/index.js {}
[1] [HPM] Proxy created: /  -> *********************
[0] error Command failed with exit code 1.
```

GarrettEHill:
> 

I don't see that error, but all the e2e type tests are failing. I don't see why that would be. Do you have any ideas?

nvms:
tests that are trying to start the backend are failing with this error:

```bash
connect ECONNREFUSED 127.0.0.1:8080
```

if you expand the `Start server` accordion in any of these jobs you will see the error in my last comment. the server fails to start because it cannot find the `testOnly` module.

I'll pull this and see if I can find the cause

thewatermethod:
Alternate approach: there is a [sequelize-created helper library](https://github.com/sequelize/umzug) for programatically running these migrations.  

Seeding would be trickier, but you could probably refactor what we have to store the bulk inserted data in separate files (seeders.js) and import it in both the seeder files and in a new seeder function for reuse in bulkCreates

GarrettEHill:
> Alternate approach: there is a [sequelize-created helper library](https://github.com/sequelize/umzug) for programatically running these migrations.
> 
> Seeding would be trickier, but you could probably refactor what we have to store the bulk inserted data in separate files (seeders.js) and import it in both the seeder files and in a new seeder function for reuse in bulkCreates

the syntax of the migration files for [umzug](https://github.com/sequelize/umzug) is different from ours. so I would need to go refactor all the migration files.

GarrettEHill:
@thewatermethod  I figured out the Umzug lib, when I looked at it a few days ago I though it was going to need a full refactor of migration files. turns out only a few tweaks were needed.

tests/utils/dbUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+    DROP SCHEMA public CASCADE;
+    CREATE SCHEMA public;
+  `);
+};
+
+const loadMigrations = async (migrationSet:string): Promise<void> => {
+  const migrationPattern = '*.js'; // File extension pattern for migration files
+  const migrationDir = `src/${migrationSet}/${migrationPattern}`; // path.join('./', migrationSet, migrationPattern);
+
+  console.log(migrationDir);
```
thewatermethod, line 19: Remove?

src/routes/apiDirectory.js
```diff
@@ -71,6 +73,10 @@ router.use('/feeds', feedRouter);
 router.use('/resources', resourcesRouter);
 router.use('/goal-templates', goalTemplatesRouter);
 
+if (isTestingOrCI()) {
+  router.use('/testing', testingRouter);
```
thewatermethod, line 77: **Not** a request for change, but a comment - this route existing, even as locked down as it is (and I acknowledge it would be tricky for some kind of malicious actor to access it), makes me a little nervous. 

Was there a specific use case we needed an API route for that the functions wouldn't be enough for?

src/routes/testingOnly/index.js
```diff
@@ -0,0 +1,10 @@
+import express from 'express';
+import reseedDB from './handlers';
+// import transactionWrapper from '../transactionWrapper';
+import testingOnly from '../../middleware/testingOnlyMiddleware';
+
+const router = express.Router();
+// must run outside of transaction wrapper as some of the action will run in a child process
+router.get('/reseed', testingOnly, reseedDB);
```
thewatermethod, line 8: do we want to put the access middleware on this?

tests/utils/dbUtils.ts
```diff
@@ -0,0 +1,64 @@
+import { exec } from 'child_process';
```
thewatermethod, line 1: You can remove some of these imports as they aren't used any longer

tests/utils/dbUtils.ts
```diff
@@ -0,0 +1,64 @@
...
+        };
+      },
+    },
+    context: db.sequelize.getQueryInterface(),
+    logger: console,
+  });
+
+  try {
+    const migrations = await umzug.up();
+    console.log(`Successfully executed ${migrations.length} migrations.`);
```
thewatermethod, line 40: I think these need to be passed to the auditLogger to show up in certain CI jobs

src/middleware/testingOnlyMiddleware.ts
```diff
@@ -0,0 +1,20 @@
+// import { Request, Response } from 'express';
+
+export const isTestingOrCI = () => {
+  const isLocal = process.env.NODE_ENV === 'development'
+    || process.env.NODE_ENV === 'test';
+  const isCI = process.env.CI !== undefined
+    && process.env.CI !== null;
+
+  return isLocal || isCI;
```
thewatermethod, line 9: This might be fine but it is a little weird that process.env.NODE_ENV = "production" and process.env.CI = "false" (or false) would return true from this function. I'm assuming that the process.env.CI value comes from circleCI and is unpredictable?

thewatermethod:
All looks pretty good!

I was unable to run the dbutils.test.js file in my local docker container, and there's some minor cleanup you might need to do, but other than that this is probably good to go

src/routes/apiDirectory.js
GarrettEHill: Yes the e2e test need this most

GarrettEHill: @thewatermethod, In talking with @nvms we came up with an idea to shift this over to a distinct nodejs app container that would only live in CI and be excluded from ever being deployed to production. as even further protection. Using this method we could add more routes in the future to allow e2e test to better prep the env by making changes to the DB directly without the need to add routes for testing to the main app.

src/routes/testingOnly/index.js
GarrettEHill: it is, thats the testingOnly middleware

src/routes/apiDirectory.js
nvms: thinking out loud here, it sounds like what we should do is...

- create a very small express app that has these destructive routes (e.g. `/reseed`)
- the existing test job in CI spawns a new container that runs this express app
- this express app doesn't get sent to CF because it's included in `.cfignore`

GarrettEHill:
@nvms @thewatermethod @AdamAdHocTeam @hardwarehuman @kryswisnaskas Please review

thewatermethod:
This looks good. Can you update the gitignore for the new folder's test output?

thewatermethod:
I'm ready to approve this once the CI is passing

nvms (approved):
looks good to me

PR #1549, opened 2023-05-19
[TTAHUB-1740] Allow groups to be marked public

thewatermethod:
## Description of change
Adds the ability to make groups public. A public group is viewable by anyone with permissions in the same region as the grants in the group.

## How to test
Confirm the following:
- You can create a public group (a checkbox has been added to the form, checked by default, indicating that a group is private)
- By impersonating or switching to a different user, you can view that group under the appropriate section, both as a filter option and through the "view" page accessibile via the groups table
- You cannot reuse a group name (case insensitive, checking for leading and trailing whitespace)
- You can edit and delete all your groups as normal 

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1740

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/scopes/activityReport/group.ts
```diff
@@ -2,6 +2,16 @@ import { Op, WhereOptions } from 'sequelize';
 import { sequelize } from '../../models';
 import { idClause } from '../utils';
 
+const constructLiteral = (query: string[], userId: number): string => {
+  const where = idClause(query);
+  return sequelize.literal(`(
+    SELECT "activityReportId" FROM "ActivityRecipients" WHERE "grantId" IN (
+      SELECT "grantId" FROM "GroupGrants" WHERE "groupId" IN (
+        SELECT "id" FROM "Groups" WHERE "id" IN (${where}) AND "userId" = ${userId} OR "isPublic" = true)
+    )
```
kryswisnaskas, line 11: Because we are skipping the models, we are also not dealing with any validations. I am assuming that's not going to lead to innacurate results, e.g. activity reports that are deleted, but still present in the "ActivityRecipients".

kryswisnaskas (commented):
Nice.

When running this locally with a fresh copy of the db (/bin/load-test-db), I got an error on creating an objective for the FEI goal.
Also, one time I saw a deleted group as an option in the filter dropdown, although I am unable to reproduce now.

src/scopes/activityReport/group.ts
thewatermethod: The scopes aren't applied without the context of additional logic. As is true for all the existing scopes, we here rely on the individual handlers that they are passed into to apply additional filtering, for example, approved reports only.

src/migrations/20230519174017-add-unique-group-name.js
```diff
@@ -0,0 +1,55 @@
...
+      await queryInterface.sequelize.query(
+        `SELECT
+              set_config('audit.loggedUser', '${loggedUser}', TRUE) as "loggedUser",
+              set_config('audit.transactionId', NULL, TRUE) as "transactionId",
+              set_config('audit.sessionSig', '${sessionSig}', TRUE) as "sessionSig",
+              set_config('audit.auditDescriptor', '${auditDescriptor}', TRUE) as "auditDescriptor";`,
+        { transaction },
+      );
+
+      await queryInterface.sequelize.query(
```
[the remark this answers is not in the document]
thewatermethod: [I updated the migration ](https://github.com/HHS/Head-Start-TTADP/pull/1549/commits/8217f22b551c94ffad4dec799a84c5459c97501c)

thewatermethod:
@kryswisnaskas 

> Also, one time I saw a deleted group as an option in the filter dropdown, although I am unable to reproduce now.

I changed the strategy for repopulating groups after delete, it should no longer be possible [2673964](https://github.com/HHS/Head-Start-TTADP/pull/1549/commits/2673964d4ec07abf2e90a75534f7b3b550941145)

kryswisnaskas:
> > When running this locally with a fresh copy of the db (/bin/load-test-db), I got an error on creating an objective for the FEI goal.
> 
> @kryswisnaskas Can you provide some context? I see an error in the console but those are present in main. I was able to add an objective via an activity report and RTR for the FEI goal.

I was going to remove this comment. It appears this is not longer an issue after fixing the @ttahub/common out of sync problem.

frontend/src/components/WidgetCard.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function WidgetCard({
```
kryswisnaskas, line 4: Could this function have comments?

thewatermethod:
> Could we deploy this to cloud.gov? I am getting an error locally on the dashboard page when creating a group with grants from different regions, e.g 3,4,7,8,9 and then accessing the regional dashboard as a user with permissions only in region 4.

Sure. I attempted to duplicate locally but didn't get any errors like you describe. It's deploying to sandbox now.

> Btw, we are expecting the filter to only bring in results for one of the grants in the group in the case above, right?

Yes.

frontend/src/pages/AccountManagement/MyGroups.js
```diff
@@ -25,6 +29,12 @@ const mapRecipients = (recipients) => recipients.map((recipient) => ({
   options: mapGrants(recipient.grants),
 }));
 
+export const GROUP_FIELD_NAMES = {
+  NAME: 'new-group-name',
+  RECIPIENTS: 'select-recipients-new-group',
+  IS_PRIVATE: 'is-private',
```
AdamAdHocTeam, line 35: Should this have a bool value?

thewatermethod: These are the HTML names of the inputs (name="is-private")

src/routes/groups/handlers.ts
```diff
@@ -74,14 +107,42 @@ export async function updateGroup(req: Request, res: Response) {
...
+              [Op.not]: groupId,
+            },
+          },
+        ],
       },
       attribtes: ['userId', 'id'],
     });
 
-    const policy = new GroupPolicy({ id: userId, permissions: [] }, [], existingGroup);
+    // there can only be one
```
AdamAdHocTeam, line 129: ⚔️

src/routes/groups/handlers.ts
```diff
@@ -74,14 +107,42 @@ export async function updateGroup(req: Request, res: Response) {
...
+    const existingGroups = await Group.findAll({
       where: {
-        id: groupId,
+        [Op.or]: [
+          {
+            id: groupId,
+          },
+          {
+            name: {
+              [Op.iLike]: req.body.name.trim(),
```
AdamAdHocTeam, line 118: Should this a like match?

AdamAdHocTeam (approved):
Overall this looks good. I was able to create groups with mixed region grants. Then impersonate a user to view these reports and grants (only seeing what I have access to). I was also able to create and modify groups. While switching between private and public settings.

* Please be sure to investigate the case @kryswisnaskas found before merging.

PR #1553, opened 2023-05-23
[NO-TICKET] fix weird error.stack bug, add random test coverage

nvms:
## Description of change

was looking through the logs of an unrelated build and saw this:

```bash
Unexpected error in catch block - TypeError: Cannot read properties of undefined (reading 'stack')
```

tracked it down to this. error might not always be an object with a `stack` property, so we shouldn't assume that it is. when `stack` isn't there, this throws inside of the `catch` block.

## How to test

should not see the above error in CI (backend test, under `Start server` job)

## Issue(s)

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


AdamAdHocTeam (commented):
Two things:

1. I still see the error in the BE test logs (maybe another location).
2. Is it safe the output the whole error to cmd line when stack is null. Or does it makes sense to output the message only?

src/lib/apiErrorHandler.js
```diff
@@ -35,7 +36,8 @@ export const handleError = async (req, res, error, logContext) => {
   if (error instanceof Sequelize.Error) {
     await handleSequelizeError(req, res, error, logContext);
   } else {
-    logger.error(`${logContext.namespace} - UNEXPECTED ERROR - ${error.stack}`);
+    const err = error?.stack || error?.message || error;
```
thewatermethod, line 39: I don't know if you want to kill the if statement and just log the whole error here as well?

nvms: ah yep, thanks for catching, missed this one (b9b320a)

src/lib/cron.js
```diff
@@ -28,7 +28,7 @@ const runJob = () => {
     return updateGrantsRecipients();
   } catch (error) {
     auditLogger.error(`Error processing HSES file: ${error}`);
-    logger.error(error.stack);
+    logger.error(error);
```
GarrettEHill, line 31: what not just make auditLogger.error call logger.error?

GarrettEHill (approved):

PR #1568, opened 2023-06-02
[TTAHUB-1654] Get resource metadata

AdamAdHocTeam:
## Description of change

Updated: 
- If ECLKC get meta data, if title fails then scrape it.
- If non-ECLKC scape title.

If we fail retry.

## How to test

- Review the code.
- Make sure all tests pass.
- Add a ECLKC and verify the title and metadata column is populated in the DB.
- Add a Non-ECLKC resource and verify that the title is still populated in the DB.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1654

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/lib/resource.js
GarrettEHill, line 1: I dont know if I am just missing something, but how to do handle the urls that point directly to a file not a rendered page?

src/migrations/20230602174017-add-resource-metadata-column.js
```diff
@@ -0,0 +1,50 @@
...
+                set_config('audit.sessionSig', '${sessionSig}', TRUE) as "sessionSig",
+                set_config('audit.auditDescriptor', '${auditDescriptor}', TRUE) as "auditDescriptor";`,
+        { transaction },
+      );
+
+      // Add national centers.
+      await queryInterface.addColumn('Resources', 'nationalCenters', { type: Sequelize.TEXT }, { transaction });
+
+      // Add metadata column.
+      await queryInterface.addColumn('Resources', 'metadata', { type: Sequelize.JSONB }, { transaction });
```
nvms, line 20: blob

src/lib/resource.js
```diff
@@ -1,51 +1,139 @@
...
+
+const getPageScrapeValues = async (resourceUrl, isEclkc) => {
+  // Use axios to get url info.
+  const res = await axios.get(resourceUrl, { maxRedirects: 3 });
+
+  // Scrape Title & National Centers from ECLKC resources.
+  const foundTitle = res.data.match(/<title[^>]*>([^<]+)<\/title>/);
+  // eslint-disable-next-line max-len
+  // Find the meta tag with name="national-centers".
+  const foundNationalCenters = isEclkc ? res.data.match(/<meta[^>]*=*"national-centers"[^>]*content="([^"]+)"[^>]*>/) : null;
```
nvms, line 15: I would suggest using the native `DOMParser` API for this, instead of regular expression matching, e.g.

```javascript
  const parser = new DOMParser();
  const doc = parser.parseFromString(response.data, "text/html");
  
  const updateObj = {};
  
  if (doc.title) {
    const titleToUpdate = doc.title.trim();
    updateObj.title = titleToUpdate;
  }
  
  if (isEclkc) {
    const nationalCentersNode = doc.querySelector("meta[name='national-centers']");
    if (nationalCentersNode && nationalCentersNode.hasAttribute("content")) {
      const nationalCentersToUpdate = nationalCentersNode.getAttribute("content").trim();
      updateObj.nationalCenters = nationalCentersToUpdate;
    }
  }
```

in my opinion this is less verbose, easier to read and more maintainable for those of us who are not regex wizards. note this is untested and may require slight modifications

src/lib/resource.js
```diff
@@ -1,51 +1,139 @@
...
+      // Create the metadata object.
+      const metadataObj = {
+        created,
+        changed,
+        title,
+        fieldTaxonomyNationalCenters,
+        fieldTaxonomyTopic,
+        langcode,
+        fieldContext,
+      };
```
nvms, line 90: we can simplify these assignments quite a bit:

```javascript
    const res = await axios.get(metadataUrl, { maxRedirects: 3 });
    const { metadata } = res.data;

    const {
      created,
      changed,
      title,
      field_taxonomy_national_centers: fieldTaxonomyNationalCenters,
      field_taxonomy_topic: fieldTaxonomyTopic,
      langcode,
      field_context: fieldContext,
    } = metadata;

    const metadataObj = {
      created,
      changed,
      title,
      fieldTaxonomyNationalCenters,
      fieldTaxonomyTopic,
      langcode,
      fieldContext,
    };
```

src/models/resource.js
nvms, line 1: it looks like every line in this file has changed. perhaps a Windows-style line ending thing again, but I'm curious to know what actually changed here. is it just the three new columns, `nationalCenters`, `metadata` and `metadataUpdatedAt`?

src/lib/resource.js
```diff
@@ -1,51 +1,139 @@
...
+    // 2.) Scrape page data (title & resource).
+    const updated = await getPageScrapeValues(resourceUrl, isEclkc);
+    if (updated === 0) {
+      // We want to cause a retry.
+      auditLogger.info(`Resource Queue: Warning, unable to retrieve resource title for resource '${resourceUrl}'.`);
+      return ({ status: httpCodes.NOT_FOUND, data: { url: resourceUrl } });
+    }
+
+    // 3.) Get metadata (if ECLKC resource).
+    await getMetadataValues(resourceId, resourceUrl, isEclkc);
```
thewatermethod, line 126: Feel free to ignore, but I think this would read better if it was like 
```js
if(isEclkc) {
  await getMetadataValues(resourceId, resourceUrl)
}
```

src/lib/resource.js
```diff
@@ -1,51 +1,139 @@
...
+
+      // get created.
+      const { created } = metadata;
+      // get changed.
+      const { changed } = metadata;
+      // get title.
+      const { title } = metadata;
+      // get field_taxonomy_national_centers.
+      // eslint-disable-next-line max-len
+      const fieldTaxonomyNationalCenters = metadata.field_taxonomy_national_centers;
```
thewatermethod, line 73: Screenshot 2023-06-08 at 3 23 25 PM

this is the national center associated with the page, why do we need to also scrape the page? 

Since the title is in the JSON as well, why not simply (for ECLKC pages) try to get the metadata first, and if that fails, THEN try scraping the page?

src/migrations/20230602174017-add-resource-metadata-column.js
```diff
@@ -0,0 +1,50 @@
...
+
+      // Add a lastUpdated metadata field to all resources.
+      await queryInterface.addColumn('Resources', 'metadataUpdatedAt', { type: Sequelize.DATE }, { transaction });
+    });
```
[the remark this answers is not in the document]
AdamAdHocTeam: For multiple entries couldn't we simply add another column (if its even needed)?

src/lib/resource.js
AdamAdHocTeam: Are you saying get the national center name from the URL?

src/lib/resource.js
AdamAdHocTeam: Good point.

src/lib/resource.js
thewatermethod: From the acceptance criteria:

> if possible capture the following elements; ++  "created", "changed", "title", “field_taxonomy_national_centers", "field_taxonomy_topic", "langcode", "field_context" and associate with the URL, goal, Activity Report. 

I'm saying - I don't think there is any need to make 2 requests for an ECLKC page that has a JSON formatted response. The JSON response has everything we need. 

In the JSON response, we have a unique id (769 in my above screenshot) for that taxonomical term, and a unique url with a human readable name. Just save that to the database. Don't bother making a separate request and parsing the document at all in that case.

When or if we do utilize a National Centers table, we can create an eclkc_id column and make the relationship there.

AdamAdHocTeam (commented):
@thewatermethod, @GarrettEHill, @nvms

Based on the feedback I have decided to scale back this change to simply storing the metadata.

We will still scrape the page title for all resources.

Please re-review when you have a moment.

src/lib/resource.js
```diff
@@ -1,51 +1,121 @@
...
+import axios from 'axios';
+import { auditLogger, logger } from '../logger';
+import { Resource } from '../models';
+
+const getPageScrapeValues = async (resourceUrl, isEclkc) => {
+  // Use axios to get url info.
+  const res = await axios.get(resourceUrl, { maxRedirects: 3 });
+
+  // Scrape Title & National Centers from ECLKC resources.
+  const foundTitle = res.data.match(/<title[^>]*>([^<]+)<\/title>/);
```
thewatermethod, line 12: probably update this content

src/lib/resource.js
```diff
@@ -1,51 +1,121 @@
...
+const getResourceMetaDataJob = async (job) => {
+  const {
+    resourceId, resourceUrl,
+  } = job.data;
+
+  try {
+    // 1.) Determine if this is an ECLKC resource.
+    const isEclkc = resourceUrl.includes('eclkc.ohs.acf.hhs.gov');
+
+    // 2.) Scrape page title for all pages.
```
thewatermethod, line 95: if it's ECLKC, why not get the title w/the metadata so we don't have to scrape in that case?

src/lib/resource.js
```diff
@@ -1,51 +1,121 @@
...
+    };
+
+    // Update URL in DB.
+    await Resource.update({
+      metadata: metadataObj,
+      metadataUpdatedAt: new Date(),
+    }, {
+      where: { url: resourceUrl },
+      individualHooks: false,
+    });
```
nvms, line 80: are we okay with any/all of these being potentially undefined? do they need fallback values before the update happens?

most of these property names are self-explanatory, but I'm not sure I know what `langcode` and `fieldContext` are, so maybe the `get langcode.` comment could be expanded to something like `get langcode - <description of what langcode is and why its important that we store it>`

nvms (commented):
- another minor comment
- would like to see the conflict resolved

I think this is mostly looking great, just some polishing is needed

src/lib/resource.js
AdamAdHocTeam: I want to keep the title as is it also triggers a retry.

src/lib/resource.js
AdamAdHocTeam: I don't have a default definition for each field. IE I would assume 'langcode' is language code but how its used I'm not 100% certain. I would think we take the values as is do you seeing it causing issue? We are still figuring out how all this will be used so I would imagine we would check for undefined.

src/lib/resource.js
AdamAdHocTeam: Comment updated.

src/lib/resource.js
AdamAdHocTeam: We also might wind up scraping the national centers in the future.

thewatermethod: So we are going to fetch the title with the metadata and then fetch it again with the scrape? 
Not sure how the retry factors in - it wouldn't need to retry if it already had the title, right?

src/lib/resource.js
nvms: ah okay, so we're just sortof grabbing everything and figuring out later if we need it or not

in that case, is there a reason we don't just grab everything from `metadata`? e.g.

```javascript
await Resource.update({
  metadata: { ...metadata },
}, { ... });
```

or is it large enough that we don't want to store the entire thing?

src/lib/resource.js
thewatermethod: Couple of things:

- When I test this out, it throws an error in the console despite successfully fetching the metadata. Probably a simple fix.
- There are ECLKC pages without a JSON response. The way this function is written, those pages would not have their titles retrieved.
- I think your code also bombs if there is no title but the other metadata is valid (an unlikely scenario)

I created [a PR to this branch](https://github.com/HHS/Head-Start-TTADP/pull/1581) demonstrating how I imagined this would work. Try once to get the metadata, if it fails, then we try the page title, if that fails, we go to the retry loop. 

I also incorporated Jon's idea, which I think it probably correct, that we should just save the entire metadata object, which simplifies the code for the metadata scrape.

nvms (approved):
looks good

docs/logical_data_model.encoded
```diff
@@ -1 +1,5 @@
+<<<<<<< HEAD
```
thewatermethod, line 1: this is a bad merge

PR #1592, opened 2023-06-21
[TTAHUB-1817] Admins should be able to read approved reports

nvms:
## Description of change

In a support case it was brought to my attention that an admin user should always be able to read/view an approved activity report.

`regionId` normally exists on `this.activityReport.regionId`, but only when the scope is constructed with a `raw` `ActivityReport` model instance. If a `not raw` instance is given, then `regionId` is at `this.activityReport.dataValues.regionId`. the helpers that check for `this.activityReport.regionId` have been updated to support both possibilities and tests have been written to verify this behavior.

- [x] admins should be able to read approved activity reports in any region
- [x] `regionId` check should be made more robust

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1817

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/pages/ApprovedActivityReport/__tests__/index.js
+++ b/frontend/src/pages/ApprovedActivityReport/__tests__/index.js
@@ -202,6 +202,8 @@ describe('Activity report print and share view', () => {
       ...report,
       version: null,
     });
+
+    fetchMock.get('/api/activity-reports/5007', 401);
   });
 
   it('renders an activity report in clean view', async () => {
@@ -221,7 +223,7 @@ describe('Activity report print and share view', () => {
   });
 
   it('handles authorization errors', async () => {
-    act(() => renderApprovedActivityReport(4999));
+    act(() => renderApprovedActivityReport(5007));
 
     await waitFor(() => {
       expect(screen.getByText(/sorry, you are not allowed to view this report/i)).toBeInTheDocument();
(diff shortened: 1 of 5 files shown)


thewatermethod:
Screenshot 2023-06-21 at 2 58 10 PM

I still see this message. It looks like we are also checking the user's regions on ```frontend/src/pages/ApprovedActivityReport/index.js```. 

If I understand the issue, it's that an admin may not have read access to all the regions but should be able to view the report anyways?

nvms:
> Screenshot 2023-06-21 at 2 58 10 PM
> 
> I still see this message. It looks like we are also checking the user's regions on `frontend/src/pages/ApprovedActivityReport/index.js`.

🤔 

> If I understand the issue, it's that an admin may not have read access to all the regions but should be able to view the report anyways?

yep, that's exactly how I understand it too

nvms:
> ApprovedActivityReport

https://github.com/HHS/Head-Start-TTADP/blob/45872857f1c21b702e404fb2bf983ebd462280c9/frontend/src/pages/ApprovedActivityReport/index.js#L89-L92

does it make sense to add something like `&& !isAdmin(user)`, or is it okay to just proceed if the response is a 200?

thewatermethod:
> does it make sense to add something like && !isAdmin(user), or is it okay to just proceed if the response is a 200?

IMO, I don't see much value in checking here as long as we're checking on the backend as well.

nvms:
> > does it make sense to add something like && !isAdmin(user), or is it okay to just proceed if the response is a 200?
> 
> IMO, I don't see much value in checking here as long as we're checking on the backend as well.

makes sense

14bb350876615261c4e8815e44a7faeaa6207e91

this seem OK? `fetchReport` should throw if its outside the 200 range, so if it's an `HTTPError` in the 400 range we can set `notAuthorized` then.

GarrettEHill (approved):

PR #1602, opened 2023-06-29
[TTAHUB-1834] Fix gap in LDM test

GarrettEHill:
## Description of change
There is a gap in the test for the generated file if there is an issue with what is checked-in but the generated file is fine.
This closes the gap

## How to test
Run the test after modifying, a model or the DB structure. If the modification is not checked-in the error will fail with the delta of the file.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1834

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


AdamAdHocTeam (commented):
@GarrettEHill  I might be doing something wrong here but I am unable to get the test to pass locally. I have tried restoring a fresh db, manually modifying the files, also completely removing them and rerunning db:migrate.

nvms:
there is nothing in the description about the migration test that this PR adds, only LDM-related things. was it meant to be added to this PR?

nvms:
we talked about this scenario on slack, but I'm also leaving it here:

```
1) pull this repository
2) drop schema, yarn docker:db:migrate && yarn docker:db:seed
3) make change to `src/models/user.js` (make a column nullable where it wasn’t before)
4) run “Logical Data Model” test, it passes
```
I thought it passed maybe due to the order in which I did this, so I:
```
1) drop and migrate/seed again (user model still has the column change at this point)
2) run the test, it still passes
```

GarrettEHill:
> we talked about this scenario on slack, but I'm also leaving it here:
> 
> ```
> 1) pull this repository
> 2) drop schema, yarn docker:db:migrate && yarn docker:db:seed
> 3) make change to `src/models/user.js` (make a column nullable where it wasn’t before)
> 4) run “Logical Data Model” test, it passes
> ```
> 
> I thought it passed maybe due to the order in which I did this, so I:
> 
> ```
> 1) drop and migrate/seed again (user model still has the column change at this point)
> 2) run the test, it still passes
> ```

nullability of a column was not checked for alignment. it is now....

src/models/activityReport.js
```diff
@@ -146,6 +146,8 @@ export default (sequelize, DataTypes) => {
     },
     version: {
       type: DataTypes.INTEGER,
+      allowNull: false,
+      defaultValue: 1,
```
thewatermethod, line 150: Default should be 2

src/models/helpers/flowControl.ts
```diff
@@ -0,0 +1,61 @@
...
+  arg:
+  string
+  | { name: string, prefix?: boolean, suffix?: boolean }
+  | Array<
+  string
+  | { name: string, prefix?: boolean, suffix?: boolean }
+  >,
+  callingFunctionName?: string | null,
+): boolean {
+  const functionName = callingFunctionName || (new Error().stack || '').split('\n')[2].trim().match(/at (\S+)/)?.[1] || '';
```
thewatermethod, line 11: since this function is only used in the this file and the ```skipif``` function below runs the exact same logic to extract a name if one is not provided, I wonder if you could just use callingFunctionName as its passed in (especially since you test it for truthiness below)

(not that its doing harm, but its extra code that should technically have tests written for each branch)

src/models/activityReport.js
GarrettEHill: OK, I will update the model and the database to reflect that

thewatermethod (approved):
Tested it with a fresh DB, looks to work as described. If there are cases it doesn't catch, we can always iterate

PR #1635, opened 2023-07-19
Shorten websocket interval, display websockets on all pages of form

thewatermethod:
## Description of change

This change applies only to the training report form and the session form. Because of the single column storage approach, multiple people working on the form at the same time is worse than on the activity report.

This change:
- Shortens the websocket interval to 10 seconds from 30
- Publishes the warning if any user is inside ANY page of the form, rather than just the same page

## How to test

Websocket event still fires, behaves as above (you can test solo by removing the logic that excludes the active user from displaying in the SocketAlert component)

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/hooks/useSocket.js
+++ b/frontend/src/hooks/useSocket.js
@@ -8,6 +8,8 @@ import {
 } from 'react';
 import useInterval from '@use-it/interval';
 
+const THIRTY_SECONDS = 30000;
+
 const WS_URL = process.env.REACT_APP_WEBSOCKET_URL || '';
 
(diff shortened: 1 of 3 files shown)


GarrettEHill (approved):
do we need a similar change for the other forms in the site? AR, RTR-goal

thewatermethod:
The AR is the only other form that has it. It's less potentially catastrophic there because only the fields that change and sent up and passed back to the API. In the case of the session and event forms, the one "data" field contains basically the entire form so there is a greater potential for lost work.

PR #1672, opened 2023-08-16
[TTAHUB-1990] Goal similarity API

nvms:
## Description of change

This is almost entirely @kcirtapfromspace's code from his NLP PR: https://github.com/HHS/Head-Start-TTADP/pull/1561

There are enough changes made that I think this warranted a different branch & PR, rather than compounding the existing branch & PR with commits that fundamentally change it.

#### What didn't change:

- The model being used.
- The similarity calculations for all goals given a recipient ID, and how this calculation is made.
- Most of the CircleCI and CF configuration for this project.
- Most of the docker-compose configuration.

#### What did change/was added:

- The name, from data_science to similarity.
- Configurable `alpha` for computation requests.
- Add two other types of similarity calculations.
  - Given a recipient ID and a goal text, find all other goals for this recipient that are within `alpha` similarity, and return them. This is fast enough that we can probably use this as a typeahead source.
  - Given two strings, just return their similarity score. This is a general-purpose calculation that I'm not sure we have use for at the moment, but may be useful in the future.
- Authentication is now just an API key passed as the `X-API-KEY` header.
- The OpenAPI schema is no longer automatically generated. The API has only one endpoint, so I just wrote it out manually.

#### Thoughts:

Should we cache similarity scores in a table with, e.g., three columns: (goal1, goal2, score) that are invalidated in a hook when the goal name is changed/deleted?

## How to test

- `docker-compose down`
- Check out this branch
- `docker-compose up`

At this point, assuming no issues, the similarity API should be alive and ready at `localhost:9100`. There is a file included at `similarity_api/src/test.http` that covers all possible ways to call the `/compute` endpoint. The pre-written requests reference goal names that exist in the transformed SQL dump on keybase.

I use [this extension](https://marketplace.visualstudio.com/items?itemName=humao.rest-client) for VSCode to run these requests. It puts a little "Send Request" action above each request block:

Screenshot 2023-09-26 at 10 15 51 AM

Or, you can use this file as a reference to craft requests using something like curl or postman.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1990

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


.circleci/config.yml
```diff
@@ -316,6 +364,50 @@ jobs:
...
+    steps:
+      - attach_workspace:
+          at: .
+      - setup_remote_docker:
+          version: 20.10.14
+      - run:
+          name: Syft SBOM
+          environment:
+            SYFT_VERSION: v0.82.0
+            IMAGE_NAME: ghcr.io/kcirtapfromspace/cloudfoundry_circleci:latest
```
thewatermethod, line 378: Does this image require Pat's repo to exist? If so, we may want to fork it and give it to adhoc or HHS

thewatermethod: (not that he'll leave us high and dry)

docker-compose.yml
```diff
@@ -31,12 +31,12 @@ services:
     command: ["--endpoint-url", "http://minio:9000", "s3api", "create-bucket", "--bucket", "$S3_BUCKET"]
     depends_on:
       - minio
-  clamav-rest:
-    image: ajilaag/clamav-rest
-    ports:
-      - "9443:9443"
-    environment:
-      - MAX_FILE_SIZE=30M
```
thewatermethod, line 39: did we move this service somewhere else?

similarity_api/README.md
```diff
@@ -0,0 +1,29 @@
+# Installation
+
+**_You should run this inside of docker._**
+
+For running this outside of docker, you'll want to set up a python environment in which you can install these dependencies. It is, however, more simple to just run this in docker.
+
```
thewatermethod, line 6:
```
You should run this inside of docker!

For running this outside of docker...
```
🤣

thewatermethod:
Everything looks really good, I can appreciate that this was a lot of work (sometimes in poorly documented spaces). 
It might help reviewers if you could fill out the "how to test" section.

(I am assuming I will fire up postman to run some manual tests against the API, once I get the container to start)

nvms:
> Everything looks really good, I can appreciate that this was a lot of work (sometimes in poorly documented spaces). It might help reviewers if you could fill out the "how to test" section.
> 
> (I am assuming I will fire up postman to run some manual tests against the API, once I get the container to start)

You should be able to get the container up and running now and it should be listening on 9100. I filled out the "how to test" section with instructions - something I completely forgot to do before, whoops!

.circleci/config.yml
nvms: Yeah, it does. Forking it is probably a good idea for continuity's sake.

similarity_api/README.md
kcirtapfromspace: good luck 🤷‍♂️

.circleci/config.yml
kcirtapfromspace: Haven't looked  at most of the changes, but this image reference is more aligned with deploying a docker container to cloud.gov. I believe I recreated this functionality with the CI/CD using the docker-compose, similar to how the js app performs its vulnerability scan.

kcirtapfromspace: check out if this is still relevant https://github.com/HHS/Head-Start-TTADP/pull/1561/files#diff-2970c60e47e213e52aa3fb4cfdf0c093e9cd3877576dc990aa046904938f03d0R35-R51

kcirtapfromspace: Here is the SBOM & coverage
https://github.com/HHS/Head-Start-TTADP/pull/1561/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R355-R407

thewatermethod:
@nvms Working perfectly now on my local docker, and that VS code extension was definitely a time saver. Do we have a way of testing it's accessibility to the Node app within Cloud.gov? Or are we assuming that given the deploy working, we are safe to save those detais for a follow-on PR?

My understanding is x-secret-key will be set via the CircleCI interface and the URL to this app will be provided to Node via VCAP, is that correct?

nvms:
> Do we have a way of testing it's accessibility to the Node app within Cloud.gov?

You can change the URL in `test.http` to `https://tta-similarity-api-sandbox.app.cloud.gov/compute` to test the deployed version.

> My understanding is x-secret-key will be set via the CircleCI interface and the URL to this app will be provided to Node via VCAP, is that correct?

Yep! We can define the secret in CI/cloud.gov as `SIMILARITY_API_KEY`. This value is used here:

https://github.com/HHS/Head-Start-TTADP/blob/33e5410ffd6f30d06fa40ee5668521db9fca13bb/similarity_api/src/routes/middleware.py#L12-L19

> ...and the URL to this app will be provided to Node via VCAP, is that correct?

I haven't thought too much about how to provide the URL. We should talk as a team about how best to consume this API. I think that decision might influence the way we approach this. Not really a huge design decision, but I'm curious to hear opinions on which approach we like the most.

```mermaid
graph TD
    A[Frontend] -->|Request| B[Similarity API]
    B -->|Response| A
```

or...

```mermaid
graph TD
    C[Frontend] -->|Request| D[Backend]
    D -->|Request| E[Similarity API]
    E -->|Response| D
    D -->|Response| C
```

I'm not sure that the latter approach really buys us anything in exchange for the added complexity.

thewatermethod:
> I haven't thought too much about how to provide the URL. We should talk as a team about how best to consume this API. I think that decision might influence the way we approach this. Not really a huge design decision, but I'm curious to hear opinions on which approach we like the most.

Agree that we should discuss as a team - I like the simpler approach, but I believe that would expose the x-secret-key in the react source code (which may not matter)

docker-compose.override.yml
```diff
@@ -59,3 +59,35 @@ services:
...
+      - "9100:8080"
+    env_file: .env
+    depends_on:
+      - db
+    environment:
+      - POSTGRES_HOST=postgres_docker
+    volumes:
+      - "./similarity_api/src:/app:rw"
+  owasp_zap_backend:
+    image: owasp/zap2docker-stable:latest
```
AdamAdHocTeam, line 75: Do we always want to pull :latest? Would there ever be a case we don't want to. Probably nothing but just figured I would mention.

kcirtapfromspace: I struggle with this as it's a vulnerability scanner. So in theory you'd want to keep that as current as possible to serve as a canary for changes.  I'm not sure pinning a version will accurately capture potential issues with the code base.

AdamAdHocTeam (approved):
Overall looks good was able to run in Windows, with the minor tweak to getting the env var.

Ran the test cases in test.http and the below pass: 

### given a text_1 and text_2, compute the similarity between them
POST https://tta-similarity-api-sandbox.app.cloud.gov/compute
content-type: application/json
X-API-KEY: secret

{ "text_1": "hey world", "text_2": "hello world" }

docker-compose.override.yml
AdamAdHocTeam: can ignore if its only the scan.

nvms:
> Overall looks good was able to run in Windows, with the minor tweak to getting the env var.
> 
> Ran the test cases in test.http and the below pass:
> 
> ### given a text_1 and text_2, compute the similarity between them
> POST https://tta-similarity-api-sandbox.app.cloud.gov/compute content-type: application/json X-API-KEY: secret
> 
> { "text_1": "hey world", "text_2": "hello world" }

ENV problem fixed here: [1ef152b](https://github.com/HHS/Head-Start-TTADP/pull/1672/commits/1ef152b055c3aeedbf9103b38dc1d359da50a1c3)

.circleci/config.yml
```diff
@@ -16,6 +16,9 @@ executors:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: secretpass
           POSTGRES_DB: ttasmarthub
+  docker-python-executor:
+    docker:
+      - image: cimg/python:3.9.17
```
GarrettEHill, line 21: is there a reason 3.9.17 is used over 3.9.18 or even 3.11.5? What python version is in the build pack, as I think they should match.

GarrettEHill: https://github.com/cloudfoundry/python-buildpack/releases
It looks like the latest python build pack no longer supports 3.9.17. So 3.9.18 or 3.10.13 or 3.11.5 should be used.

I think 3.10.13 looks to be the newest stable version. as 3.11.5 is still in bug fix according to https://devguide.python.org/versions/

docker-compose.override.yml
GarrettEHill: Its also in the override, so its never deployed

similarity_api/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.9.18
```
GarrettEHill, line 1: This does not match other versions of python used. 

https://github.com/cloudfoundry/python-buildpack/releases
It looks like the latest python build pack no longer supports 3.9.17. So 3.9.18 or 3.10.13 or 3.11.5 should be used.

I think 3.10.13 looks to be the newest stable version. as 3.11.5 is still in bug fix according to https://devguide.python.org/versions/

similarity_api/activate-env.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+# This assumes you've already run `pyenv install` in this directory.
+pyenv local 3.9.18
```
GarrettEHill, line 4: This does not match other versions of python used.

https://github.com/cloudfoundry/python-buildpack/releases
It looks like the latest python build pack no longer supports 3.9.17. So 3.9.18 or 3.10.13 or 3.11.5 should be used.

I think 3.10.13 looks to be the newest stable version. as 3.11.5 is still in bug fix according to https://devguide.python.org/versions/

similarity_api/runtime.txt
```diff
@@ -0,0 +1 @@
+python-3.9.18
```
GarrettEHill, line 1: This does not match other versions of python used.

https://github.com/cloudfoundry/python-buildpack/releases
It looks like the latest python build pack no longer supports 3.9.17. So 3.9.18 or 3.10.13 or 3.11.5 should be used.

I think 3.10.13 looks to be the newest stable version. as 3.11.5 is still in bug fix according to https://devguide.python.org/versions/

similarity_api/activate-env.sh
GarrettEHill: could this value be detected from within the env or some other method so it only needs to be set/defined in 1 location. That way we never get a mixed version issue popping up.

similarity_api/runtime.txt
GarrettEHill: could this value be detected from within the env or some other method so it only needs to be set/defined in 1 location. That way we never get a mixed version issue popping up.

similarity_api/src/db/db.py
```diff
@@ -0,0 +1,14 @@
...
+db = SQLAlchemy()
+
+def query_many(query_str, data):
+    try:
+        query_str = text(query_str)
+        result = db.session.execute(query_str, data).all()
+        out = [{column: value for column, value in row._mapping.items()} for row in result]
+        return out
+    except Exception as e:
+        return None
```
GarrettEHill, line 13: can we either log or display the error on stderr so we can know when something is going wrong

.circleci/config.yml
nvms: I bumped from 3.9.17 to 3.9.18 b58ac61 for this exact reason (CF deprecation) but forgot to update the version here as well

reconciled here: 2f1d84d19

similarity_api/Dockerfile
nvms: 2f1d84d19

similarity_api/activate-env.sh
nvms: 2f1d84d19

As far as detecting from ENV - that sounds like a good idea. At build time we'd have to populate the contents of `runtime.txt` with the value? So, do this in CI before deploying to cloud.gov?

similarity_api/runtime.txt
nvms: Answered this above ^.

similarity_api/src/db/db.py
nvms: abe177d43f6a6d50d76beeb6c44d25adad3020b4

kryswisnaskas:
This is great. It's exciting seeing it in Sandbox! 🎉 I was able to issue a curl request and get a score. 👍 

I'm just starting to review this. So far, I have a couple of comments:

We get this in the log stream:

13:53:02.083: [APP/PROC/WEB.0] WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.

https://stackoverflow.com/questions/12269537/is-the-server-bundled-with-flask-safe-to-use-in-production

I believe Patrick had plans to use Gunicorn. There are other choices as well (e.g. mod_wsgi, uWSGI). Could we deploy a production grade server to Sandbox? That would give us a preview of what we could expect in production.


> > I haven't thought too much about how to provide the URL. We should talk as a team about how best to consume this API. I think that decision might influence the way we approach this. Not really a huge design decision, but I'm curious to hear opinions on which approach we like the most.
> 
> Agree that we should discuss as a team - I like the simpler approach, but I believe that would expose the x-secret-key in the react source code (which may not matter)

We presented the second approach to the IPT board, as depicted in our boundary diagram. The initial idea was to have the API function similarly to how we currently use the virus scanning application, with the interaction being limited to internal backend and Python API communications, enhancing security. That doesn't mean we couldn't consider revising this approach if there is a better one and the pros of the first option outweigh the cons.

nvms:
> Could we deploy a production grade server to Sandbox? That would give us a preview of what we could expect in production.

Yep! The app is not started with gunicorn now in sandbox.

> We presented the second approach to the IPT board, as depicted in our boundary diagram. The initial idea was to have the API function similarly to how we currently use the virus scanning application, with the interaction being limited to internal backend and Python API communications, enhancing security. That doesn't mean we couldn't consider revising this approach if there is a better one and the pros of the first option outweigh the cons.

Yeah, I'm now leaning towards this approach as well.

similarity_api/src/settings.py
```diff
@@ -0,0 +1,32 @@
...
+DB_USERNAME = os.environ.get("POSTGRES_USERNAME") or (vcap_credentials and vcap_credentials.get('username'))
+DB_PASSWORD = os.environ.get("POSTGRES_PASSWORD") or (vcap_credentials and vcap_credentials.get('password'))
+DB_HOST = os.environ.get("POSTGRES_HOST") or (vcap_credentials and vcap_credentials.get('host'))
+DB_NAME = os.environ.get("POSTGRES_DB") or (vcap_credentials and vcap_credentials.get('db_name'))
+PORT = (vcap_credentials and vcap_credentials.get('port')) or 5432
+
+print("DB_USERNAME: {}".format(DB_USERNAME))
+print("DB_HOST: {}".format(DB_HOST))
+print("DB_NAME: {}".format(DB_NAME))
+print("PORT: {}".format(PORT))
```
kryswisnaskas, line 27:
```suggestion

```

kryswisnaskas: I believe this can now be removed. This needs to be done for prod and non-prod envs since we don't want these values in the logs.

manifest.yml
```diff
@@ -46,3 +48,21 @@ applications:
...
+    stack: cflinuxfs4
+    health-check-type: process
+    buildpacks:
+      - python_buildpack
+    path: similarity_api/src
+    services:
+      - ttahub-((env))
+    env:
+      NEW_RELIC_APP_NAME: tta-similarity-api-((env))
+      NEW_RELIC_LICENSE_KEY: ((NEW_RELIC_LICENSE_KEY))
```
kryswisnaskas, line 62: We still need to implement the `X-API-KEY` secret to be secret which should be unique per environment, including adding the vars to CircleCI.

kryswisnaskas:
It looks good! 👍 I was able to run more tests including the recipient id.

Couple of outstanding items relate to removing sensitive info from logs and implementing api keys as secrets.

"Should we cache similarity scores in a table with, e.g., three columns: (goal1, goal2, score) that are invalidated in a hook when the goal name is changed/deleted?"

This could be something for a different PR. Good idea to explore.

kryswisnaskas (changes requested):
Marking it as "Request changes" to prevent accidental merge

similarity_api/src/settings.py
nvms: 23b837eb0

manifest.yml
nvms: I will add these to CI and all ☁️.gov envs

nvms: This is done

nvms:
> Marking it as "Request changes" to prevent accidental merge

Thank you - changes have all been addressed, I believe

kryswisnaskas (approved):

PR #1695, opened 2023-08-29
an alternative approach to the cleanup migration

hardwarehuman:
This illustrates some changes. They work, but there's other issues we need to discuss and investigate before we call it done.

### DIFF
--- a/src/migrations/20230828103020-program-personnel-cleanup.js
+++ b/src/migrations/20230828103020-program-personnel-cleanup.js
@@ -24,13 +24,27 @@ module.exports = {
          GROUP BY 1,2,3,4,5,6
          )
          DELETE FROM "ProgramPersonnel"
-         WHERE id NOT IN (SELECT first_insert FROM initial_entries);
+         USING "ProgramPersonnel" pp
+         LEFT JOIN initial_entries
+           ON first_insert = pp.id
+         WHERE "ProgramPersonnel".id = pp.id
+           AND first_insert IS NULL;
          
          /* 2. Make a copy of the audit log before cleaning it up */
-         SET session_replication_role = replica;
-         CREATE TABLE "ZZarchiveZALProgramPersonnel20230828103020" AS SELECT * FROM "ZALProgramPersonnel";
-         SET session_replication_role = DEFAULT ;
+         CREATE TABLE "ZZarchiveZALProgramPersonnel20230828103020" AS SELECT * FROM "ZALProgramPersonnel";`,
+        { transaction },
+      );
 
+      // Disable audit log
+      await queryInterface.sequelize.query(
+        `
+        SELECT "ZAFSetTriggerState"(null, null, null, 'DISABLE');
+        `,
+        { transaction },
+      );
+
+      await queryInterface.sequelize.query(
+        `
          /* 3. Clean out the huge number of records related to deleted dupes */
          DELETE FROM "ZALProgramPersonnel" WHERE data_id NOT IN (SELECT id FROM "ProgramPersonnel");
           `, { transaction });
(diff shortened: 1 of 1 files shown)


AdamAdHocTeam (commented):
It looks like this just uses a CTE instead of a temp table.

The run against the most recent db backup from yesterday was 31 minutes just for the delete.

image

src/migrations/20230828103020-program-personnel-cleanup.js
```diff
@@ -10,35 +10,29 @@ module.exports = {
...
+         /* 1. Remove records we don't want */
+         WITH initial_entries AS (
+         SELECT
+           "firstName",
+           "lastName",
+           "role",
+           "grantId",
+           "programId",
+           active,
+           MIN(id) AS first_insert
```
GarrettEHill, line 22: This is not the best set of fields to match on, I saw some records that only changed email address, while name remained the same. and others where the name changed and email remained the same. I think email needs to be included for comparison.

src/migrations/20230828103020-program-personnel-cleanup.js
```diff
@@ -10,35 +10,29 @@ module.exports = {
...
+         FROM "ProgramPersonnel"
+         GROUP BY 1,2,3,4,5,6
          )
+         DELETE FROM "ProgramPersonnel"
+         WHERE id NOT IN (SELECT first_insert FROM initial_entries);
```
GarrettEHill, line 27: Why are you using the "in select" syntax vs the using syntax

src/migrations/20230828103020-program-personnel-cleanup.js
```diff
@@ -10,35 +10,29 @@ module.exports = {
...
+         
+         /* 2. Make a copy of the audit log before cleaning it up */
+         SET session_replication_role = replica;
+         CREATE TABLE "ZZarchiveZALProgramPersonnel20230828103020" AS SELECT * FROM "ZALProgramPersonnel";
```
GarrettEHill, line 31: This is a real table and have an audit log generated for it. I don't think that is what is intended.

src/migrations/20230828103020-program-personnel-cleanup.js
```diff
@@ -10,35 +10,29 @@ module.exports = {
...
+           active,
+           MIN(id) AS first_insert
+         FROM "ProgramPersonnel"
+         GROUP BY 1,2,3,4,5,6
          )
+         DELETE FROM "ProgramPersonnel"
+         WHERE id NOT IN (SELECT first_insert FROM initial_entries);
+         
+         /* 2. Make a copy of the audit log before cleaning it up */
+         SET session_replication_role = replica;
```
GarrettEHill, line 30: This is not how to turn off audit log

src/migrations/20230828103020-program-personnel-cleanup.js
```diff
@@ -10,35 +10,29 @@ module.exports = {
...
+         CREATE TABLE "ZZarchiveZALProgramPersonnel20230828103020" AS SELECT * FROM "ZALProgramPersonnel";
+         SET session_replication_role = DEFAULT ;
 
-         /* 3. Delete everything we don't want to keep */
-         DELETE FROM "ProgramPersonnel" WHERE id NOT IN (SELECT "idToKeep" FROM "ProgramPersonnelToKeep");
+         /* 3. Clean out the huge number of records related to deleted dupes */
+         DELETE FROM "ZALProgramPersonnel" WHERE data_id NOT IN (SELECT id FROM "ProgramPersonnel");
```
GarrettEHill, line 35: Why are you using the "in select" syntax vs the using syntax

thewatermethod:
Should we close this?

thewatermethod:
Should we close this?

hardwarehuman:
Yeah, this was for discussion not merging, and the outcome of the discussion has been incorporated elsewhere.

PR #1696, opened 2023-08-29
Fix bug with removing recipients

thewatermethod:
## Description of change

We saw this bug in standup yesterday. To duplicate in main: 1)On a new AR, add a recipient 2) pick the FEI goal, no need to fill out any fields 3) click activity summary in the left nav to go back and add a second recipient 4) save and continue, you'll get a 500 error.

This PR fixes it. **I understand that the goal does not stay open**. I'd like to fix this bug and log a ticket to fix that one, as it seems to be something in the backend goal service that I don't want to untangle and write a test for before having this fix out there. 

## How to test

See description

## Issue(s)

No ticket for this

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/pages/ActivityReport/formDataHelpers.js
+++ b/frontend/src/pages/ActivityReport/formDataHelpers.js
@@ -50,8 +50,9 @@ export const findWhatsChanged = (object, base) => {
       let currentlyEditing = false;
 
       accumulator.goals = [
+        (base.goalForEditing || null),
         ...(base.goals || []),
-      ].map((goal) => ({
+      ].filter((g) => g).map((goal) => ({
         ...goal,
         grantIds,
         isActivelyEdited: (() => {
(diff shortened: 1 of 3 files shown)


nvms (approved):
Followed the steps and did not see a 500

👍

frontend/src/pages/ActivityReport/formDataHelpers.js
```diff
@@ -50,8 +50,9 @@ export const findWhatsChanged = (object, base) => {
       let currentlyEditing = false;
 
       accumulator.goals = [
+        (base.goalForEditing || null),
         ...(base.goals || []),
-      ].map((goal) => ({
+      ].filter((g) => g).map((goal) => ({
```
GarrettEHill, line 55: this is just removing nulls, right?

src/models/hooks/activityReportGoalFieldResponse.js
```diff
@@ -32,7 +32,7 @@ const recalculateOnAR = async (
   instance,
 ) => sequelize.models.GoalFieldResponse.update(
   {
-    onAR: Sequelize.literal(`(
+    onAR: Sequelize.literal(`COALESCE ((
```
GarrettEHill, line 35: how is it coming back with null, this works. but it is likely pointing to another issue somewhere else.

thewatermethod: You are correct - the bug was in the other file, on the front end. The open goal was not being included in the data being passed up, which meant that the backend would attempt to delete that goal as part of the cleanup of "saveGoalsForReport." That's why this code was being triggered, as the goal was being deleted so would it's responses. 

I didn't think it would hurt to leave the COALESCE in but we don't need it to fix the bug.

PR #1704, opened 2023-09-05
[TTAHUB-2020] Recipient leadership on the RTR Profile tab

thewatermethod:
## Description of change

Add the "recipient leadership" table to the Profile tab of the Recipient record page

The table should only show:
- CFO
- Director of Head Start
- Director of Early Head Start

I had to update our existing script to parse the HSES data to access the CFO. We won't see CFO when this launches, but we should see it once this is deployed to production.

[The mockup](https://www.figma.com/proto/AKodYuKbKzxPmqtoIYkSuj/Profile-Spotlight?page-id=539%3A57933&type=design&node-id=539-61920&viewport=639%2C-64%2C0.13&t=kSeK4MJQLbMadUrJ-1&scaling=min-zoom&starting-point-node-id=539%3A59921) is here, but the Jira specified a few changes from the mockup.

- The pagination is removed, since we don't have historical data
- The widget is flipped to be on the top of the column
- The link to "grantee staff history" is removed, since we don't have historical data

## How to test

With a recent prod snapshot or data that includes imported ProgramPersonnel, view some recipients on the RTR and observe that they have director data. 

You'll see that multigrant recipients have some deduping going on. ([There is a good example here](/recipient-tta-records/1411/region/2/profile)) There shouldn't be multiple people with the same name AND the same title. It's ok, however, if there is Jim, Director for Early Head Start and a Jim, Director for Head Start. Or either of those and a Jim, CFO.

Make sure that it matches the mockup given the caveats above and that the data reflects what is present in the database and looks correct.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2020

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/lib/updateGrantsRecipients.js
```diff
@@ -71,8 +71,10 @@ async function getProgramPersonnel(grantId, programId, program) {
         lastName,
         suffix: getPersonnelField(currentRole, 'suffix', program),
         title: getPersonnelField(currentRole, 'title', program),
-        email,
-
+        email: getPersonnelField(currentRole, 'email', program),
```
AdamAdHocTeam, line 74: Good catch.

src/routes/recipient/handlers.js
```diff
@@ -107,26 +128,36 @@ export async function searchRecipients(req, res) {
...
   try {
-    const { recipientId, regionId } = req.params;
-    // Check if user has access to this region.
-    const userId = await currentUserId(req, res);
-    const readRegions = await getUserReadRegions(userId);
-    if (!readRegions.includes(parseInt(regionId, 10))) {
-      res.sendStatus(403);
-      return;
-    }
+    const proceedQuestionMark = await checkAccessAndExistence(req, res);
```
AdamAdHocTeam, line 131: proceed ?

AdamAdHocTeam (commented):
When making the window smaller, might want to add some space to the right of the sort.

image

AdamAdHocTeam (commented):
I could be wrong here but doesn't arrow pointing up mean AS (A-Z) seems backwards.

image

image

thewatermethod:
> I could be wrong here but doesn't arrow pointing up mean ASC (A-Z) seems backwards.

Ascending does mean *up* and that is the way it is on the AR table.

thewatermethod:
> When making the window smaller, might want to add some space to the right of the sort.

@AdamAdHocTeam ok, did so

AdamAdHocTeam (approved):
Looks good.

PR #1708, opened 2023-09-08
[Prod] Allow selection of root causes on FEI goals from un-approved reports, correctly identify form fields as required

thewatermethod:
## Description of change

- For a single recipient report, adding a root cause to the FEI goal can be done as before, uninterrupted on an AR.
- Multiple recipient reports now allow the editing of the FEI goal from the RTR as soon as it is added to a report and that report is saved. If you've added an FEI goal to an multi-recipient AR, and you attempt to submit that report before you've added root causes to all the goals, you'll see a message alongside the "incomplete pages" alert telling you to go to the RTR for these recipients
- Update react-select and various markup and language to try and improve transparency to a screen-reader about whether a field is required
- Update FEI goals to use a new template.

## How to test

- Create a single recipient report for a user without the FEI goal. You should be able to set a root cause without issue.
- For a multi-grant report: Attempt to do the same. Confirm that you can't set a root cause on a multi-grant report, that you can't submit until you've set those root causes, and that you can find and set them on the RTR.
- Access the site via a screenreader. I used VoiceOver on Mac, so that's all I 100% vouch for.
- Make sure all required fields in the AR & RTR are announced as required. If an oddball screenreader (the kind you might find in the Chrome webstore) doesn't read off required, let's flag it, but if the appropriate input has the required attribute, the major screen readers will handle it. react-select is handled differently, so we'll need to confirm those work as well.
- FEI goals should have the correct template set (see ticket TTAHUB-2083 for details).

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2053
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1288
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2083

### Every PR

- [x] Meets issue criteria
- [x] Boundary diagram updated
- [x] [Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions
- [x] UI review complete

### Production Deploy

- [x] Staging smoke test completed

### After merge/deploy


kryswisnaskas (approved):
M.erging this week since Patrice is out.

PR #1716, opened 2023-09-13
[TTAHUB-2036] Add TR delete and edit

AdamAdHocTeam:
## Description of change

Add ability for admin to delete or edit TR.

**DELETE:** If a event is suspended or not started it can be deleted by an ADMIN.
**EDIT:** If an even is not completed can change the title and creator (IST).

## How to test

- Make sure you can delete (only as admin) if the report is not started or suspended
- Make sure you can change the title and creator (only as admin)
- Make sure you can still take a TR through the entire process (completed session reports etc.)

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2036

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


frontend/src/pages/TrainingReports/components/EventCard.js
```diff
@@ -11,14 +11,20 @@ import { checkForDate } from '../../../utils';
...
 import SessionCard from './SessionCard';
 import './EventCard.scss';
+import Modal from '../../../components/Modal';
+import isAdmin from '../../../permissions';
 
 function EventCard({
   event,
   onRemoveSession,
+  onDeleteEvent,
+  index,
```
thewatermethod, line 21: I'm sure this is the name I used when I sent you the snippet to get this to work, but do you think we could change this prop to be called "zIndex"?

src/services/event.ts
```diff
@@ -286,6 +298,42 @@ export async function findEventsByRegionId(id: number): Promise<EventShape[] | n
   return findEventHelper({ regionId: id }, true) as Promise<EventShape[]>;
 }
 
+export async function findEventCreators(regionId: number): Promise<EventShape[] | null> {
```
thewatermethod, line 301: I know we talked about this, but I really don't think you need this handler - the handler in users/services.js that gets stored on the frontend as "additionalData" has all the users this retrieves as "collaborators" already, this is literally the exact same query

frontend/src/pages/TrainingReports/index.js
```diff
@@ -171,6 +171,22 @@ export default function TrainingReports({ match }) {
...
+      await deleteEvent(String(eventId));
+
+      // update the UI, exclude the deleted event.
+      const events = displayEvents.map((e) => ({ ...e })).filter((e) => e.id !== eventId);
+
+      // update the events state.
+      setDisplayEvents(events);
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.log(e);
```
thewatermethod, line 186: Do we want to show a UI error in this case?

thewatermethod: (or a loading screen)

frontend/src/pages/TrainingReports/components/EventCard.js
```diff
@@ -80,77 +100,93 @@ function EventCard({
...
-          <p className="usa-prose text-bold margin-y-0">Event organizer</p>
-          <p className="usa-prose margin-y-0">{data.eventOrganizer}</p>
+    <>
+      <Modal
+        modalRef={modalRef}
+        title="Are you sure you want to delete this event?"
+        modalId="remove-objective-modal"
+        onOk={async () => onDeleteEvent(id)}
+        okButtonText="Delete"
+        okButtonAriaLabel="This button will delete the event"
```
thewatermethod, line 110: You can just say "delete" "delete event" here, otherwise it reads off "BUTTON: THIS BUTTON WILL DELETE THE EVENT" or something equally obnoxious

thewatermethod (approved):
Overall looks good, works good

Would definitely change the aria-label before merging, other than that just had a few questions or optional changes

src/services/event.ts
thewatermethod: adding the owner would be a one-liner on the front-end

AdamAdHocTeam: The handler for this handles a use case where the IST (owner) might not come back with the available creators.

Do we still want to change this?

frontend/src/pages/TrainingReports/components/EventCard.js
AdamAdHocTeam: done.

frontend/src/pages/TrainingReports/index.js
AdamAdHocTeam: Displayed error.

frontend/src/pages/TrainingReports/components/EventCard.js
AdamAdHocTeam: Updated to "delete event".

src/services/event.ts
thewatermethod:
```js 
const currentOwner = creators.find((creator) => creator.id === ownerId);

if (!currentOwner) {
     // If the current ownerId is not in the creators array, add it.
    const owner = await userById(ownerId);
    creators.push({ id: owner.id, name: owner.name });
 }
 ```
 
 could be ported to the frontend with one less DB query, right? We already have the user on the frontend so assuming we access the UserContext we'd just check to see if they were an admin, and run this exact logic on the options for the select, something like:
 
 ```js
  const { user } = useContext(UserContext);
  const hasAdminRights = isAdmin(user);
  const { users: { collaborators, pointOfContact, eventCreators } } = additionalData;

  const ownerName = owner && owner.name ? owner.name : '';

  const eventCreatorOptions = !eventCreators
    ? []
    : eventCreators.map((o) => ({ value: o.id, label: o.name }));
 
 const currentOwner = eventCreatorOptions.find((creator) => user.id === creator.value);
 if (!currentOwner) {
     // If the current ownerId is not in the creators array, add it.
    eventCreatorOptions.push({ id: user.id, name: user.name });
 }
 
 
 ```

Comment (thewatermethod) [reply src/services/event.ts:301]: 2 less DB queries, since we don't have to get userbyid again either

File: frontend/src/pages/TrainingReports/components/EventCard.js [hunk 5]
```diff
@@ -80,77 +100,93 @@ function EventCard({
...
-          </p>
-        </div>
-        <div className="ttahub-event-card__event-column ttahub-event-card__event-column__organizer padding-right-3">
-          <p className="usa-prose text-bold margin-y-0">Event organizer</p>
-          <p className="usa-prose margin-y-0">{data.eventOrganizer}</p>
+    <>
+      <Modal
+        modalRef={modalRef}
+        title="Are you sure you want to delete this event?"
+        modalId="remove-objective-modal"
```
thewatermethod, line 107: sorry, just noticed this - each id needs to be distinct (so it should be something like `delete-event-${id}` or whatever

AdamAdHocTeam: modalId updated.

frontend/src/pages/TrainingReportForm/index.js
```diff
@@ -156,7 +156,8 @@ export default function TrainingReportForm({ match }) {
       }
 
       try {
-        const users = await getTrainingReportUsers(eventRegion);
+        const eventId = hookForm.getValues('id');
```
thewatermethod, line 159: Putting "hookForm" inside the useEffect causes this data to be fetched 6 times, I think we might want to refactor that

frontend/src/pages/TrainingReportForm/index.js
```diff
@@ -176,7 +177,8 @@ export default function TrainingReportForm({ match }) {
       }
       try {
         const event = await eventById(trainingReportId);
-        resetFormData(hookForm.reset, event);
+        resetFormData(hookForm.reset,
+          { ...event, data: { ...event.data } });
```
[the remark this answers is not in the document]
AdamAdHocTeam: Yea agree must be an artifact from the rolled back change.

frontend/src/pages/TrainingReportForm/index.js
AdamAdHocTeam: I switched to use trainingReportId this comes in from the route match. I only see two useEffect calls now.

thewatermethod (approved):

PR #1750, opened 2023-10-04
[TTAHUB-2146] Queue rate limiter

AdamAdHocTeam:
## Description of change

The issue this attempts to fix is overloading the queue with retries. Based on @GarrettEHill initial investigation we were restarting worker jobs before the previous job had time to clean up its memory.

I tested this on dev for 6 month's (1515 resources) I didn't see the memory of the app spike above 220 MB. I will need to test this again for a yr once its on staging.

## How to test

- Review both the documentation on BULL (https://optimalbits.github.io/bull/) and the code change
- Make sure you agree with the change and settings.
- Triggering ES events like submitting or deleting a report can also be tested on DEV.

## Issue(s)

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/lib/queue.js
```diff
@@ -19,7 +19,13 @@ const generateRedisConfig = () => {
       tlsEnabled: true,
       // TLS needs to be set to an empty object for redis on cloud.gov
       // eslint-disable-next-line no-empty-pattern
-      redisOpts: { redis: { password, tls: {} } },
+      redisOpts: {
+        redis: { password, tls: {} },
+        limiter: {
+          max: 1000,
+          duration: 5000,
```
GarrettEHill, line 26: We might need a longer max duration if that is 5 sec. Some of the operations we put on the worker is because it is a long task. I think maybe 5 min would be fine.

GarrettEHill (approved):

kryswisnaskas (approved):
👍

thewatermethod:
This changes configuration for all redis connections, not just the queue. Did we test to see if it has any effect on the other things using redis, like the queue or websockets?

AdamAdHocTeam:
> This changes configuration for all redis connections, not just the queue. Did we test to see if it has any effect on the other things using redis, like the cache or websockets?

Are we sure this is the case as its on the constructor for the queue()?

PR #1763, opened 2023-10-12
[TTAHUB-1896] Scrape job clean up and listeners cleanup

AdamAdHocTeam:
## Description of change

This is from @GarrettEHill PR: https://github.com/HHS/Head-Start-TTADP/pull/1640

The idea here is that with the rate limiter in place (https://github.com/HHS/Head-Start-TTADP/pull/1750) we wont need any special handling of queue exceptions.

This includes includes the handling of various resource types and increase the listeners.

## How to test

- Run the resource scrape job for a few moths ensure it updates both titles, meta data, and mime types (new column).
- Add a new resource from the UI make sure the above is also updated (both PDF and URL).
- The worker should show no errors for valid resource URL's.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1896

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/migrations/20231018211202-resource-additional-columns.js
```diff
@@ -0,0 +1,29 @@
...
+        type: Sequelize.INTEGER,
+        allowNull: true,
+      }, { transaction });
+    });
+  },
+  down: async (queryInterface) => {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await prepMigration(queryInterface, transaction, __filename);
+
+      // Remove grantee name field to grants.
```
thewatermethod, line 24:
```suggestion
```

src/lib/resource.js
```diff
@@ -1,74 +1,393 @@
...
+};
+
+/**
+ * Retrieves metadata values from a JSON resource.
+ *
+ * @param {string} url - The URL of the resource.
+ * @returns {Promise<Object>} - A promise that resolves to an object containing the metadata,
+ * status code, and MIME type of the resource.
+ */
+const getMetadataValuesFormJson = async (url) => {
```
thewatermethod, line 109:
```suggestion
const getMetadataValuesFromJson = async (url) => {

```

src/lib/resource.js
```diff
@@ -1,74 +1,393 @@
...
+  // eslint-disable-next-line no-useless-escape
+  /<meta[ \t]+(?:name|property)="([^"]*)"[ \t]+content="([^"]*)"[ \t]?(?:[\/])?>/gi,
+];
+/**
+ * Retrieves metadata values from an HTML page.
+ *
+ * @param {string} url - The URL of the HTML page.
+ * @returns {Promise<object>} - An object containing the metadata, status code, and MIME type.
+ */
+const getMetadataValuesFormHtml = async (url) => {
```
thewatermethod, line 172:
```suggestion
const getMetadataValuesFromHtml = async (url) => {

```

src/lib/resource.js
```diff
@@ -78,37 +397,50 @@ const getResourceMetaDataJob = async (job) => {
     // Determine if this is an ECLKC resource.
     const isEclkc = resourceUrl.includes('eclkc.ohs.acf.hhs.gov');
 
-    let updatedCount;
-    let titleRetrieved = false;
+    let statusCode;
+    let mimeType;
+    let title = null;
```
thewatermethod, line 402: title is unused

src/lib/resource.js
```diff
@@ -1,74 +1,393 @@
...
+  let statusCode; // Variable to store the status code of the resource.
+  let metadata; // Variable to store the metadata of the resource.
+  let title = null; // Variable to store the title of the resource, initialized as null.
+  let mimeType; // Variable to store the MIME type of the resource.
+
+  try {
+    const [
+      fromJson,
+      fromHtml,
+    ] = await Promise.allSettled([
```
thewatermethod, line 255: why would we need to get both?

deployment_config/sandbox_vars.yml
```diff
@@ -1,7 +1,7 @@
 env: sandbox
 web_instances: 1
 web_memory: 2GB
-worker_instances: 1
+worker_instances: 2
```
thewatermethod, line 4: is this permanent?

src/lib/queue.js
```diff
@@ -61,6 +61,19 @@ const {
...
+export async function increaseListeners(queue, num = 1) {
+  const redisClient = queue.client;
+  const maxListeners = redisClient.getMaxListeners();
+  const currentCounts = queue.eventNames().reduce((counts, eventName) => ({
+    ...counts,
+    [eventName]: queue.listenerCount(eventName),
+  }), {});
+  const totalCount = Object.values(currentCounts).reduce((acc, count) => acc + count, 0);
+  if (totalCount + num > maxListeners) {
+    redisClient.setMaxListeners(Math.max(totalCount + num, maxListeners));
```
thewatermethod, line 73: Does it make sense to just increase the max listeners to an arbitrary yet realistic number instead of just scaling this infinitely? This would hide exactly the kind of mistake that max listeners is supposed to highlight

deployment_config/sandbox_vars.yml
AdamAdHocTeam: no

src/lib/queue.js
AdamAdHocTeam: @GarrettEHill can you comment on this?

src/lib/resource.js
AdamAdHocTeam: @GarrettEHill do we want both here?

src/lib/queue.js
GarrettEHill: > MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 ready listeners added to [Redis]. Use emitter.setMaxListeners() to increase limit
 Without this we are getting this warning. So we need something to exstend the value

GarrettEHill: We would still likely need something to check for runnaway listeners though

thewatermethod: The warning is not saying redis can't handle anymore listeners, the warning is saying "hey we noticed you are adding a lot of listeners, is something wrong bud?" 

So it doesn't make sense to scale that number up forever, is what I'm saying, since all it would do is hide a potentially valuable warning.

thewatermethod: I suppose that this would increase the max listeners we expect, and any number over that would mean a memory leak or something else fishy. 

I wonder if we might want to include a maximum overall number before setting the max listeners.

src/lib/queue.js
```diff
@@ -61,6 +61,19 @@ const {
 
 export { generateRedisConfig };
 
+export async function increaseListeners(queue, num = 1) {
+  const redisClient = queue.client;
+  const maxListeners = redisClient.getMaxListeners();
```
thewatermethod, line 66:
```suggestion
  const maxListeners = Math.min(redisClient.getMaxListeners(), LISTENER_CAP);
```

src/lib/queue.js
AdamAdHocTeam: Added cap.

src/lib/queue.js
AdamAdHocTeam: applied suggestion.

src/lib/resource.js
AdamAdHocTeam: Spoke with @GarrettEHill we get both here to maximize chances for success in retrieval.  We then combine the two results via spread. The HTML values will only pull in if they are not blank. So we shouldn't have to worry about a case of JSON title being overwritten by HTML.

thewatermethod: Double tapping seems to like it would be a great way to get blocked by a rate limiter. The concern is making two requests where one would do, not that one would overwrite the other. We should only fetch HTML if we don't have JSON.

src/lib/queue.js
```diff
@@ -61,6 +61,21 @@ const {
...
+  const currentCounts = queue.eventNames().reduce((counts, eventName) => ({
+    ...counts,
+    [eventName]: queue.listenerCount(eventName),
+  }), {});
+  const totalCount = Object.values(currentCounts).reduce((acc, count) => acc + count, 0);
+  const newListenerCount = totalCount + num;
+  if (newListenerCount > maxListeners) {
+    redisClient.setMaxListeners(newListenerCount);
+  }
+}
```
thewatermethod, line 77: This doesn't do what I suggested. I want the listeners to stop scaling at 100. If we leave it like this, they will still scale infinitely.

thewatermethod: In fact, if they ever get over 100, "setMaxListeners" will be called every time no matter what, since written this way, maxListeners is the lesser of the current max listeners or 100.

src/lib/queue.js
thewatermethod: See my comment elsewhere, I put the Math.min in the wrong place.

src/lib/queue.js
AdamAdHocTeam: I'm having trouble following your comments lets talk in the AM.

GarrettEHill: Was this resolved, should it throw an error

GarrettEHill:
What is this waiting on?

GarrettEHill (approved):

src/lib/queue.js
AdamAdHocTeam: This was fixed.

thewatermethod (approved):

PR #1769, opened 2023-10-17
[TTAHUB-2171] Table changes for merge goals

thewatermethod:
## Description of change

Modify goals table and model to accommodate the incoming "merge goals" feature.

## How to test

Confirm changes make sense

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2171

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/models/goal.js
```diff
@@ -93,6 +101,16 @@ export default (sequelize, DataTypes) => {
     previousStatus: {
       type: DataTypes.TEXT,
     },
+    mergedIntoId: {
```
GarrettEHill, line 104: I know this is for merging goals, but in effect it directly correlates to the mapsTo convention we have been using recently. Personally I would prefer that naming convention.

thewatermethod: makes sense, will change

src/models/goal.js
```diff
@@ -43,6 +43,14 @@ export default (sequelize, DataTypes) => {
...
         as: 'resources',
       });
+      Goal.belongsTo(models.Goal, {
+        foreignKey: 'mapsTo',
+        as: 'parentGoal',
+      });
+      Goal.hasMany(models.Goal, {
+        foreignKey: 'mapsTo',
+        as: 'childGoals',
+      });
```
kryswisnaskas, line 53: I've been thinking about the "`mapsTo`" convention, especially in the context of the parentGoal and childGoals associations. To align better with those, I'm leaning towards using `parentGoalId`. It ties in well with the hierarchy and still conveys the idea of an original or source goal in a merge context. What are your thoughts?

thewatermethod: @GarrettEHill has one preference, you have another. 

I think the idea is communicated well enough with either "mapsTo" or "parentGoalId", so I'm not sure it matters to me.

kryswisnaskas: @thewatermethod thanks for your flexibility on this. I appreciate that you are open to either convention.

@GarrettEHill, I get that you've introduced the "mapsTo convention and as the one close to the code, it's meaningful for you to have a convention that feels right. However, considering the potential move towards allowing users to use SQL queries and ensuring clarity both for as and future developers, I'm leaning towards a more descriptive and conventional name like parentGoalId.

I typically don't weigh in much on naming conventions, but this particular instance caught my attention due to its potential wider implications.

Can we discuss this and find a compromise?

thewatermethod: I lean more towards descriptive names in general, but the convention being all over the place is persuasive.

"mapsToGoalId" or "mapsToParentGoalId"?

kryswisnaskas: @thewatermethod, I appreciate the compromise. Given the context and future clarity, I would lean towards mapsToParentGoalId as it clearly describes the parent-child relationship. However, if we believe the parent context is assumed and want a slightly shorter name, mapsToGoalId works as well. Let's finalize on one of these.

thewatermethod: Great, thanks @kryswisnaskas, I updated the name

kryswisnaskas (approved):
Looks good! 👍

PR #1770, opened 2023-10-17
Automate CDI grants links to previous recipients

kryswisnaskas:
## Description of change
HSES sends us information on the previous grant once a grant goes to CDI. Currently, we use the Admin UI to manually associate the CDI grants with recipients. This PR automates this process.

## How to test
Verify tests pass; look at the code.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-814

### Every PR

- [ ] Meets issue criteria
- [n/a] UI review complete

### Production Deploy

### After merge/deploy


src/lib/updateGrantsRecipients.js
```diff
@@ -339,6 +360,14 @@ export async function processFiles(hashSumHex) {
 
       await Promise.all(grantUpdatePromises);
 
+      // Automate CDI linking to preceding recipients
+      const cdiGrantsToLink = await Grant.unscoped().findAll({
+        where: { regionId: 13, endDate: { [Op.gte]: '2021-03-17' } },
```
AdamAdHocTeam, line 365: I might be wrong on this suggestion but can we exclude cdi grants that already have the region id set?

AdamAdHocTeam, line 365: Im trying to think if there is a way to limit the number of records coming back here. But there might not be...

thewatermethod (approved):

src/lib/updateGrantsRecipients.js
```diff
@@ -137,6 +137,27 @@ async function getProgramPersonnel(grantId, programId, program) {
   return programPersonnelArray;
 }
 
+export const updateCDIGrantsWithOldGrantData = async (grantsToUpdate) => {
+  try {
+    const updatePromises = grantsToUpdate.map(async (grant) => {
+      if (grant.oldGrantId) {
+        const oldGrant = await Grant.findByPk(grant.oldGrantId);
```
GarrettEHill, line 144: Why do the find and update in two separate calls to the db. It's good that all the update promises are awaiting together, but it could be even more efficient if each operation had a single db call that could be awaited together at the end

src/lib/updateGrantsRecipients.js
GarrettEHill: Why is there a need to limit it?

AdamAdHocTeam (approved):

src/lib/updateGrantsRecipients.js
AdamAdHocTeam: Simply to limit the amount of unnecessary processing but it might not be possible/needed.

kryswisnaskas: Good thinking, but if a grant is marked with region 13, that means it needs an update, if possible. We could restrict to only the grants that don't have null grantId's, but there are only 3 grants right now.

src/lib/updateGrantsRecipients.js
kryswisnaskas: Yes, it's true that there is a way to perform one trip to the database and slightly optimize performance. I chose this solution, though, since it has the advantage of being straightforward and readable. The performance optimization would be more relevant if there were many grants needing an update. As it stands, between 3/17/21 and 10/16/23, there are only about 110 grants with region 13. That makes it roughly 3.55 grants per month.

PR #1783, opened 2023-10-25
[TTAHUB-2109] Admin form for creating multi-recipient goals

thewatermethod:
## Description of change

An admin form for creating multi-recipient goals and optionally, draft AR.

## How to test

Go to the admin section, the goals section.
Pick a region and fill out the form. You need to have groups in your database for this feature to work. Confirm that the goals and AR are created correctly.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2109

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/migrations/20231025135316-add-to-created-via-enum.js
```diff
@@ -0,0 +1,19 @@
+const { GOAL_CREATED_VIA } = require('../constants');
+const { prepMigration } = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      return Promise.all(Object.values(GOAL_CREATED_VIA).map((method) => queryInterface.sequelize.query(`
```
GarrettEHill, line 10: Could you put this in the migration lib instead, along with the down version. That way we can reuse it for subsequent enum updates.

src/routes/admin/group.ts
```diff
@@ -0,0 +1,35 @@
...
+
+const namespace = 'ADMIN:GROUPS';
+const logContext = { namespace };
+
+/**
+   *
+   * @param {Request} req - request
+   * @param {Response} res - response
+   */
+export async function getGroupsByRegion(req: Request, res: Response) {
```
GarrettEHill, line 16: Groups can span regions, will that cause any issues?

src/services/goals.js
```diff
@@ -2389,3 +2390,166 @@ export async function destroyGoal(goalIds) {
...
 }
+
+export async function createMultiRecipientGoalsFromAdmin(data) {
+  // const {
+  //   region, // string
+  //   group, // string, group ID
+  //   createReport, // bool
+  //   useCuratedGoal, // bool
+  //   creator, // string, user ID
+  //   templateId, // string, template ID
```
GarrettEHill, line 2401: TemplateId is a string?

src/services/goals.js
```diff
@@ -2389,3 +2390,166 @@ export async function destroyGoal(goalIds) {
...
+  }
+
+  let endDate = null;
+
+  if (data.goalDate) {
+    endDate = data.goalDate;
+  }
+
+  const goals = await Promise.all(grantIds.map(async (grantId) => {
+    const goal = await Goal.create({
```
GarrettEHill, line 2474: Please use bulk create with the flag to return the id for each of the goals created. This will be significantly faster.

src/services/goals.js
```diff
@@ -2389,3 +2390,166 @@ export async function destroyGoal(goalIds) {
...
+        2: 'Not started',
+        3: 'Not started',
+        4: 'Not started',
+      },
+    };
+
+    activityReport = await ActivityReport.create(reportData);
+
+    // create activity recipients
+    await Promise.all(grantIds.map(async (grantId) => {
```
GarrettEHill, line 2527: Please use bulk create.

src/services/goals.js
```diff
@@ -2389,3 +2390,166 @@ export async function destroyGoal(goalIds) {
...
+      await ActivityRecipient.create({
+        activityReportId: activityReport.id,
+        grantId,
+      });
+    }));
+
+    // create goals
+    await Promise.all(goals.map(async (goal) => {
+      await ActivityReportGoal.create({
```
GarrettEHill, line 2536: Please use bulk create

src/services/goals.js
```diff
@@ -2389,3 +2390,166 @@ export async function destroyGoal(goalIds) {
...
+      pageState: {
+        1: 'Not started',
+        2: 'Not started',
+        3: 'Not started',
+        4: 'Not started',
+      },
+    };
+
+    activityReport = await ActivityReport.create(reportData);
+
```
GarrettEHill, line 2525: The creates for ActivityRecipients and Activity report goals can be done in the same Promise all

src/services/groups.ts
```diff
@@ -37,6 +38,26 @@ interface GroupResponse {
   isPublic: boolean;
 }
 
+export async function groupsByRegion(region: number): Promise<GroupResponse[]> {
+  return Group.findAll({
```
GarrettEHill, line 42: Please filter the attributes to only those needed

src/services/goals.js
thewatermethod: form data from an input

src/routes/admin/group.ts
thewatermethod: It think it won't (for this application, anyways)

src/migrations/20231025135316-add-to-created-via-enum.js
thewatermethod: I don't have a down, but I will add this to the migration lib

GarrettEHill (approved):

kryswisnaskas:
It generally looks good! 🎉  I was able to create standard goals for groups with recipients that didn't have them yet. It correctly flagged when grants already had the standard goals. I was also able to submit and approve the ARs that it created. 

A couple of other observations/questions:
Is the FEI goal with 1 root cause expected to be non-editable on the RTR?

I also noticed couple of minor things in the UI: 

-  the feedback message after creating a goal for 1 ar says "Created 1 goals". I know this is only the admin UI, but it would be nice to have a singular form in this case.

-  after creating a goal, the user has to navigate away and come back to "Goals" to be able to create again.

thewatermethod:
> the feedback message after creating a goal for 1 ar says "Created 1 goals". I know this is only the admin UI, but it would be nice to have a singular form in this case.

I'm not going to do this, for the simple reason that this is explicitly for multi-recipient goals, and there would be no real case in which we are creating one goal, and so I don't want to have to write a test for a purely hypothetical sitation

thewatermethod:
> after creating a goal, the user has to navigate away and come back to "Goals" to be able to create again.

I will add a route back

kryswisnaskas:
> @kryswisnaskas
> 
> > Is the FEI goal with 1 root cause expected to be non-editable on the RTR?
> 
> Can you explain what you are seeing, how you got there, and what you expect to see?

Here are the steps I took:

- created a group with recipients that don't have FEI goals yet (I have a query to find them, if needed)
- used the Admin UI to create a standard, FEI goal using the group from above
- added only one root cause
- opted to also create an AR
- after the AR was created, the goal displayed on the RTR page, but the root causes were not editable. Unless something has changed, I was expecting to be able to add a second root cause there
- navigated to the newly created AR
- submitted and approved it
- RTR behavior was the same as above

Hope this helps.

kryswisnaskas:
> > the feedback message after creating a goal for 1 ar says "Created 1 goals". I know this is only the admin UI, but it would be nice to have a singular form in this case.
> 
> I'm not going to do this, for the simple reason that this is explicitly for multi-recipient goals, and there would be no real case in which we are creating one goal, and so I don't want to have to write a test for a purely hypothetical situation

Thanks for sharing your thoughts on this. I totally get where you're coming from, especially about prioritizing real-use cases. If we ever see this situation pop up more frequently in the future, we can think about it then. Thanks again for your hard work on this feature!

thewatermethod:
@kryswisnaskas thanks for taking the time to write that up. 

I attempted to reproduce following these steps, taken from yours:

- create an FEI goal for a group, with an AR. Select only one root cause
- no recipients had the FEI goal yet
- Impersonate the event creator, so I had the write permissions in that region
- Go to the AR. Root cause is editable, with one selected
- Go to RTR. Root cause is editable, with one selected

Screenshot 2023-11-01 at 11 22 39 AM

Can you tell me if I missed a step or did something you didn't?

kryswisnaskas:
I reran the test with the correct permissions this time and didn't encounter that error. Perhaps there was something in my environment previously that prevented me from seeing the root causes with read-only permissions.

kryswisnaskas (approved):

PR #1786, opened 2023-10-26
[TTAHUB-2185] FEI Root Cause filter

AdamAdHocTeam:
## Issue(s)

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/scopes/activityReport/feiRootCause.js
+++ b/src/scopes/activityReport/feiRootCause.js
@@ -1,0 +1,21 @@
+import { filterAssociation } from './utils';
+
+const feiResponseSql = `
+WITH unnested_responses AS (
+    SELECT "activityReportGoalId", unnest("response") AS res
+    FROM "ActivityReportGoalFieldResponses"
+)
+SELECT
+  DISTINCT "ActivityReportGoals"."activityReportId"
+FROM "ActivityReportGoals" "ActivityReportGoals"
+INNER JOIN unnested_responses arr
+    ON arr."activityReportGoalId" = "ActivityReportGoals"."id"
+WHERE arr."res"`;
+
+export function withFeiResponse(responses) {
+  return filterAssociation(feiResponseSql, responses, false);
+}
+
+export function withoutFeiResponse(responses) {
+  return filterAssociation(feiResponseSql, responses, true);
+}
--- a/src/scopes/activityReport/index.js
+++ b/src/scopes/activityReport/index.js
@@ -30,6 +30,7 @@ import { withDeliveryMethod, withoutDeliveryMethod } from './deliveryMethod';
 import { withResourceAttachment, withoutResourceAttachment } from './resourceAttachment';
 import { withResourceUrl, withoutResourceUrl } from './resourceUrl';
 import { onlyCollaborators, onlyCreators, bothCollaboratorsAndCreators } from './specialistName';
+import { withFeiResponse, withoutFeiResponse } from './feiRootCause';
 
 export const topicToQuery = {
   reportId: {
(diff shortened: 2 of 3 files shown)


src/scopes/activityReport/feiRootCause.js
```diff
@@ -0,0 +1,21 @@
...
+    FROM "ActivityReportGoalFieldResponses"
+)
+SELECT
+  DISTINCT "ActivityReportGoals"."activityReportId"
+FROM "ActivityReportGoals" "ActivityReportGoals"
+INNER JOIN unnested_responses arr
+    ON arr."activityReportGoalId" = "ActivityReportGoals"."id"
+WHERE arr."res"`;
+
+export function withFeiRootCause(responses) {
```
GarrettEHill, line 15: With the route written generically, why is the scope bespoke to only the root cause, it is was written to be generic then as others get added, it would already be filterable.

src/scopes/activityReport/feiRootCause.js
```diff
@@ -0,0 +1,21 @@
...
+  return filterAssociation(feiResponseSql, responses, false);
+}
+
+export function withoutFeiRootCause(responses) {
```
GarrettEHill, line 19: The way the query is written, the negative match will only match those with something populated that is not what was specified. It will not match those with nothing populated.

src/widgets/topicFrequencyGraph.js
```diff
@@ -209,8 +209,7 @@ export async function topicFrequencyGraphViaGoals(scopes) {
 
   topicsAndParticipants.forEach((goalData) => {
     goalData
-      .get('topics')
-      .forEach(({ topic, reportIds }) => {
+      .get('topics')?.forEach(({ topic, reportIds }) => {
         const topicResponce = topicsResponse
```
GarrettEHill, line 213: How is this related to the other changes

AdamAdHocTeam: Looks like  a bad change will roll back.

AdamAdHocTeam: O wait this is because when you filter on RTR for a specific recipient and no results come back we need to check before accessing.

src/scopes/activityReport/feiRootCause.js
AdamAdHocTeam: This appears to be behavior in other filters like author, collaborators, grant numbers, program specialist, recipient, report text, etc....

src/scopes/activityReport/feiRootCause.js
AdamAdHocTeam: This is only for FEI Root Cause on AR atm.

thewatermethod:
Everything works great. 

I do agree with Garrett that it might be worthwhile to preserve the generic nature of the feature, given how the rest of the conditional fields stuff is written. Especially since it looks like all you'd have to do is turn a couple of things into props on that <FilterRootCauseSelect /> front end component (labelText, goalTemplateName, maybe more), and unless I'm mistaken, all you'd have to do on the backend is change the filter name? Not sure if I'm missing some complexity somewhere.

frontend/src/components/filter/FilterActivityReportGoalResponseSelect.js
```diff
@@ -0,0 +1,41 @@
...
+  }, [setGoalTemplatePrompts, title]);
+
+  const onApplyClick = (selected) => {
+    onApply(selected);
+  };
+  return (
+    <FilterSelect
+      onApply={onApplyClick}
+      inputId={inputId}
+      labelText="Select root cause to filter by"
```
thewatermethod, line 31: should labeltext be a prop as well?

thewatermethod (approved):

PR #1788, opened 2023-10-28
[TTAHUB-2176] SERVICE:ACTIVITY_REPORTS: invalid input value for enum "enum_Objectives_suspendReason"

GarrettEHill:
## Description of change
Add a helper that can validate enums are a valid value before sending to the DB.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2176

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


thewatermethod:
Code looks good, probably need to investigate those test failures a bit before we can review, seem to be directly related to the change

GarrettEHill:
@thewatermethod @nvms @AdamAdHocTeam, please review

src/models/hooks/activityReportObjective.js
```diff
@@ -34,6 +36,10 @@ const recalculateOnAR = async (sequelize, instance, options) => {
   `, { transaction: options.transaction });
 };
 
+const beforeValidate = async (sequelize, instance, options) => {
+  validateChangedOrSetEnums(sequelize, instance);
```
AdamAdHocTeam, line 40: If this throws an exception where do we bubble up to from a hook?

GarrettEHill: the very top, the operation will fail

AdamAdHocTeam (commented):
What are the steps to reproduce the client reported issue?

src/models/hooks/activityReportObjective.js
AdamAdHocTeam: So rollback of the current operation. How does this look on the FE, does the user know what happened in terms of a invalid enum.

GarrettEHill: the issue this is directly solving was empty string was trying to be saved. Which is not valid. This will turn empty string into null and save that if valid.

GarrettEHill: It just also now checks for all other valid values in the same pass. So that the error message is more descriptive.

GarrettEHill:
> What are the steps to reproduce the client reported issue?

I could see the API was called with an empty string, but not how that was called. So I don't know how to reproduce it from the UI

AdamAdHocTeam (approved):
LGTM I was able to save ActivityReportObjecitve with ENUM values.

thewatermethod (approved):
I do wish we knew how this error was happening, this unfortunately doesn't really address whatever that is, but impossible to fix without more information (this may very well gather it)

PR #1789, opened 2023-10-30
[TTAHUB-2179] Goal similarity score caching

nvms:
## Description of change

This PR introduces a table to store cached goal similarity scores. The python app is responsible for nightly scoring. The logic for the scoring is also part of this PR.

Using the cached scores is not part of this PR, because I don't think we know exactly how we want to do that yet.

Also as part of this PR is TTAHUB-2252 which ensures the NLP model is cached by CircleCI.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2179
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2252

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/migrations/20231030150724-sim_score_cache.js
```diff
@@ -0,0 +1,37 @@
...
+        'SimScoreCaches',
+        {
+          id: {
+            type: Sequelize.INTEGER,
+            allowNull: false,
+            primaryKey: true,
+            autoIncrement: true,
+          },
+          recipient_id: { allowNull: false, type: Sequelize.INTEGER },
+          goal1: { allowNull: false, type: Sequelize.INTEGER },
```
thewatermethod, line 21: should these be noted as references in the migration? 

```js
{
   type: Sequelize.INTEGER,
   allowNull: true,
   default: null,
   references: {
      model: {
      tableName: 'Goals',
   },
   key: 'id',
}
 ```

File: src/models/hooks/goal.js [hunk 2]
```diff
@@ -57,6 +57,22 @@ const preventNameChangeWhenOnApprovedAR = (sequelize, instance) => {
   }
 };
 
+const invalidateSimilarityScores = async (sequelize, instance, options) => {
+  const changed = Array.from(instance.changed());
+
+  if (changed.includes('name')) {
+    await sequelize.models.SimScoreCache.destroy({
+      where: {
+        [sequelize.Op.or]: [
```
thewatermethod, line 66: this errors when you try to change a goal (sequelize.Op is ```undefined```), might need ```const { Op }  = require('sequelize');``` in this file

src/migrations/20231030150724-sim_score_cache.js
nvms: yup! thank you

src/models/hooks/goal.js
nvms: 🙏 e02b41a00e0385cc3d788b045cbaf0a9fe0b4750

src/migrations/20231030150724-sim_score_cache.js
nvms: 539b79ae0c62125b04f9073e935387d0f03d7224

thewatermethod:
Was able to connect! Had to use "db" instead of "localhost"

Seeing an error when the job runs, partial stack trace below

```
File "/app/sim/compute.py", line 88, in calculate_batch_similarity
   sentence_embeddings = np.array([nlp(goal).vector for goal in batch])
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (500,) + inhomogeneous part.

File: similarity_api/src/settings.py [hunk 3]
```diff
@@ -25,3 +28,7 @@ def get_postgres_credentials_from_vcap():
 SQLALCHEMY_DATABASE_URI = f"postgresql://{DB_USERNAME}:{DB_PASSWORD}@{DB_HOST}:{PORT}/{DB_NAME}"
 
 API_KEY = os.environ.get("SIMILARITY_API_KEY")
+
+SCHEDULER_JOBSTORES = {
+  'default': SQLAlchemyJobStore(url='sqlite:///jobs.sqlite')
```
thewatermethod, line 33: Should we add this file to .gitignore?

nvms:
> Was able to connect! Had to use "db" instead of "localhost"
> 
> Seeing an error when the job runs, partial stack trace below
> 
> ```
> File "/app/sim/compute.py", line 88, in calculate_batch_similarity
>    sentence_embeddings = np.array([nlp(goal).vector for goal in batch])
> ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (500,) + inhomogeneous part.
> ```



> Was able to connect! Had to use "db" instead of "localhost"
> 
> Seeing an error when the job runs, partial stack trace below
> 
> ```
> File "/app/sim/compute.py", line 88, in calculate_batch_similarity
>    sentence_embeddings = np.array([nlp(goal).vector for goal in batch])
> ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (500,) + inhomogeneous part.
> ```

Well this one sent me down a rabbit hole. The goal causing this issue had an empty string for the `name`. I updated the query to ignore empty goal names: 6076b3adec23b23e1501bdd2aa062233799a50e3

thewatermethod (approved):

src/migrations/20231030150724-sim_score_cache.js
```diff
@@ -0,0 +1,55 @@
...
+} = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      await queryInterface.createTable(
+        'SimScoreCaches',
```
GarrettEHill, line 12: I think this needs to have goal in the name of the table, we may need to have similarity scores for other entity types in the future. Objectives, Uploaded resources, ...

similarity_api/src/app.py
```diff
@@ -1,5 +1,9 @@
+import logging
+
 from application import create_app
 
+logging.basicConfig(level=logging.INFO)
```
GarrettEHill, line 5: Can logging level be set by an ENV and default to this value

GarrettEHill: That way we can turn logging up or down without redeploy

similarity_api/src/application.py
```diff
@@ -1,10 +1,37 @@
...
     app.config.from_pyfile("settings.py")
+
     db.init_app(app)
+
     from routes.create import create_routes
     create_routes(app)
+
+    scheduler.init_app(app)
+    scheduler.start()
+    # scheduler.add_job(func=cache_scores, id='cache', trigger='interval', seconds=10, timezone='EST')
```
GarrettEHill, line 23: Can this line be removed

similarity_api/src/application.py
```diff
@@ -1,10 +1,37 @@
...
+    scheduler.add_job(func=cache_scores, id='cache', trigger='cron', minute='0', hour='3', timezone='GMT', replace_existing=True, max_instances=1)
+
+    atexit.register(lambda: scheduler.shutdown(wait=False))
+
+    # By default, APScheduler logs are a bit noisy.
+    logging.getLogger("apscheduler.scheduler").setLevel(logging.DEBUG)
```
GarrettEHill, line 29: Can logging level be set by an ENV and default to this value

similarity_api/src/routes/create.py
```diff
@@ -1,6 +1,8 @@
 from flask import Flask, jsonify, request, send_from_directory
 from routes.middleware import api_key_header
-from sim.compute import compute_goal_similarities, find_similar_goals, calculate_string_similarity
+from sim.compute import (calculate_string_similarity,
+                         compute_goal_similarities, find_similar_goals)
```
GarrettEHill, line 4: if its going to be multi-line, please make each import on their own line

similarity_api/src/settings.py
GarrettEHill: what is the path to the file? does it need to live across executions? could it be placed in a location that is always destroyed when the container is relaunched

similarity_api/src/sim/compute.py
```diff
@@ -15,17 +18,11 @@ def compute_goal_similarities(recipient_id, alpha):
     FROM "Goals" g
     JOIN "Grants" gr ON g."grantId" = gr."id"
     JOIN "Recipients" r ON gr."recipientId" = r."id"
-    WHERE r."id" = :recipient_id AND g."name" IS NOT NULL;
+    WHERE r."id" = :recipient_id AND NULLIF(TRIM(g."name"), '') IS NOT NULL;
```
GarrettEHill, line 21: Could we add a new line before the AND, and the ON's

similarity_api/src/sim/compute.py
```diff
@@ -34,6 +31,51 @@ def compute_goal_similarities(recipient_id, alpha):
...
+  """
+  Inserts the similarity score into the SimScoreCaches table.
+  """
+  query(
+    """
+    INSERT INTO "SimScoreCaches" (recipient_id, goal1, goal2, score, "createdAt", "updatedAt")
+    SELECT :recipient_id, :goal1, :goal2, :score, :createdAt, :updatedAt
+    WHERE NOT EXISTS (
+      SELECT 1 FROM "SimScoreCaches"
+      WHERE recipient_id = :recipient_id AND goal1 = :goal1 AND goal2 = :goal2
```
GarrettEHill, line 73: can we have a newline before FROM and AND's

similarity_api/src/sim/compute.py
```diff
@@ -101,17 +143,11 @@ def find_similar_goals(recipient_id, goal_name, alpha):
         FROM "Goals" g
         JOIN "Grants" gr ON g."grantId" = gr."id"
         JOIN "Recipients" r ON gr."recipientId" = r."id"
-        WHERE r."id" = :recipient_id AND g."name" IS NOT NULL;
+        WHERE r."id" = :recipient_id AND NULLIF(TRIM(g."name"), '') IS NOT NULL;
```
GarrettEHill, line 146: can we have a newline before ON's and AND

src/migrations/20231030150724-sim_score_cache.js
```diff
@@ -0,0 +1,55 @@
...
+        {
+          id: {
+            type: Sequelize.INTEGER,
+            allowNull: false,
+            primaryKey: true,
+            autoIncrement: true,
+          },
+          recipient_id: { allowNull: false, type: Sequelize.INTEGER },
```
GarrettEHill, line 20: please add the reference section

src/migrations/20231030150724-sim_score_cache.js
```diff
@@ -0,0 +1,55 @@
...
+            type: Sequelize.INTEGER,
+            default: null,
+            references: {
+              model: {
+                tableName: 'Goals',
+              },
+            },
+          },
+          score: { allowNull: false, type: Sequelize.DECIMAL(3, 1) },
+          createdAt: { allowNull: false, type: Sequelize.DATE },
```
GarrettEHill, line 42: can we have a column for ignore bool. or should this be a migration onto this table later?

src/models/hooks/goal.js
```diff
@@ -57,6 +58,22 @@ const preventNameChangeWhenOnApprovedAR = (sequelize, instance) => {
...
+  if (changed.includes('name')) {
+    await sequelize.models.SimScoreCache.destroy({
+      where: {
+        [Op.or]: [
+          { goal1: instance.id },
+          { goal2: instance.id },
+        ],
+      },
+      transaction: options.transaction,
+    });
```
GarrettEHill, line 73: is destroy the right path, could we just set the score to null? Since we know the record will be needed again soon

src/models/simScoreCache.js
```diff
@@ -0,0 +1,46 @@
...
+   * RolesTopic table. Junction table between Roles and Topics to support many to many relationship.
+   *
+   * @param {} sequelize
+   * @param {*} DataTypes
+   */
+export default (sequelize, DataTypes) => {
+  class SimScoreCache extends Model {
+    static associate(models) {
+      SimScoreCache.belongsTo(models.Goal, { foreignKey: 'id', as: 'goalOne' });
+      SimScoreCache.belongsTo(models.Goal, { foreignKey: 'id', as: 'goalTwo' });
```
GarrettEHill, line 15: Please add the reversed associations. You have SimScoreCache -> Goals. But not Goals -> SimScoreCache

similarity_api/src/settings.py
nvms: `similarity_api/src/jobs.sqlite`. It does not need to live across deploys or redeploys. It is created if it doesn't exist, and when the app starts for the first time, the job is deleted from the DB if it exists and is recreated. In other words, we can safely not deploy this file. I added it to gitignore here: c26190b06c09b5594d1596abe9ec0ccf1e133c04

similarity_api/src/app.py
nvms: 👍 b2726d0509c32c8284d625630376cc81f37bbeb8

similarity_api/src/application.py
nvms: I thought it would be helpful for other developers to also have a way to quickly create short-interval jobs for debugging, but I can remove it. cee88711e27b46deb7e81d0072309dd3b8e1a027

similarity_api/src/routes/create.py
nvms: I've been using [isort](https://pypi.org/project/isort/) to handle import linting and sorting so that we don't have to think about these things. I added a config (a2a08fedc5775eb438f580666b9384cddfd213e7) for isort and used `force_grid_wrap` which implements the rule you are looking for.

similarity_api/src/sim/compute.py
nvms: 7913b81e9

src/migrations/20231030150724-sim_score_cache.js
nvms: 383308046

src/migrations/20231030150724-sim_score_cache.js
nvms: Maybe we should wait on this one until we've figured out this logic a bit more. If it's as simple as we think it will be, then the migration will be extremely quick and easy -- but maybe that won't be the case.

src/models/hooks/goal.js
nvms: I really don't think there's much difference, but I could be wrong. Is there a good reason to make this change that I'm not seeing?

src/models/simScoreCache.js
nvms: 050b0985b - not sure how I feel about the naming choice for the `as`, but I'm also not sure it matters much. 🤷

similarity_api/src/application.py
nvms: 287ad2e2165a6f80c3ef6c424dfa90b3fc3e5609

src/migrations/20231030150724-sim_score_cache.js
nvms: be077c98b3d91e2d120df36d2b124810edf217d8

PR #1798, opened 2023-10-31
[TTAHUB-2259] Similarity node communication layer

nvms:
## Description of change

This is the communication layer between node and the similarity API. A PR from @thewatermethod was merged into this that formats the response for consumption by the FE.

One route is added:

```
GET /goals/similar/:recipientId?cluster=true/false
```

## How to test

Make sure you have a value for `SIMILARITY_ENDPOINT` in your `.env`. Locally, this value should be `http://similarity_api:9100/compute`. Go to the RTR, select a recipient and go to the RTTAPA page. Look in the network page and ensure the call to the endpoint was successful and returns something like:

```javascript
[
  {
    "ids": [...],
    "goals": [
      {
        "name": "...",
        "source": null,
        "status": "Closed",
        "responsesForComparison": "",
        "ids": [...]
      },
      {
        "name": "...",
        "source": null,
        "status": "Closed",
        "responsesForComparison": "",
        "ids": [...]
      }
    ]
  }
]
```

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2259

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/routes/goals/index.js
```diff
@@ -14,5 +15,6 @@ router.get('/', transactionWrapper(retrieveGoalsByIds));
 router.get('/:goalId/recipient/:recipientId', transactionWrapper(retrieveGoalByIdAndRecipient));
 router.put('/changeStatus', transactionWrapper(changeGoalStatus));
 router.delete('/', transactionWrapper(deleteGoal));
+router.post('/similar', transactionWrapper(postSimilarGoalsForRecipient));
```
thewatermethod, line 18: any reason why this is a post and not a get?

src/services/similarity.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * @async
+ * @param {number} recipient_id - The ID of the recipient.
+ * @param {boolean} [cluster] - Specifies whether to cluster the results. Default value is false.
+ * @returns {Promise<Array>} A promise that resolves to an array of similar goals.
+ */
+export default async function getSimilarGoalsForRecipient(recipient_id, cluster) {
+  const { SIMILARITY_ENDPOINT } = process.env;
+
+  const response = await fetch(
```
thewatermethod, line 10: should this fetch be within the try?

src/services/similarity.js
```diff
@@ -0,0 +1,53 @@
...
+        recipient_id,
+        cluster,
+        alpha: 0.9,
+      }),
+    },
+  );
+
+  /**
+    * response without clustering looks like:
+    * result: [
```
thewatermethod, line 27: I'm sure it's evident but with clustering is it something like ```[1, 2, 3, 4]```?

thewatermethod:
Might be good to add some tests

src/services/similarity.js
thewatermethod: yeah, just looked more closely at the similarity code, I see what it'll look like

similarity_api/src/sim/compute.py
```diff
@@ -108,39 +108,60 @@ def calculate_batch_similarity(batch, nlp):
...
-                        "name": goal,
-                        "grantId": recipient_id,
-                    },
-                    "goal2": {
+            if cluster:
+                matches = []
+                for k in cur_potential_idx:
+                    if i + j > i + k or current_ids[k] in matched_ids:
+                        continue
+                    matches.append({
```
thewatermethod, line 134: Can we also include the "isCurated" column in the initial query to the Goals table and only allow one "isCurated" goal in this?

src/routes/goals/index.js
nvms: oversight on my part, made it a get here: 66e2a0450bd2a72cff78d4b2ae89d238342ee1b2

src/services/similarity.js
nvms: it 100% should be: 0e7c21352b5c958802a655d8e929847e8a9fe9ff

src/routes/goals/handlers.js
```diff
@@ -227,3 +228,40 @@ export async function retrieveGoalByIdAndRecipient(req, res) {
     await handleErrors(req, res, error, `${logContext}:RETRIEVE_GOAL_BY_ID_AND_RECIPIENT`);
   }
 }
+
+export async function getSimilarGoalsForRecipient(req, res) {
+  const recipientId = parseInt(req.params.recipient_id, 10);
+
+  if (Number.isNaN(recipientId)) {
+    return res.status(400).send('Recipient ID must be an integer');
```
thewatermethod, line 236: one small thing: we have middleware for checking the params in this manner - you'd still have to parseInt here but you wouldn't need the if statement or a test for it... I know we're pretty inconsistent about using the middleware

src/routes/goals/handlers.js
```diff
@@ -227,3 +228,40 @@ export async function retrieveGoalByIdAndRecipient(req, res) {
...
+      return new Goal(user, goal).canView();
+    }));
+
+    if (!canView.every((permission) => permission)) {
+      return res.sendStatus(401);
+    }
+
+    return res.json(similarGoalIds);
+  } catch (error) {
+    return res.status(500);
```
thewatermethod, line 265: this should probably use our "handleErrors" function

src/routes/goals/handlers.js
```diff
@@ -227,3 +228,40 @@ export async function retrieveGoalByIdAndRecipient(req, res) {
...
+  try {
+    const similarGoalIds = await similarGoalsForRecipient(recipientId, cluster);
+
+    const ids = Array.from(similarGoalIds.reduce((acc, resp) => {
+      const id1 = parseInt(resp.goal1.id, 10);
+      const id2 = parseInt(resp.goal2.id, 10);
+      acc.add(id1);
+      acc.add(id2);
+      return acc;
+    }, new Set()));
```
thewatermethod, line 252: won't this produce a one-dimensional array? in other words, we'd never see the case like in the mockup where a recipient has two groups of similar goals, i.e

Screenshot 2023-11-08 at 8 52 22 AM

src/services/similarity.js
```diff
@@ -0,0 +1,52 @@
...
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({
+          recipient_id,
+          cluster,
+          alpha: 0.9,
+        }),
+      },
+    );
+    return await response.json();
```
thewatermethod, line 48:
```suggestion
    return response.json();
```

src/routes/goals/handlers.js
nvms: ah, yeah: 6e3e705a5c00e2859eb76d781c362b2aac57d149

src/routes/goals/handlers.js
nvms: 🙏 d25bd1971207b3c867a1bba6c8f1552ab92d4b2b

nvms: er, fc66a49b7d89377db46c1c268725252eb0b3ec91

src/routes/goals/handlers.js
thewatermethod: ohhhhh ok gotcha, should have read more closely

similarity_api/src/sim/compute.py
nvms: yeah, I can add an `isCurated` property based on `GoalTemplate.creationMethod === 'Curated'`. can you explain the second part of your question a bit more? I don't think I understand it completely

src/services/similarity.js
```diff
@@ -0,0 +1,52 @@
...
+        *    goal2: {
+          *      id: 2, name: "Identify strategies", grantId: 2,
+          *    },
+        *    similarity: 0.921823748234,
+        *  },
+      * ]
+    */
+  try {
+    const { SIMILARITY_ENDPOINT } = process.env;
+    const response = await fetch(
```
thewatermethod, line 34: I am very out of the loop I guess, I had no idea fetch was in node core now

similarity_api/src/sim/compute.py
nvms: show all the matches, but if the matches include more than one curated goal, only show one curated goal? is that right?

thewatermethod: A group of matched goals can only include one "curated" goal, regardless of text match

thewatermethod: yes, exactly

src/routes/goals/handlers.js
thewatermethod: Usage might be a little off: 

We use the middleware in routes/goals/index.js like

```js
router.get(
  '/similar/:recipient_id', 
  (req, res, next) => checkIdParam(req, res, next, 'recipient_id'),
  transactionWrapper(getSimilarGoalsForRecipient)
);
```

then, in getSimilarGoalsForRecipient you'd just do one of these:

```js
const recipientId = parseInt(req.params.recipient_id, DECIMAL_BASE); 
```

nvms: 091bf508c2647b77908e3398e6efce1e9d48ef9b

oh jeez, I feel dumb. thank you

src/services/similarity.js
nvms: yeah! I don't remember when it happened (16? 18?) but I'm happy it did

src/routes/goals/handlers.js
```diff
@@ -227,3 +229,36 @@ export async function retrieveGoalByIdAndRecipient(req, res) {
...
+export async function getSimilarGoalsForRecipient(req, res) {
+  const recipientId = parseInt(req.params.recipientId, DECIMAL_BASE);
+  const cluster = Object.prototype.hasOwnProperty.call(req.query, 'cluster');
+  const userId = await currentUserId(req, res);
+  const user = await userById(userId);
+
+  try {
+    const similarGoalIds = await similarGoalsForRecipient(recipientId, cluster);
+
+    const ids = Array.from(similarGoalIds.reduce((acc, resp) => {
```
thewatermethod, line 242: This reduce doesn't work if cluster is true, as the data comes back in a different shape.
My [PR](https://github.com/HHS/Head-Start-TTADP/pull/1814) has a refactor. 
Maybe for now we don't need the non-cluster version?

similarity_api/src/sim/compute.py
thewatermethod: @nvms you can disregard this, unless you've already gone far down the rabbit hole. I had a brainstorm last night about a different way to solve this problem.

src/routes/goals/handlers.js
nvms: thanks for noticing this - just merged your PR

src/testUtils.js
```diff
@@ -244,10 +244,12 @@ export async function createGoal(goal) {
...
-  const dbGoalTemplate = await GoalTemplate.findOrCreate({
-    where: { templateName: dg.name },
-    defaults: { templateName: dg.name },
-  });
+  const dbGoalTemplate = goal.goalTemplateId
+    ? { id: goal.goalTemplateId }
+    : await GoalTemplate.findOrCreate({
+      where: { templateName: dg.name },
+      defaults: { templateName: dg.name },
+    });
```
kryswisnaskas, line 252:
```suggestion
  const dbGoalTemplate = goal.goalTemplateId
  ? { id: goal.goalTemplateId }
  : (await GoalTemplate.findOrCreate({
      where: { templateName: dg.name },
      defaults: { templateName: dg.name },
    }))[0];
```

kryswisnaskas: since `findOrCreate `returns an array

thewatermethod:
@nvms Pushed some fixes up for unit tests that went haywire, should this be opened up to wider review now?

nvms:
> @nvms Pushed some fixes up for unit tests that went haywire, should this be opened up to wider review now?

Thank you! Yes, good idea

src/testUtils.js
nvms: Thank you!

thewatermethod:
Looks we still need tests for the similarity.js service

Screenshot 2023-11-15 at 12 03 53 PM

thewatermethod:
Note for testing with docker
I had to change similarty_api/src/settings.py:23 from

```python
DB_HOST = os.environ.get("POSTGRES_HOST") or (vcap_credentials and vcap_credentials.get('host'))
```

to

```python
DB_HOST = 'postgres_docker'
```

as "localhost" doesn't work

thewatermethod (approved):
Just ran some more UI tests, looks very good. Just need some other reviews

kryswisnaskas:
Could we deploy this to Sandbox? I am getting some different intermittent errors running locally. Ocassionally, the communication layer does work as advertised 👍 

I am also seeing:

```
2023-11-22T15:45:16.771Z AUDIT error:   undefined - id: 2519 UNEXPECTED ERROR - TypeError: Cannot read properties of undefined (reading 'map')
    at getGoalIdsBySimilarity (c:\hs\Head-Start-TTADP\src\goalServices\goals.js:2951:43)
```

thewatermethod:
@nvms looks like we may be missing something from the config to connect the services ?

```
Error: SyntaxError: Unexpected token < in JSON at position 0\n    at similarGoalsForRecipient (/home/vcap/app/build/server/src/services/similarity.js:22:15)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async getSimilarGoalsForRecipient (/home/vcap/app/build/server/src/routes/goals/handlers.js:212:28)\n    at async /home/vcap/app/build/server/src/routes/transactionWrapper.js:23:30\n    at async /home/vcap/app/node_modules/sequelize/lib/sequelize.js:507:18","timestamp":"2023-11-22T20:36:08.303Z"}
```

thewatermethod:
> Looks like SIMILARITY_API_KEY was missing. I added it to the application and am redeploying. I will make sure all other envs have this as well

@nvms are we sure that was it? I'm still seeing the same error on sandbox (ran the deploy pipeline just in case)

nvms:
I think perhaps it was this?

47bec58e431b7f28930aa9d61de82cf16a541544

🤦

src/goalServices/helpers.js
```diff
@@ -0,0 +1,22 @@
...
+  .map((gfr) => gfr.response).sort().join();
+
+const goalFieldTransate = {
+  name: 'goalText',
+  status: 'goalStatus',
+  source: 'source',
+  responsesForComparison: 'responsesForComparison',
+};
+
+const findOrFailExistingGoal = (needle, haystack, translate = goalFieldTransate) => haystack.find(
```
AdamAdHocTeam, line 11: Are these technical terms?

src/routes/goals/handlers.js
```diff
@@ -256,3 +259,38 @@ export async function mergeGoalHandler(req, res) {
     await handleErrors(req, res, err, `${logContext}:MERGE_GOAL`);
   }
 }
+
+export async function getSimilarGoalsForRecipient(req, res) {
+  const recipientId = parseInt(req.params.recipientId, DECIMAL_BASE);
```
AdamAdHocTeam, line 264: Do we need to check permissions here? Or maybe it's on the FE. For example we have a user visiting the page who can't perform a merge then we don't need to do any this.

src/services/recipient.js
```diff
@@ -523,13 +527,13 @@ export async function getGoalsByActivityRecipient(
...
       [sequelize.literal(`
-        CASE 
-          WHEN COALESCE("Goal"."status",'')  = '' OR "Goal"."status" = 'Needs Status' THEN 1 
-          WHEN "Goal"."status" = 'Draft' THEN 2 
-          WHEN "Goal"."status" = 'Not Started' THEN 3 
-          WHEN "Goal"."status" = 'In Progress' THEN 4 
-          WHEN "Goal"."status" = 'Closed' THEN 5 
-          WHEN "Goal"."status" = 'Suspended' THEN 6 
+        CASE
+          WHEN COALESCE("Goal"."status",'')  = '' OR "Goal"."status" = 'Needs Status' THEN 1
```
AdamAdHocTeam, line 531: Should we be using ENUM (constants) here?

src/routes/goals/handlers.js
thewatermethod: we check here, a little below this line

src/services/recipient.js
nvms: Perhaps, but this particular change is just the removal of some unnecessary trailing whitespace.

AdamAdHocTeam (approved):
Overall looks good. Tested locally and on sandbox.

Left some minor comments on code 👀.

Thanks for hooking up support for local db dev envs.

src/routes/goals/handlers.js
AdamAdHocTeam: Just want to make sure we are not doing processing we don't need to.

nvms: A read permission check happens a few lines down:

https://github.com/HHS/Head-Start-TTADP/blob/76beb59e008198cb88105da49411718c9ad09654/src/routes/goals/handlers.js#L278-L289

Since this handler is not responsible for any mutation of a Goal, I don't think we need to check for write permissions here - probably should happen in an actual merge handler.

src/goalServices/helpers.js
nvms: 🪡

PR #1815, opened 2023-11-07
[TTAHUB-2289] ActivityReportGoalFieldResponse incident

GarrettEHill:
## Description of change
Restore original field responses

## How to test
data in ActivityReportGoalFieldResponse  is no longer a single value

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2289

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


src/migrations/20231107000000-ActivityReportGoalFieldResponse-incedint.js
```diff
@@ -0,0 +1,67 @@
...
+            SELECT
+              dml_txid,
+              dml_by,
+              dml_type,
+              COUNT(*),
+              MIN(dml_timestamp),
+              MAX(dml_timestamp)
+            FROM "ZALActivityReportGoalFieldResponses" zargfr
+            GROUP BY 1,2,3
+            HAVING COUNT(*) > 400
```
AdamAdHocTeam, line 24: I get that your flagging incidents in this CTE. But why not just simply take everything that appears more than once with the same txn id as an incident. Why do we say exactly 400. Can we have the same txn id over mutliple rows?

AdamAdHocTeam: Or the inverse of this and get the max id (the value to set).

src/migrations/20231107000000-ActivityReportGoalFieldResponse-incedint.js
```diff
@@ -0,0 +1,67 @@
...
+            JOIN "Goals" g
+            ON arg."goalId" = g.id
+            JOIN "GoalFieldResponses" gfr
+            ON g.id = gfr."goalId"
+            JOIN "ZALActivityReportGoalFieldResponses" zargfr
+            ON argfr.id = zargfr.data_id
+            LEFT JOIN incidents i
+            ON zargfr.dml_txid = i.dml_txid
+            AND zargfr.dml_type = i.dml_type
+            where argfr.response != gfr.response
```
AdamAdHocTeam, line 52: Do we need to take into account if the goal is on a approved report?

AdamAdHocTeam: I didn't see AND i.count IS NULL

src/migrations/20231107000000-ActivityReportGoalFieldResponse-incedint.js
AdamAdHocTeam: NVM I see what you are saying here we need to catch updates to all rows as "bad".

AdamAdHocTeam (commented):
@GarrettEHill we talked yesterday about a use case where the values in the GoalFieldResponse table might have been inadvertently updated with a save to the AR.

Two questions:
1. Are we relying on the GFR table as a source of truth (I don't think we are).
2. Do we need a similar script for the GFR table (or did you find this wasn't an issue).

src/migrations/20231107000000-ActivityReportGoalFieldResponse-incedint.js
AdamAdHocTeam: I don't think we can use gfr table as a source of truth.

If the GFR table is potentially messed up also we might need to first fix that table from audit log.

GarrettEHill: I don't use the GFR table as a source of truth, I use it to see if there is a delta. When a delta is present, I recalculate that value from the audit log not the GFR

GarrettEHill:
> @GarrettEHill we talked yesterday about a use case where the values in the GoalFieldResponse table might have been inadvertently updated with a save to the AR.
> 
> Two questions:
> 
> 1. Are we relying on the GFR table as a source of truth (I don't think we are).
I use the GFR table as a reference, but not as a source for the final value produced
> 2. Do we need a similar script for the GFR table (or did you find this wasn't an issue).
I checked the GFR table, no issue was present

AdamAdHocTeam (approved):
Overall looks good and spot checked a handful of them.

src/migrations/20231107000000-ActivityReportGoalFieldResponse-incedint.js
```diff
@@ -0,0 +1,67 @@
...
+            AND i.count IS NULL
+            GROUP BY 1,2,3,4,5
+          )
+          UPDATE "ActivityReportGoalFieldResponses" argfr
+          SET
+            response = COALESCE(r.last_valid_response, ARRAY[]::text[])
```
kryswisnaskas, line 58: Wouldn't it make more sense to use the 'goal response' instead of setting the response to empty?

GarrettEHill: The value in the goal response may have been set after the report was approved. Should I set the value on the report to a value set after the report was approved?

kryswisnaskas: Yes, it would be helpful if you could set those values, even though they were chosen after the ARs were approved. Angela also agrees with this approach.

kryswisnaskas (approved):
👍

PR #1865, opened 2023-12-11
[TTAHUB-2379] Update required fields on communication log

thewatermethod:
## Description of change
- Result and next steps should not be required. (This means the next steps component, also used in other forms, should accept a "required" parameter)
- The RTTAPA updates table should link to the log entry in the same way the communication log does

## How to test
Confirm the above are resolved. Note I left the validation "onBlur" for the dates just in case someone tries to type "cactus" or "never" in there, but the form can be saved with any value.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2379

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.js
+++ b/frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.js
@@ -181,7 +181,7 @@ export default function NextStepsRepeater({
                   name={`${name}[${index}].completeDate`}
                   value={item.completeDate}
                   dataTestId={`${name === 'specialistNextSteps' ? 'specialist' : 'recipient'}StepCompleteDate-input`}
-                  required={false}
+                  required={required}
                 />
               </div>
             </FormGroup>
--- a/frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.js
+++ b/frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.js
@@ -19,6 +19,7 @@ export default function NextStepsRepeater({
   name,
   ariaName,
   recipientType,
+  required,
 }) {
   const [heights, setHeights] = useState([]);
 
@@ -95,6 +96,13 @@ export default function NextStepsRepeater({
     ? `When does the ${recipientLabel} anticipate completing step ${index + 1}?`
     : `When do you anticipate completing step ${index + 1}?`);
 
+  const textareaRegister = (() => {
+    if (required) {
+      return register({ required: 'Enter a next step' });
+    }
+    return register();
+  })();
+
   return (
     <>
       <div className="ttahub-next-steps-repeater">
@@ -109,7 +117,7 @@ export default function NextStepsRepeater({
                 htmlFor={`${stepType}-next-step-${index + 1}`}
               >
                 {`Step ${index + 1}`}
-                <Req />
+                {required && (<Req />)}
               </Label>
               {(errors[name]
                 && errors[name][index] && errors[name][index].note)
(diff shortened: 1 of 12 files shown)


frontend/src/pages/RecipientRecord/pages/CommunicationLogForm/pages/nextSteps.js
```diff
@@ -1,43 +1,55 @@
...
-const fields = Object.keys(nextStepsFields);
 const path = 'next-steps';
+const visitedField = `pageVisited-${path}`;
+const fields = [visitedField];
 const position = 3;
 
+const NextSteps = () => {
+  const { register } = useFormContext();
+
+  useCompleteSectionOnVisit(visitedField);
```
thewatermethod, line 21: the only changes to this file really, despite the large diff, is that I added this hook to manipulate the value in the hidden input, of which the adding is the other change

frontend/src/pages/RecipientRecord/pages/ViewCommunicationLog/components/DisplayNextSteps.js
```diff
@@ -2,16 +2,24 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import ReadOnlyField from '../../../../../components/ReadOnlyField';
 
-export default function DisplayNextSteps({ title, steps }) {
+export const skipDisplaySteps = (steps) => {
```
thewatermethod, line 5: Function to eliminate empty next steps sections from the "view communication log" page

frontend/src/pages/ActivityReport/Pages/components/NextStepsRepeater.js
```diff
@@ -173,6 +181,7 @@ export default function NextStepsRepeater({
                   name={`${name}[${index}].completeDate`}
                   value={item.completeDate}
                   dataTestId={`${name === 'specialistNextSteps' ? 'specialist' : 'recipient'}StepCompleteDate-input`}
+                  required={false}
```
AdamAdHocTeam, line 184: Should this be the value of 'required' var?

frontend/src/widgets/RttapaUpdates.js
```diff
@@ -42,7 +42,11 @@ export default function RttapaUpdates({
             <tbody>
               {logs.map((log) => (
                 <tr key={`logrow${log.id}`}>
-                  <td data-label="Date">{log.data ? log.data.communicationDate : ''}</td>
+                  <td data-label="Date">
+                    <Link to={`/recipient-tta-records/${recipientId}/region/${regionId}/communication/${log.id}/view`}>
+                      {log.data ? log.data.communicationDate : ''}
```
AdamAdHocTeam, line 47: I know this is the way it was before, do we need to worry about the log.data.communicationDate being blank and not having a link here? Do we want a default value just in case.

frontend/src/pages/RecipientRecord/pages/CommunicationLogForm/pages/log.js
```diff
@@ -144,14 +144,14 @@ const Log = ({ datePickerKey }) => {
...
+          label="Result"
           name="result"
+          required={false}
         >
           <Dropdown
-            required
             id="result"
             name="result"
-            inputRef={register({ required: 'Select a result of communication' })}
+            inputRef={register()}
```
AdamAdHocTeam, line 154: Result defaults to the below value:

image

Do we want it to default to - Select - so they don't have any information they haven't set applied?

frontend/src/widgets/RttapaUpdates.js
thewatermethod: I've already discussed this with UX. There is currently no way to save a log entry without a date, so no need for a fallback here. If it turns out to be an issue we can circle back.

frontend/src/pages/RecipientRecord/pages/CommunicationLogForm/pages/log.js
thewatermethod: Sure, I added a default disabled option

AdamAdHocTeam: Thanks, I think this makes more sense then saving something they didn't select.

AdamAdHocTeam (commented):
We are still default selecting 'New TTA accepted' instead of 'Select one'.

AdamAdHocTeam (approved):
Issues are fixed. Nice job.

PR #1870, opened 2023-12-13
[TTAHUB-2391] Merging curated goals feature flag and logic

nvms:
## Description of changep

> I think we could just use the most recent root cause. If they closed and opened a new one with a new root, then we should use that one. - Tammy

Changes:

- Add feature flag: `closed_goal_merge_override`
- Similarity result aggregator checks for the feature flag, and if you have it, then we don't exclude closed goals in the list of available goals to merge.
- On the FE, we now auto select the latest curated goal (as the final goal) when more than one curated goals are selected.
- The goal selected as the final goal is the one whose root causes will be used for the newly created goal, the other root causes are discarded.

## How to test

In the most recent prod dump, go to the RTR for recipient 1884, region 8. Without this change, you would see one group of similar goals. With this change you would see two groups of similar goals if you have the feature flag or are an admin.

Selecting the first group, you see one in progress FEI goal, and one closed FEI goal. Go through the merge process. Whichever goal you choose as the final goal, you will keep those root causes.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2391

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


GarrettEHill (approved):

thewatermethod:
Works great!

My only question is whether you want to adjust this code: 

```js
  // frontend/src/pages/RecipientRecord/pages/MergeGoals/components/GoalCard.js
  if (selectedGoalsIncludeCurated && !goal.isCurated) {
    showFormControl = false;
  }
```

to hide the radio buttons for the unselected goal, or if we are letting the admin decide which goal to keep. If I'm following all this correctly, since the backend will now check the final goal to see which goal to keep goal field responses from, this would allow users to break "tammy's rule" above.

This might be fine given this is not intended to be a perfect UI but instead a way to fix a customer issue, up to you

thewatermethod (approved):

nvms:
> since the backend will now check the final goal to see which goal to keep goal field responses from, this would allow users to break "tammy's rule" above.

Good catch! Yeah, this needs to be updated. I'll make this change.

PR #1871, opened 2023-12-13
Limit AR Groups

AdamAdHocTeam:
## Description of change

We need to limit what groups are returned to the AR (public vs private) not just region.

**NOTE:** This code will be replaced by Garrett's re-write and is a place holder for now.

## How to test

Make sure when you are creating a recipient AR and use the 'use group' check box. You can ONLY see groups that are public or created by the user and private.

## Issue(s)

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/groups.ts
+++ b/src/services/groups.ts
@@ -39,7 +39,6 @@ interface GroupResponse {
 }
 
 export async function groupsByRegion(region: number, userId?: number): Promise<GroupResponse[]> {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let where: WhereOptions = {
     '$grants.regionId$': { [Op.eq]: region },
   };
--- a/src/services/groups.ts
+++ b/src/services/groups.ts
@@ -38,7 +38,7 @@ interface GroupResponse {
   isPublic: boolean;
 }
 
-export async function groupsByRegion(region: number, userId: number): Promise<GroupResponse[]> {
+export async function groupsByRegion(region: number, userId?: number): Promise<GroupResponse[]> {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   let where: any = {
     '$grants.regionId$': { [Op.eq]: region },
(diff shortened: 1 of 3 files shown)


src/services/groups.ts
```diff
@@ -38,16 +38,29 @@ interface GroupResponse {
   isPublic: boolean;
 }
 
-export async function groupsByRegion(region: number): Promise<GroupResponse[]> {
+export async function groupsByRegion(region: number, userId?: number): Promise<GroupResponse[]> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let where: any = {
```
thewatermethod, line 43:
```suggestion
  let where: WhereOptions = {
```

thewatermethod: also change line to import { Op, WhereOptions } from 'sequelize';

thewatermethod: (please, sorry)

thewatermethod (approved):

src/services/groups.ts
```diff
@@ -38,16 +38,29 @@ interface GroupResponse {
   isPublic: boolean;
 }
 
-export async function groupsByRegion(region: number): Promise<GroupResponse[]> {
```
thewatermethod, line 41: this function is now basically the same as the one below it, "groups." This is fine I guess but a better solution would be to just include recipients in that query and use that one, rather than proliferating these nearly identical functions

AdamAdHocTeam: Yes you are correct in Garretts re-write he has removed one of these functions. So ultimately the call from AR index will go away.

src/services/groups.ts
AdamAdHocTeam: done.

src/services/groups.ts
```diff
@@ -38,16 +38,29 @@ interface GroupResponse {
...
+export async function groupsByRegion(region: number, userId?: number): Promise<GroupResponse[]> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
thewatermethod, line 42: let's remove that eslint override also since we aren't using "any" now

PR #1878, opened 2023-12-19
[TTAHUB-2153] BE for IPD Courses

AdamAdHocTeam:
## Description of change

This add the BE endpoint to import the ipd courses.

## How to test

- Review all code.
- Make sure tests pass.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2153

### Every PR

- [x] Meets issue criteria
- [x] JIRA ticket status updated
- [x] Code is meaningfully tested
- [ ] Meets accessibility standards (WCAG 2.1 Levels A, AA)
- [ ] API Documentation updated
- [ ] Boundary diagram updated
- [ ] Logical Data Model updated
- [ ] [Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions
- [ ] UI review complete

### Production Deploy

- [ ] Staging smoke test completed

### After merge/deploy

- [x] Update JIRA ticket status

PR #1885, opened 2023-12-22
[TTAHUB-2369] Admin course import FE

AdamAdHocTeam:
## Description of change

Add a course import to to the admin menu.

## How to test

- Review all code.
- All tests pass.
- Test the three cases for course import:
1. **Creates**: Course doesn't exist in db.
2. **Updates**: Course already exists in db with and without (spacing special chars are diff) exact wording , updated at is changed.
3. **Marks deleted**: Course pre existed in the db but is no longer in the import.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2369

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy


frontend/src/pages/Admin/components/CsvImport.js
```diff
@@ -0,0 +1,267 @@
...
+      importIds.shift();
+
+      // Remove any blank values.
+      importIds = importIds.filter((id) => id !== '');
+
+      const distinctImportIds = [...new Set(importIds)];
+
+      // Check if we have duplicate primary Id Column values.
+      if (distinctImportIds.length !== importIds.length) {
+        setError(`Duplicate ${primaryIdColumn}'s found. Please correct and try again.`);
```
thewatermethod, line 133:
```suggestion
        setError(`Duplicate ${primaryIdColumn}s found. Please correct and try again.`);
```

thewatermethod:
Really like the generic component approach, I think that makes a lot of sense. Worked well when I was testing it.

Couple of thoughts:

As is, this seems like it would be pretty annoying to use. The file attached to the Jira, for example, failed because it contained a duplicate, but it gave no feedback on what the duplicate was. Do we want to fail on duplicate or just skip? If we fail, do we want to identify what the duplicate was? I know that we are trying to set our customer up to maintain this, but we should keep in mind that we'll be supporting it if it becomes a struggle.

Also, are we sure that we want to delete all the courses that don't appear on the CSV when it is uploaded? I assume you checked, I don't see that in the Jira specs.

src/migrations/20231206210453-update-flags-for-user.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     await queryInterface.sequelize.transaction(async (transaction) => {
       const sessionSig = __filename;
       await prepMigration(queryInterface, transaction, sessionSig);
-      return updateUsersFlagsEnum(queryInterface, transaction, ['training_reports']);
+      // return updateUsersFlagsEnum(queryInterface, transaction, ['training_reports']);
```
thewatermethod, line 9: this shouldn't be in this PR, should it?

src/constants.js
```diff
@@ -223,6 +223,7 @@ const FEATURE_FLAGS = [
   'goal_source',
   'merge_goals',
   'monitoring',
+  'training_reports',
```
thewatermethod, line 226: this shouldn't be in this PR, should it?

src/services/course.test.js
```diff
@@ -0,0 +1,275 @@
+import { Op } from 'sequelize';
+import { expect } from '@playwright/test';
```
thewatermethod, line 2: I don't think we want to import this here

frontend/src/pages/Admin/components/CsvImport.js
AdamAdHocTeam: Fixed.

AdamAdHocTeam:
> Really like the generic component approach, I think that makes a lot of sense. Worked well when I was testing it.
> 
> Couple of thoughts:
> 
> As is, this seems like it would be pretty annoying to use. The file attached to the Jira, for example, failed because it contained a duplicate, but it gave no feedback on what the duplicate was. Do we want to fail on duplicate or just skip? If we fail, do we want to identify what the duplicate was? I know that we are trying to set our customer up to maintain this, but we should keep in mind that we'll be supporting it if it becomes a struggle.
> 
> Also, are we sure that we want to delete all the courses that don't appear on the CSV when it is uploaded? I assume you checked, I don't see that in the Jira specs.

As we discussed message for duplicates has been updated. Also the BE ticket outlines marking deleted when course no longer appears in the list.

src/services/course.test.js
AdamAdHocTeam: Good call.

src/migrations/20231206210453-update-flags-for-user.js
AdamAdHocTeam: There were actually two commented out migrations. Not sure why but was having issues updating db.

src/constants.js
AdamAdHocTeam: I think this was because my db already expected it. Will remove.

nvms (commented):
Minor comments

src/services/course.js
```diff
@@ -19,9 +19,12 @@ export async function csvImport(buffer) {
 
   const parsed = parse(buffer, { skipEmptyLines: true, columns: true });
   let rowCount = 1;
-  const results = parsed.map(async (course) => {
-    try {
-      let rawCourseName = course['course name'];
+  let results;
+  try {
+    results = await Promise.all(parsed.map(async (course) => {
```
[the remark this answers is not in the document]
AdamAdHocTeam: Will look to make this in the BE PR: https://github.com/HHS/Head-Start-TTADP/pull/1878

src/services/course.js
```diff
@@ -19,9 +19,12 @@ export async function csvImport(buffer) {
...
+      // Get the first property value form course object.
+      let rawCourseName = Object.values(course)[0];
```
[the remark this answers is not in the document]
AdamAdHocTeam: done in other pr.

src/services/course.js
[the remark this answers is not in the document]
AdamAdHocTeam: No need to make this change end point is wrapped in txn. Also bad practice per sql doc according to @thewatermethod.

thewatermethod:
looks good, should the target branch for this be the backend PR instead of main?

AdamAdHocTeam:
> looks good, should the target branch for this be the backend PR instead of main?

Done.

thewatermethod (approved):

PR #1889, opened 2023-12-27
[TTAHUB-2170] Update Files: cannot read properties of null (dataValues)

nvms:
## Description of change

This one is more straight forward. It seems like we access dataValues in only one spot in this pipeline. We should check the return of `findOne` before accessing `dataValues`. If it is `null`, throw an error to short circuit the `processFile` operation.

## How to test

`files.test.js` has been updated to check for this

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2170

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/workers/files.js
+++ b/src/workers/files.js
@@ -12,6 +12,8 @@ const {
   REJECTED,
 } = FILE_STATUSES;
 
+const KEY_NOT_FOUND = 'File with key not found.';
+
 /**
  * Get's file Id from DB using the s3 key
  * @param {string} key - uuid with a file extension representing the s3 key of the file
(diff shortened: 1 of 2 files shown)


src/workers/files.js
```diff
@@ -55,6 +58,11 @@ const processFile = async (key) => {
     });
     await updateFileStatus(key, APPROVED);
   } catch (error) {
+    // If the key couldn't be found, we can't update the status.
+    if (error.message === KEY_NOT_FOUND) {
+      throw error;
+    }
```
kryswisnaskas, line 64: Do we want to throw here rather than allow a re-try? The current code will attempt to retry if scanning is not rejected.

nvms: Yeah, that's a good point. Updated here: 1dc7abf0c12830abdd1dbbc28fc27a8b21eca8c8

thewatermethod (approved):

PR #1899, opened 2024-01-04
[TTAHUB-2124] Remove aro topics from version 1 reports

AdamAdHocTeam:
## Description of change

When exporting an older report (id: 10558), we noticed that both the ARO and AR (legacy) level topics were populated. This should not be the case as this report was created and approved before the introduction of ARO's. After some investigation it turns out that the creation of these topics was from a migration file '20220802000000-goal-objective-metadata-caching'. The attached script will delete any ARO topics that were created as a result of this migration on version 1 reports.

## How to test

- Review the migration.
- Make sure you agree with the cleanup (criteria used to determine what we delete).
- Export report 10558 and see the 'Objective 1.1 Topics' field populated in addition to the 'Legacy Topics covered'.
- Run the migration and export the report csv only the legacy topics should be populated.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2124

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20240104110924-remove-unwanted-aro-topics.js
+++ b/src/migrations/20240104110924-remove-unwanted-aro-topics.js
@@ -1,0 +1,30 @@
+const {
+  prepMigration,
+} = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, _Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      await queryInterface.sequelize.query(
+        `DELETE FROM "ActivityReportObjectiveTopics"
+        WHERE "activityReportObjectiveId" IN (
+            SELECT aro."id"
+            FROM "ActivityReportObjectives" aro
+            INNER JOIN "ActivityReports" ar
+                ON aro."activityReportId" = ar.id
+            WHERE aro.id IN (
+                SELECT ("new_row_data"->'activityReportObjectiveId')::int
+                FROM "ZALActivityReportObjectiveTopics" where "dml_txid" = '00000000-0000-0000-0000-000002266502'
+            )
+            AND ar."version" = 1
+        );`,
+        { transaction },
+      );
+    });
+  },
+
+  down: async () => {},
+};
--- a/src/migrations/20240104110924-remove-unwanted-aro-topics.js
+++ b/src/migrations/20240104110924-remove-unwanted-aro-topics.js
@@ -9,18 +9,15 @@ module.exports = {
       const sessionSig = __filename;
       await prepMigration(queryInterface, transaction, sessionSig);
       await queryInterface.sequelize.query(
-        `DELETE FROM "ActivityReportObjectiveTopics"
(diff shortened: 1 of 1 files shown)


src/migrations/20240104110924-remove-unwanted-aro-topics.js
```diff
@@ -0,0 +1,30 @@
...
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, _Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      await queryInterface.sequelize.query(
+        `DELETE FROM "ActivityReportObjectiveTopics"
+        WHERE "activityReportObjectiveId" IN (
```
hardwarehuman, line 13: Setting aside logic, I think we'd want to write it like this:
```
DELETE FROM  "ActivityReportObjectiveTopics" arot
USING "ActivityReportObjectives" aro
JOIN "ZALActivityReportObjectiveTopics" zaro
  ON dml_txid = '00000000-0000-0000-0000-000002266502'
  AND aro.id = (zaro.new_row_data->>'activityReportObjectiveId')::int
JOIN "ActivityReports" ar
  ON aro."activityReportId" = ar.id
  AND ar.version = 1
WHERE arot."activityReportObjectiveId" = aro.id
  AND data_id = arot.id
```
There are slight differences in the results because 10 of the ActivityReportObjectiveTopics (AROTs) were actually added later to AROs that were created during the migration. If we want to just delete _all_ AROTs associated with the AROs created in the initial migration, even if the topics were associated later, then we'd need a CTE, but I'm not sure if we do want to do that.

hardwarehuman (changes requested):
We should either slightly modify the description of what we want to say that we want to delete all AROTs attached to AROs created in the initial migration, or we should switch to the suggested SQL that is more narrowly targeted only to AROTs created specifically in that migration.

src/migrations/20240104110924-remove-unwanted-aro-topics.js
hardwarehuman: Per our discussion, we truly don't want _any_ AROTs for those AROs, so the correct logic is:
```
DELETE FROM  "ActivityReportObjectiveTopics" arot
USING "ActivityReportObjectives" aro
JOIN "ZALActivityReportObjectiveTopics" zaro
  ON dml_txid = '00000000-0000-0000-0000-000002266502'
  AND aro.id = (zaro.new_row_data->>'activityReportObjectiveId')::int
JOIN "ActivityReports" ar
  ON aro."activityReportId" = ar.id
  AND ar.version = 1
WHERE arot."activityReportObjectiveId" = aro.id
```

hardwarehuman (approved):

AdamAdHocTeam (commented):
@GarrettEHill is there any concern that some of the created AROT are supposed to exist. For example, if a report was created and not approved before the conversion date. They might have assumed it will be a v2 going forward (and have AROT).

GarrettEHill (approved):

PR #1911, opened 2024-01-12
[TTAHUB-2331] TR goal creates goal on RTR

nvms:
## Description of change

- Add new join table `EventReportPilotGoals` to track goals that have been created from a TR. Stores `goalId`, `eventId`, `sessionId` and `grantId`.
- `SessionReport` hooks:
  - `createGoalsForSessionRecipientsIfNecessary` (`afterCreate` & `afterUpdate`) - does what you probably think it does.
  - `destroyAssociatedGoals` (`beforeDestroy`)
- `EventReport` hooks:
  - `updateGoalText` (`afterUpdate`) - if Goal.name changes, updates Goal.name for all goals created from the TR, unless the session is complete.
- `getGoalsByActivityRecipient` has some logic to exclude returning a Goal when it is associated with a session that is not yet "Complete".

## How to test

This query will be helpful for validating each of these steps:

```sql
SELECT e.id AS "ERPG id",
       erp.data->>'goal' as "EventReportPilot.data->>'goal'",
       g.id AS "Goal.id",
       g.name AS "Goal.name",
       g."grantId" AS "Goal.grantId",
       g.status AS "Goal.status",
       g."previousStatus" AS "Goal.previousStatus",
       s.id AS "SessionReportPilot.id",
       s.data->>'status' AS "SessionReportPilot.data->>'status'"
FROM "EventReportPilotGoals" as e
LEFT JOIN "EventReportPilots" as erp ON e."eventId" = erp.id
LEFT JOIN "Goals" as g ON e."goalId" = g.id
LEFT JOIN "SessionReportPilots" as s ON e."sessionId" = s.id
WHERE e."goalId" IS NOT NULL;
```

- Create an event with text in the "Event goal" field.
- Ensure no goal was created (because no grantId to reference yet).
- Create a session for this event by adding one recipient.
- Run query, ensure a Goal was created and is in Draft and does not appear on the RTR.
- Edit session, add another recipient. Ensure another Goal is created and is also Draft.
- Change the goal text on the event page, run query, ensure "Goal.name" is also updated to match the "event data.goal" text.
- Complete the session, run query, ensure status is now "In Progress" and:
  - Changes to "Event goal" do not change "Goal.name" any longer.
  - The goal appears on the RTR and can be edited.
  - Creating a new session with new recipients will create a Goal with "In Progress" status by default.
- Creating a new session with the same recipient(s) as another session **does not create another Goal for that recipient**.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2331

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/models/hooks/sessionReportPilot.js
```diff
@@ -130,15 +131,74 @@ const participantsAndNextStepsComplete = async (sequelize, instance, options) =>
...
+        if (!grant) throw new Error('Grant not found');
+
+        const newGoal = await sequelize.models.Goal.create({
+          name: event.data.goal,
+          grantId,
+          createdAt: new Date(),
+          updatedAt: new Date(),
+          status: 'Not Started',
+          createdVia: 'tr',
+        }, { transaction: options.transaction });
```
GarrettEHill, line 168: I think you might need to check to see if a goal for that grant with the same text exists, when it does use that

src/models/hooks/eventReportPilot.js
```diff
@@ -98,11 +99,37 @@ const notifyVisionAndGoalComplete = async (_sequelize, instance) => {
...
+  const data = JSON.parse(instance.data.val);
+  const { goal, goals } = data;
+
+  if (!goal || !goals.length) {
+    return;
+  }
+
+  // eslint-disable-next-line no-restricted-syntax
+  for await (const g of goals) {
+    const foundGoal = await sequelize.models.Goal.findByPk(g.goalId, { transaction: instance.transaction });
```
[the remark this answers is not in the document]
nvms: 33791cf4e94fb0d360356e689fde71b130d9dd02

src/models/hooks/eventReportPilot.js
```diff
@@ -98,11 +99,37 @@ const notifyVisionAndGoalComplete = async (_sequelize, instance) => {
...
+    foundGoal.name = goal;
+    await sequelize.models.Goal.update({ name: goal }, { where: { id: g.goalId }, transaction: instance.transaction });
```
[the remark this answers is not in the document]
nvms: Nice catch - dc3061bfaaff6f5e3f5cb6e8fd1a3e62d41e777a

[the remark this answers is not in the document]
nvms: > I dont see how we are preventing the goal from changing once the event has a complete session.

e0ccacf72e5a8175b89dd0a811d3813b2750c3fb

Can likely just update the query in this commit to exclude where `onApprovedAR`.

nvms:
> @nvms is there any plan with this PR to _link_ the goal and the TR?
> 
> The mockup for the RTR for the frontend part of this feature looks like this:
> 
> Screenshot 2024-01-12 at 10 53 55 AM
> 
> (i.e. a new type of objective under a goal)
> 
> I could alter the query that feeds the RTR to go like Goal > Training Report > Sessions.data.objective pretty easily if we had that linkage

I think we talked about this in Slack, but yeah, the PR now has a junction table that can be queried to solve your usecase

thewatermethod:
@nvms lmk if/when this is ready for review

src/services/recipient.js
```diff
@@ -480,15 +480,32 @@ export async function getGoalsByActivityRecipient(
...
+
+  const incompleteSessionGoalIds = incompleteSessionGoals[0].map((j) => j.goalId);
+
   // Goal where.
   let goalWhere = {
     [Op.or]: [
       { onApprovedAR: true },
       { isFromSmartsheetTtaPlan: true },
-      { createdVia: ['rtr', 'admin', 'merge'] },
+      { createdVia: ['rtr', 'admin', 'merge', 'tr'] },
```
thewatermethod, line 501: Would something like this work:

```js
{ createdVia: ['rtr', 'admin', 'merge'] },
{ 
  [Op.and]: {
    createdVia: ['rtr', 'tr'],
    status: {
      [Op.not]: 'Draft',
    } 
  },
}
 ```
 instead of that SQL query above?

File: src/models/hooks/eventReportPilot.js [hunk 5]
```diff
@@ -98,6 +100,76 @@ const notifyVisionAndGoalComplete = async (_sequelize, instance) => {
   }
 };
 
+/**
+ * This hook updates `Goal.name` for all goals that are associated with this training report.
+ */
+const updateGoalText = async (sequelize, instance, options) => {
```
thewatermethod, line 106: should this create a goal if one doesn't exist? given the majority of our events will already exist when this is deployed

thewatermethod:
> Create an event with text in the "Event goal" field.

Events are imported; looking at the last set of [imported prod data](https://ocio-jira.acf.hhs.gov/browse/TTAHUB-1978), none of the imported events had goals. Can we rely on that field being present in the future?

src/models/hooks/eventReportPilot.js
thewatermethod: majority of our sessions, I mean

thewatermethod:
https://github.com/HHS/Head-Start-TTADP/assets/3288586/11a39327-dfaf-43f5-bc3f-764cc8b90494

Found a bug, I think ^
Can't seem to edit goal name on an in progress event with a completed session

thewatermethod:
Everything looks pretty good - just wondering about how we are going to handle completed events or events where there will be no more sessions

src/models/hooks/eventReportPilot.js
```diff
@@ -98,6 +100,76 @@ const notifyVisionAndGoalComplete = async (_sequelize, instance) => {
...
+    return;
+  }
+
+  if (current.goal === previous.goal) {
+    return;
+  }
+
+  // Disallow goal name propagation if any session on this event has been completed,
+  // effectively locking down this goal text.
+  // We should probably disable this in the UI too.
```
thewatermethod, line 128: https://github.com/HHS/Head-Start-TTADP/pull/1912

src/models/hooks/eventReportPilot.js
nvms: 1336c303bcf44aee6425e7da06252deb64f859d3

nvms:
> events where there will be no more sessions

When an event's goal text is updated, a goal will be created for all event sessions. For cases where the event goal text will not change, maybe we just do a migration (future PR maybe?) if that's the behavior they want

thewatermethod:
> Should goals be destroyed in this case? Marked in progress? Not sure what makes most sense here aside from just not allowing the event to be marked as completed.

@nvms Yeah this is my bad I think. I guess I didn't understand that a goal would be created if the text changed, I thought when I looked at this before goals were only being created when a session was created. So the case I was worried about is invalid.

> When an event's goal text is updated, a goal will be created for all event sessions. For cases where the event goal text will not change, maybe we just do a migration (future PR maybe?) if that's the behavior they want

Probably just want to be clear about what we're delivering then, I don't think this lines up with customer expectations that the events already completed (there's at least 1) won't have goals

src/models/hooks/sessionReportPilot.js
```diff
@@ -130,15 +131,179 @@ const participantsAndNextStepsComplete = async (sequelize, instance, options) =>
...
+    const goals = await EventReportPilotGoal.findAll({
+      where: {
+        sessionId: instance.id,
+        eventId: instance.eventId,
+      },
+      transaction: options.transaction,
+    });
+
+    // eslint-disable-next-line no-restricted-syntax
+    for await (const goal of goals) {
```
thewatermethod, line 225: I don't think we should destroy Goals here? Won't this destroy goals for other sessions, or goals that could have been used on an AR? So maybe just EventReportPilotGoals. Unless I'm missing something.

thewatermethod (commented):
I think this looks good, seems to work pretty well - I had a question about the delete hooks, and I think we should make sure this satisfies the requirements as it elides completed training reports, but I'm ready to approve otherwise.

nvms:
> I think this looks good, seems to work pretty well - I had a question about the delete hooks, and I think we should make sure this satisfies the requirements as it elides completed training reports, but I'm ready to approve otherwise.

I removed the hook to destroy goals when the event is deleted - not part of AC anyways, and probably not the best thing to do anyways.

447b89b8dfa941ea73ccbf921314feef237417ca

nvms:
> I guess I didn't understand that a goal would be created if the text changed, I thought when I looked at this before goals were only being created when a session was created. So the case I was worried about is invalid.

Nope you were right before, that was the behavior. I took your suggestion to create goals during the `updateGoalText` hook and implemented it in 1336c303bcf44aee6425e7da06252deb64f859d3

> Probably just want to be clear about what we're delivering then, I don't think this lines up with customer expectations that the events already completed (there's at least 1) won't have goals

💯 I can explain that a migration will be pushed out for this in a future PR. I think it does make sense to include it in this PR, but I'm also very aware that this PR is blocking a number of others. 🙃

thewatermethod (approved):

PR #1915, opened 2024-01-16
[TTAHUB-2334] Patch incorrect state code for Palau

kryswisnaskas:
## Description of change
Updates grant ids 10291 and 14869 with the correct state ("PW")

## How to test
Make sure the tests pass

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2334

### Every PR

- [x] Meets issue criteria
- [n/a] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/lib/updateGrantsRecipients.js
+++ b/src/lib/updateGrantsRecipients.js
@@ -20,7 +20,7 @@ const fs = require('mz/fs');
 
 // TTAHUB-2126 TTAHUB-2334
 // Update the specific attribute (e.g., state code) for each grant,
-// identified by its grant number, in the map below. 
+// identified by its grant number, in the map below.
 // This patch sets the new value for the grant to ensure data accuracy.
 const grantPatches = new Map([
   [12128, { stateCode: 'OH' }],
--- a/src/lib/updateGrantsRecipients.js
+++ b/src/lib/updateGrantsRecipients.js
@@ -18,7 +18,10 @@ import { GRANT_PERSONNEL_ROLES } from '../constants';
 
 const fs = require('mz/fs');
 
-// TTAHUB-2126
+// TTAHUB-2126 TTAHUB-2334
+// Update the specific attribute (e.g., state code) for each grant,
+// identified by its grant number, in the map below. 
+// This patch sets the new value for the grant to ensure data accuracy.
 const grantPatches = new Map([
   [12128, { stateCode: 'OH' }],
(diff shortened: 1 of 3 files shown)


src/lib/updateGrantsRecipients.js
```diff
@@ -21,6 +21,8 @@ const fs = require('mz/fs');
 // TTAHUB-2126
 const grantPatches = new Map([
   [12128, { stateCode: 'OH' }],
+  [10291, { stateCode: 'PW' }],
+  [14869, { stateCode: 'PW' }],
 ]);
```
GarrettEHill, line 26: Based on the naming convention, I assume the number is a grant ID.  Would it be better to use Recipient Id or any descendent of one of these grants? This current implementation will fix the problem today, but as soon as one of these grants gets replaced it will be an issue again

kryswisnaskas: Yes, you're correct – the number is a grant ID. Thank you for pointing out that this wasn't immediately obvious; I’ve updated the comment to clarify this.

Regarding the choice between Grant and Recipient granularity, there are specific benefits to addressing these changes at the grant level. It allows us to selectively control which grants receive data modifications from HSES. For instance, in this ticket, OHS requested changes for only 2 of the 4 grants. Additionally, these changes are intended to be temporary, with the anticipation that Grant Solutions will update the data over time.

For this particular recipient, a replacement grant isn’t a concern for the time being, as the current active grant is valid until mid 2028. Let's proceed with this approach for now, and if we find a growing need to patch grant numbers, we can then reassess and consider shifting to the Recipient level.

GarrettEHill: I understand that line of reasoning, would making an override table be of use. Then an admin screen could be made to see the current overrides and maintain them going forward. This would remove the need for a migration file or code change to effect specific user data. I just like to push for "Fix it once" type solutions over "Fix it for now" type of solutions.

kryswisnaskas: Thanks for your suggestion about the override table and admin screen. It's great to see your forward-thinking approach. However, for this specific ticket, we need to focus on a simpler, immediate solution that aligns with the current scope and client requirements. 
But hey, let’s not lose your idea – maybe we can circle back to it for a future ticket where we can dive into something more complex.

thewatermethod (approved):

PR #1928, opened 2024-01-22
Exclude deletedAt records from non pkey indexes

hardwarehuman:
## Description of change

When records are marked as deleted using Sequelize's standard `deletedAt` column, collisions in indexes that don't include the table's primary key can occur when attempting to re-add those records. This was primarily being seen in GoalCollaborators but could impact many tables, so they're all addressed here

## How to test

1) Create and approve a report, creating a goal

2) Unlock your newly approved report

3) Approve unlocked report - this would previously generate an error

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2460

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


thewatermethod:
Seeing lots of CI failures that _look_ related

```
SequelizeDatabaseError: there is no unique or exclusion constraint matching the ON CONFLICT specification
```

src/migrations/20240119000000-exclude_deleted_from_indexes.js
```diff
@@ -0,0 +1,52 @@
...
+        ALTER TABLE "GoalCollaborators" DROP CONSTRAINT "GoalCollaborators_goalId_userId_collaboratorTypeId_unique";
+        CREATE UNIQUE INDEX "GoalCollaborators_goalId_userId_collaboratorTypeId_unique"
+        ON "GoalCollaborators" ("goalId", "userId", "collaboratorTypeId")
+        WHERE "deletedAt" IS NULL;
+
+        ALTER TABLE "ObjectiveCollaborators" DROP CONSTRAINT "ObjectiveCollaborators_objectiveId_userId_collaboratorTypeId_un";
+        CREATE UNIQUE INDEX "ObjectiveCollaborators_objectiveId_userId_collaboratorTypeId_un"
+        ON "ObjectiveCollaborators" ("objectiveId", "userId", "collaboratorTypeId")
+        WHERE "deletedAt" IS NULL;
+
```
GarrettEHill, line 40: Topics seems to be missing

GarrettEHill: Maybe Topic's unique constraint is better as it uses deletedAt as part of the unique index instead of a where clause. That allows it to still be used as a constraint.
`CREATE UNIQUE INDEX "Topics_name_key" ON public."Topics" USING btree (name, "deletedAt", "mapsTo")`

GarrettEHill: I think this might be a better way to go. Make the unique index with "deletedAt" as part of the index. Then make a unique constraint using the unique index. It will keep all the original goals and still allow for reuse after delete.

GarrettEHill: make sure to drop the old indexes too

GarrettEHill: Original Indexes:
`	SELECT DISTINCT *
	FROM pg_indexes i
	join information_schema.columns isc
	on i.tablename = isc.table_name
	and isc.column_name = 'deletedAt'
	where tablename not like 'pg_%'
	and indexname not like '%_pkey'
	and indexdef ilike '% UNIQUE %'
	and tablename not like 'Monitoring%';`

Original constraints:
`SELECT
tc.table_name,
  'ALTER TABLE ' || tc.table_schema || '.' || tc.table_name || 
  ' ADD CONSTRAINT ' || tc.constraint_name || 
  ' UNIQUE (' || string_agg(kcu.column_name, ', ') || ');' as create_statement
FROM 
  information_schema.table_constraints AS tc
JOIN 
  information_schema.key_column_usage AS kcu
  ON tc.constraint_name = kcu.constraint_name
join information_schema.columns isc
	on tc.table_name = isc.table_name
	and isc.column_name = 'deletedAt' 
WHERE 
  tc.constraint_type = 'UNIQUE'
GROUP BY 
  tc.table_schema, tc.table_name, tc.constraint_name;`

hardwarehuman: I think which approach is better depends on whether we expect to actually see the deleted records in any queries. If they're always excluded, then a smaller index with only the non-excluded records is usually better.

hardwarehuman: Due to ON CONFLICT with upserts not working with partial indexes, I am switching to including `deletedAt` in the indexes regardless of other considerations

hardwarehuman: Actually, adding `deletedAt` doesn't work either because it means the conflict column specification doesn't match the new constraint. I think we may need to do a legit undelete

hardwarehuman: We could have gone back to using just partial indexes because @thewatermethod removed the upsert, but I think enough folks are waiting on this to where we just want to merge

GarrettEHill (approved):

PR #1934, opened 2024-01-23
[TTAHUB-2299] Add effectiveDate for all designated personnel

hardwarehuman:
## Description of change

This adds an effectiveDate for all assigned Program Personnel. It uses the actual date we set them active rather than the date we added the program personnel record. This is the more conservative approach selected by Angela.

## How to test

Every Recipient TTA page for a recipient that has had leadership personnel reassignments should have email change dates for when the current person was assigned to their role, e.g. https://ttahub.ohs.acf.hhs.gov/recipient-tta-records/1476/region/4/profile

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2299

### Every PR

- [x] Meets issue criteria

### Before merge to main

### Production Deploy

### After merge/deploy


thewatermethod:
I'll be honest, I struggle with reviewing these migration/SQL-only PRs. If at all possible (and only if it's not a big pain), a query that identifies the records that should be updated so I can check before and after would be a big help

src/migrations/20240116000001-add-effectivedate_for_migrated_personnel.js
```diff
@@ -0,0 +1,91 @@
...
+          "programId" pid,
+          "grantId" grid,
+          role
+        FROM "ProgramPersonnel"
+        ),
+        all_personnel_roles AS (
+        SELECT DISTINCT
+          (new_row_data->>'programId')::int zpid,
+          (new_row_data->>'grantId')::int zgrid,
+          new_row_data->>'role' zrole,
```
GarrettEHill, line 40: Is there concern that a record in the ZALProgramPersonel table may not have data in all four columns, as the audit log only has deltas. So if the only change was email or role the other columns might be empty.

kryswisnaskas: At least for emails, we don't seem to have records that don't also include `active:true` (I haven't checked roles).

src/migrations/20240116000001-add-effectivedate_for_migrated_personnel.js
```diff
@@ -0,0 +1,91 @@
...
+  prepMigration,
+} = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, _Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      await queryInterface.sequelize.query(
```
kryswisnaskas, line 11:
```suggestion
      await queryInterface.sequelize.query(/* sql */
```

kryswisnaskas (approved):
Looks good - only a minor suggestion. 👍

src/migrations/20240116000001-add-effectivedate_for_migrated_personnel.js
```diff
@@ -0,0 +1,91 @@
...
+        `
+        -------------
+        -- PROCESS --
```
AdamAdHocTeam, line 14: Might be best to move each comment to its corresponding section. That way we don't have to scroll back up (just a suggestion).

AdamAdHocTeam (commented):
So I looked into two examples that still have a null effective date after running the script. That I would think should be updated (but maybe not).

- Program Personnel Id: 1156134, was brought in as active but still shows null effective date.
- Program Personnel Id: 1151698, was brought in as in-active but was switched to active with a audit row (show below) but still has null effective date.
image

Let me know if you wan to huddle to go over these cases.

src/migrations/20240116000001-add-effectivedate_for_migrated_personnel.js
hardwarehuman: There's no real way to identify the relationship without all four, but it's possible to change the relationship one column at a time, which wasn't a scenario I'd considered and should be handled

src/migrations/20240116000001-add-effectivedate_for_migrated_personnel.js
hardwarehuman: I put a copy of the number header throughout the logic

src/migrations/20240116000001-add-effectivedate_for_migrated_personnel.js
hardwarehuman: Getting this right is a pain so I instead added a protective query at the beginning that will cause a failure if the relationship values get updated in a way that would break the logic.

hardwarehuman:
> So I looked into two examples that still have a null effective date after running the script. I would think these should be updated (but maybe not).
> 
> * Program Personnel Id: 1156134, was brought in as active but still shows null effective date.
> * Program Personnel Id: 1151698, was brought in as in-active but was switched to active with a audit row (show below) but still has null effective date.
> Let me know if you wan to huddle to go over these cases.

The issue was that my logic assumed that if a new person was assigned to a role, they replaced the previous person assigned to the role, but this was incorrect; multiple people can be validly assigned to the same role at the same time and thus have valid effectiveDate values. The logic is now changed to set effectiveDate to the last activation date for all current assignments.

thewatermethod (approved):
Looks good, will need to merge main in again as the migration has fallen out of order

src/migrations/20240301000000-add-effectivedate_for_migrated_personnel.js
```diff
@@ -0,0 +1,100 @@
...
+        SELECT *
+        FROM program_personnel_association_updates
+        WHERE email_cnt > 1
+          OR program_cnt > 1
+          OR grant_cnt > 1
+          OR role_cnt > 1
+        ;
+        -- This will cause a divide by zero error that fails the transaction if invalid_test is not empty
+        SELECT
+          1/(LEAST(COUNT(*),1) - 1)
```
hardwarehuman, line 52:
```suggestion
          1/(GREATEST(COUNT(*),1) - 1)
```

PR #1943, opened 2024-01-25
[TTAHUB-2494] Fix resource links reducer error

thewatermethod:
## Description of change
Another error introduced by my recent refactor; resource links reduced for an objective don't have a "id" but instead use a value

## How to test
Add two resource links to an activity report objective, leave, and then return. Both resource links should be present.

## Issue(s)
* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2494

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/goalServices/goals.js
+++ b/src/goalServices/goals.js
@@ -476,6 +476,7 @@ const reduceRelationThroughActivityReportObjectives = (
   join,
   relation,
   exists = {},
+  uniqueBy = 'id',
 ) => {
   const existingRelation = exists[relation] || [];
   return uniqBy([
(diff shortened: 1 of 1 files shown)


src/goalServices/goals.js
```diff
@@ -518,6 +519,7 @@ export function reduceObjectivesForActivityReport(newObjectives, currentObjectiv
         'activityReportObjectiveResources',
         'resource',
         exists,
+        'value',
```
thewatermethod, line 522: A resource, for reasons that are obscure to me, has it's id renamed as value in the SQL query (I think this is also my fault), so this reducer was comparing undefined to undefined and therefore not returning all the resources.

This change allows us to configure which key we de-duplicate by

AdamAdHocTeam (approved):
Looks good would wait for CI to pass.

PR #1944, opened 2024-01-25
[TTAHUB-2494] Fix for activity report approvers

thewatermethod:
## Description of change
We (I) removed the upserts from the approval process (big mistake?)
In doing so, there was one straight up mistake and one thing that we missed.

## How to test
1/25/24 prod dataset is on keybase. Impersonate the user listed in the Jira ticket and submit report 39409.
Clear local storage; refresh the page

You will see that the report is **actually submitted**

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2494

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/models/hooks/activityReportApprover.js
+++ b/src/models/hooks/activityReportApprover.js
@@ -41,6 +41,12 @@ const updateReportStatus = async (sequelize, instance) => {
       submissionStatus: REPORT_STATUSES.SUBMITTED,
     },
   });
+
+  // we only update any of the report status fields if the report is submitted
+  if (!report) {
+    return;
+  }
+
   // We allow users to create approvers before submitting the report. Calculated
   // status should only exist for submitted reports.
 
--- a/src/policies/activityReport.js
+++ b/src/policies/activityReport.js
@@ -161,8 +161,15 @@ export default class ActivityReport {
     return approverUserIds.includes(this.user.id);
   }
 
+  // This is a helper function to determine if the report is in a state where it can be edited
   reportHasEditableStatus() {
+    // if the report is in draft
     return this.activityReport.submissionStatus === REPORT_STATUSES.DRAFT
-      || this.activityReport.calculatedStatus === REPORT_STATUSES.NEEDS_ACTION;
+    // or if it's been marked as needs action
+      || this.activityReport.calculatedStatus === REPORT_STATUSES.NEEDS_ACTION || (
+    // or if it's submitted and the calculated status is not submitted for some reason
+      this.activityReport.submissionStatus === REPORT_STATUSES.SUBMITTED
+        && this.activityReport.calculatedStatus !== REPORT_STATUSES.SUBMITTED
+    );
   }
 }
--- a/src/policies/activityReport.js
+++ b/src/policies/activityReport.js
@@ -163,13 +163,18 @@ export default class ActivityReport {
 
   // This is a helper function to determine if the report is in a state where it can be edited
   reportHasEditableStatus() {
-    // if the report is in draft
-    return this.activityReport.submissionStatus === REPORT_STATUSES.DRAFT
-    // or if it's been marked as needs action
-      || this.activityReport.calculatedStatus === REPORT_STATUSES.NEEDS_ACTION || (
-    // or if it's submitted and the calculated status is not submitted for some reason
-      this.activityReport.submissionStatus === REPORT_STATUSES.SUBMITTED
-        && this.activityReport.calculatedStatus !== REPORT_STATUSES.SUBMITTED
-    );
+    // if the report is in draft, it's editable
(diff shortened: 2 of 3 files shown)


src/models/hooks/activityReportApprover.js
```diff
@@ -41,6 +41,12 @@ const updateReportStatus = async (sequelize, instance) => {
       submissionStatus: REPORT_STATUSES.SUBMITTED,
     },
   });
+
```
thewatermethod, line 44: if you look at the query above, I changed it from 
```js
  const report = await sequelize.models.ActivityReport.findByPk(instance.activityReportId, {
    attributes: ['submissionStatus', 'calculatedStatus'],
  });
```
to
```js
  const report = await sequelize.models.ActivityReport.findByPk(instance.activityReportId, {
    attributes: ['submissionStatus', 'calculatedStatus'],
    where: {
      submissionStatus: REPORT_STATUSES.SUBMITTED,
    },
  });
```
in order to be a little more succinct
However, it should bomb out early as before the rest of the function was locked in a 
```js
if(report.submittedStatus === REPORT_STATUSES.SUBMITTED){
/// stuff
}
```

and we don't want to update a report that hasn't been submitted by a user. I don't think this is causing a bug but it is how it was working before and I don't want to introduce any new ones if I can help it

src/policies/activityReport.js
```diff
@@ -161,8 +161,15 @@ export default class ActivityReport {
...
+  // This is a helper function to determine if the report is in a state where it can be edited
   reportHasEditableStatus() {
+    // if the report is in draft
     return this.activityReport.submissionStatus === REPORT_STATUSES.DRAFT
-      || this.activityReport.calculatedStatus === REPORT_STATUSES.NEEDS_ACTION;
+    // or if it's been marked as needs action
+      || this.activityReport.calculatedStatus === REPORT_STATUSES.NEEDS_ACTION || (
+    // or if it's submitted and the calculated status is not submitted for some reason
+      this.activityReport.submissionStatus === REPORT_STATUSES.SUBMITTED
+        && this.activityReport.calculatedStatus !== REPORT_STATUSES.SUBMITTED
```
thewatermethod, line 172: This is essentially damage control - the reports affected by this bug can't be edited if we don't add this little bit of business here. I can't see how it hurts anything - the report stays editable until the hooks have moved the report to "actually submitted" (see calculatedStatus) which is what we check on the frontend as well

src/services/activityReportApprovers.js
```diff
@@ -16,11 +16,22 @@ export async function upsertApprover(values) {
     paranoid: false,
   });
 
-  if (approver && status) {
-    await approver.update({
-      status,
-      note: values.note,
-    }, {
+  if (approver) {
```
thewatermethod, line 19: this is the actual bug. Newly submitted reports don't have an approver status. So if you
1) added an approver
2) saved
3) submitted report

You would trigger this, as the approver hooks would not refire. With this bugfix, everytime we call syncApprovers, they will ALWAYS fire

src/services/activityReportApprovers.js
```diff
@@ -56,7 +67,7 @@ export async function upsertApprover(values) {
  * @param {*} userIds - array of userIds for approver records, ActivityReportApprovers will be
  * deleted or created to match this list
  */
-export async function syncApprovers(activityReportId, userIds = [], log = false) {
```
thewatermethod, line 59: This is simply a bit of code I should have removed yesterday

src/policies/activityReport.js
```diff
@@ -161,8 +161,20 @@ export default class ActivityReport {
     return approverUserIds.includes(this.user.id);
   }
 
+  // This is a helper function to determine if the report is in a state where it can be edited
   reportHasEditableStatus() {
-    return this.activityReport.submissionStatus === REPORT_STATUSES.DRAFT
-      || this.activityReport.calculatedStatus === REPORT_STATUSES.NEEDS_ACTION;
+    // if the report is in draft, it's editable
+    if (this.activityReport.submissionStatus === REPORT_STATUSES.DRAFT
+      || this.activityReport.calculatedStatus === REPORT_STATUSES.DRAFT) {
```
thewatermethod, line 168: This is essentially damage control - the reports affected by this bug can't be edited if we don't add this little bit of business here. I can't see how it hurts anything - the report stays editable until the hooks have moved the report to "actually submitted" (see calculatedStatus) which is what we check on the frontend as well

src/services/activityReportApprovers.js
thewatermethod: the bug being the report would then not be moved to a calculatedStatus of "submitted"

GarrettEHill (approved):

PR #1977, opened 2024-02-09
[TTAHUB-2463] Delete comlog entry per user request

hardwarehuman:
## Description of change

Deletes a communications log record and its link to files, then also deletes the files if they are now unlinked.
There's not actually any files linked to this comlog entry, but the logic is there so that if we reuse it, it does everything correctly.

## How to test

There are before and after temp tables that illustrate the stats, though the basic thing to watch for in this case is that in https://ttahub.ohs.acf.hhs.gov/recipient-tta-records/1230/region/3/communication? the "TTA planning or scheduling" from Jan 8th should be gone. You can also run the SQL and check the before and after temp tables.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2463

### Every PR

- [ ] Meets issue criteria
- 
### Before merge to main

### Production Deploy

### After merge/deploy


thewatermethod (approved):
Looks/works good! Nice work making this script repeatable and understandable.

Let's merge main into this to fix the CI errors before moving it forward!

PR #2010, opened 2024-03-06
[TTAHUB-2661] Monitoring issue found when using bulkCreate

GarrettEHill:
## Description of change
There is an issue in sequelize bulkCreate where it does not update the sequence used for the primary key. This prevents additional inserts into the table as any insert will have a key conflict.

## How to test
Run monitoring, add a file to the sftp, run monitoring again. The new file should be handle correctly.

## Issue(s)

* https://ocio-jira.acf.hhs.gov/browse/TTAHUB-2661

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20240306000000-monitoring-nonbulkcreate.js
+++ b/src/migrations/20240306000000-monitoring-nonbulkcreate.js
@@ -1,0 +1,17 @@
+const {
+  prepMigration,
+} = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      // await prepMigration(queryInterface, transaction, sessionSig);
+      // eslint-disable-next-line @typescript-eslint/quotes
+      await queryInterface.sequelize.query(/* sql */`SELECT setval('"ImportFiles_id_seq"', COALESCE((SELECT MAX(id)+1 FROM "ImportFiles"), 1), false);`, { transaction });
+      // eslint-disable-next-line @typescript-eslint/quotes
+      await queryInterface.sequelize.query(/* sql */`SELECT setval('"ImportDataFiles_id_seq"', COALESCE((SELECT MAX(id)+1 FROM "ImportDataFiles"), 1), false);`, { transaction });
+    });
+  },
+};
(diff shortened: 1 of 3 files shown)


src/migrations/20240306000000-monitoring-nonbulkcreate.js
```diff
@@ -0,0 +1,17 @@
...
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      // await prepMigration(queryInterface, transaction, sessionSig);
+      // eslint-disable-next-line @typescript-eslint/quotes
+      await queryInterface.sequelize.query(/* sql */`SELECT setval('"ImportFiles_id_seq"', COALESCE((SELECT MAX(id)+1 FROM "ImportFiles"), 1), false);`, { transaction });
+      // eslint-disable-next-line @typescript-eslint/quotes
+      await queryInterface.sequelize.query(/* sql */`SELECT setval('"ImportDataFiles_id_seq"', COALESCE((SELECT MAX(id)+1 FROM "ImportDataFiles"), 1), false);`, { transaction });
+    });
+  },
+};
```
kryswisnaskas, line 17: Can we also include a down migration?

kryswisnaskas (commented):
It looks good. A couple of comments/questions.

Are there any concerns about the performance now considering that we are switching to "create"?

I also added a comment on the migration. 🌮

src/migrations/20240306000000-monitoring-nonbulkcreate.js
GarrettEHill: the migration is just the fix the issue with the current state where the value in the id column and the sequence do not align. there is nothing to do in a down

GarrettEHill:
> It looks good. A couple of comments/questions.
> 
> Are there any concerns about the performance now considering that we are switching to "create"?
> 
> I also added a comment on the migration. 🌮

Performance will be a little less, but this was to create the records for the resistance of files (both the zips and the files in the zips) which is not a huge amount of records. so the delta in performance should not be an issue.

src/migrations/20240306000000-monitoring-nonbulkcreate.js
kryswisnaskas: An empty down migration is fine.

kryswisnaskas: If we don't include anything and need to perform a rollback, the SequelizeMeta table would become out of sync.

GarrettEHill: ok i will put in an empty down

kryswisnaskas (approved):

thewatermethod (approved):

PR #2027, opened 2024-03-14
[TTAHUB-2703] delete test communication log

hardwarehuman:
## Description of change

A customer created a couple 'test' communication logs to demo the feature and wants them removed. There's no `CommunicationLogs.deletedAt` so it's a hard delete, and there's currently a lot of different ways to connect files to things so it's a little more complicated than it would otherwise be.

## How to test

Neither https://ttahub.ohs.acf.hhs.gov/recipient-tta-records/1839/region/10/communication/888/view
nor https://ttahub.ohs.acf.hhs.gov/recipient-tta-records/1839/region/10/communication/788/view should turn up anything, and the comlog page for that recipient should be clear of the test comlogs. You can also reference the `before_count` and `after_count` temp tables to see the removals.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2703

### Every PR

- [ ] Meets issue criteria

### Before merge to main

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20240313000002-delete_comlog_entry.js
+++ b/src/migrations/20240313000002-delete_comlog_entry.js
@@ -1,0 +1,130 @@
+const {
+  prepMigration,
+} = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, _Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      await queryInterface.sequelize.query(
+        `
+        -- Delete communications log records as desired by a customer.
+        -- Because there's no soft delete column (deletedAt), we're:
+        
+        -- Doing a hard DELETE FROM in the communications log and link tables (currently just CommunicationLogFiles)
+        -- Removing any Files that only exist due to the Communication Log.
+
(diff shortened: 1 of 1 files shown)


src/migrations/20240313000002-delete_comlog_entry.js
```diff
@@ -0,0 +1,130 @@
...
+        FROM before_set;
+
+        -- Check to see if any files are only present because of this communication log
+        -- unfortunately there's currently many ways to link a file
+        DROP TABLE IF EXISTS files_to_delete;
+        CREATE TEMP TABLE files_to_delete
+        AS
+        WITH candidate_files AS (SELECT DISTINCT fid FROM before_set),
+        file_links AS (
+        SELECT
```
GarrettEHill, line 52: Because this is getting intensive, should we make a helper function that check all tables with a foreign key to the files table base on a supplied list. That way we know they will all always be checked and properly handled?

GarrettEHill (approved):

src/migrations/20240313000002-delete_comlog_entry.js
hardwarehuman: I guess we could do that if we're not going to do the generic refactor.

GarrettEHill: refactor does not look to be happening any time soon

PR #2040, opened 2024-03-18
[TTAHUB- 2650] Fix session delete

kryswisnaskas:
## Description of change
The session deletion feature was broken when there was a goal associated with the event. In this case, the event/goal link table also included a session ID. This linkage generated a foreign key constraint error when attempting to delete a session. The fix involves allowing null values in the sessionId field of the link table, as we still need to retain the records there.

## How to test

- Make sure there are recipients of an event.
- Add a goal to the event. 
- Add a session to that event.
- Verify that the session can be deleted.
- Change the goal's text.
- Confirm that the goal was updated for the correct recipients.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2650

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


thewatermethod:
1. Added a goal to an event
2. Added a session, choose a recipient
3. Deleted session
4. Confirmed sessionId is set to null
5. Add another session

This is what I see:
```SELECT * FROM "EventReportPilotGoals"```
Screenshot 2024-03-18 at 4 08 47 PM

Do we want to keep these null'ed records around? They won't show up in the UI anywhere AFAIK

nvms (approved):
I manually removed EventReportPilotGoals where sessionId is null just to make sure nothing unexpected happens afterwards, like when creating a new session or updating goal text. I think we can clean these up like @thewatermethod suggests without any side effects

I recall when we talked about this the other day, the decision was made to make it nullable instead of deleting the row, but for the life of me I can't remember what exactly the issue was. 🙃 

At any rate, this works well and solves the issue!

kryswisnaskas:
> I recall when we talked about this the other day, the decision was made to make it nullable instead of deleting the row, but for the life of me I can't remember what exactly the issue was. 🙃

Absolutely, we opted for making it nullable to preserve the hook that updates goal's text based on event and goal associations. Deleting the row would interrupt this mechanism. Let me know if you need more information!

PR #2043, opened 2024-03-19
[TTTAHUB-2736] Update goal similarity api to use region id

AdamAdHocTeam:
## Description of change

The goal similarity API is not taking into account the region id when getting similar goals. This changes passes the region id to the python endpoint to ensure suggested matches are for the same region.

**Note:** Not sure how we want to handle these if they already exist in the database. We can do a one off delete for these recipients in particular.

## How to test

- Review the code.
- Using a prod db edit a multi region recipient from the RTR (see jira)
- Make sure the list of suggested merge goals are all for the same region.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2736

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/similarity_api/src/routes/create.py
+++ b/similarity_api/src/routes/create.py
@@ -28,16 +28,22 @@ def create_routes(app: Flask):
       cluster = data["cluster"] if "cluster" in data else False
       include_curated_templates = data["include_curated_templates"] if "include_curated_templates" in data else False
 
+      regionId = data["regionId"] if "regionId" in data else None
+
+      ## If region is not specified, then return error.
+      if regionId is None:
+            return jsonify({"error": "regionId not provided"}), 400
+
       if "text" in data and "recipient_id" in data:
           recipient_id = data["recipient_id"]
           text = data["text"]
-          return find_similar_goals(recipient_id, text, alpha, include_curated_templates)
+          return find_similar_goals(recipient_id, text, alpha, regionId, include_curated_templates)
       elif "text_1" in data and "text_2" in data:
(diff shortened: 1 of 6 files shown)


thewatermethod (changes requested):
I'll ping you in the morning about this

thewatermethod:
Unfortunately, this actually only moves the problem, it doesn't fix it. Remember that the results from the API are stored in our database. Since this doesn't store the region ID, it means that for a recipient with multiple regions, the first region that is visited will have results generated. 

For example, if recipient 15 has region 1 and region 3, and the first page visited is /recipient-tta-records/15/region/1/rttapa, the similarity API will return results for region 1 and save them with that recipient ID. Then, when /recipient-tta-records/15/region/3/rttapa is visited, the same results will be displayed, since the GoalSimilarityGroups do not record a region ID. And when clicked on, they'll be blank, as reported in the original support case.

Suggested fix would be to discard this PR and modify the query that returns the goal groups to filter by region ID. Even better would be to build upon this PR and record region ID in the GoalSimilarityGroups table, which would cut down on the churning of data in this table. 

FYI: there is a version column on the GoalSimiliarityGroups table. If you increment it in the code, you won't need to do any DB cleanup migration as they will stop being returned to the frontend and will be cleaned up over time.

AdamAdHocTeam:
This will be fixed via mb/TTAHUB-2736/alternate-solution as this approach still has the old data in the db.

PR #2044, opened 2024-03-19
[TTAHUB-2570] Write Resource Dashboard to use SQL temp tables

AdamAdHocTeam:
## Description of change

This is the first step to caching a flat data table to filter on for the resource dashboard. As of now this seems to speed up the query by 40% and hopefully will allow for multiple user to use the page simultaneously.

**NOTE:** Don't forget to disable the cache in (cache.ts) although it will be set to the default when live.

*** As of right now we **only** care about resources on the appropriate objective fields. Some filters might be off as we include all source field types for resources.

## How to test

- Review the code.
- Attempt to load the resource dashboard with no filters (should see a reasonable time not 10s).
- Add filters for a date range and make sure the results make sense.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2515

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+      SELECT
+      d.url,
+      d.title,
+      to_char(s."date", 'Mon-YY') AS "rollUpDate",
+      s."date",
+      coalesce(u."resourceCount", 0) AS "resourceCount",
+      t."totalCount"
+      FROM distincturls d
+      JOIN series s
+        ON 1=1
```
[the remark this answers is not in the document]
GarrettEHill: cross join should also be last based on how the other joins are working to reduce the effort of the other joins, this should speed it up a little

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    )
+    SELECT
+      d.name,
+      to_char(s."date", 'Mon-YY') AS "rollUpDate",
+      s."date",
+      coalesce(t."resourceCount", 0) AS "resourceCount",
+      tt."totalCount"
+    FROM ${tblNames.createdTopicsTempTableName} d
+    JOIN series s
+      ON 1=1
```
GarrettEHill, line 539: User cross join. cross join should also be last based on how the other joins are working to reduce the effort of the other joins, this should speed it up a little

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    {
+      type: QueryTypes.SELECT,
+      transaction,
+    },
+  );
+}
+
+function getOverview(tblNames, totalReportCount, transaction) {
+  // - Number of Participants -
+  const numberOfParticipants = sequelize.query(`
```
GarrettEHill, line 556: please add the /* sql */ at the start of the query string to allow for syntax highlighting to make the code easier to read.

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+  WITH ar_participants AS (
+    SELECT
+    id,
+    "numberOfParticipants"
+    FROM ${tblNames.createdFlatResourceTempTableName} f
+    GROUP BY id, "numberOfParticipants"
```
GarrettEHill, line 562: Please use DISTINCT instead of group by, as it should be faster

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+  )
+  SELECT
+         SUM("numberOfParticipants") AS participants
+  FROM ar_participants;
+  `, {
+    type: QueryTypes.SELECT,
+    transaction,
+  });
+
+  const numberOfRecipSql = `
```
GarrettEHill, line 572: please add the /* sql */ at the start of the query string to allow for syntax highlighting to make the code easier to read.

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    JOIN "ActivityRecipients" arr
+      ON ar.id = arr."activityReportId"
+    JOIN "Grants" g
+      ON arr."grantId" = g.id AND g."status" = 'Active'
+     JOIN "Recipients" r
+      ON g."recipientId" = r.id
+  )
+  SELECT
+         count(r.id) AS recipients
+  FROM recipients r;
```
GarrettEHill, line 590:
```suggestion
  SELECT
         count(DISTINCT r."recipientId") AS recipients
  FROM ${tblNames.createdFlatResourceTempTableName} ar
  JOIN "ActivityRecipients" arr
  ON ar.id = arr."activityReportId"
  JOIN "Grants" g
  ON arr."grantId" = g.id AND g."status" = 'Active';
```

GarrettEHill: Why is the status active filtered here as the grants were active at the time of the report, even if they may be inactive now.

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+  `;
+
+  // - Number of Recipients -
+  const numberOfRecipients = sequelize.query(numberOfRecipSql, {
+    type: QueryTypes.SELECT,
+    transaction,
+  });
+
+  // - Reports with Resources Pct -
+  const pctOfResourcesSql = `
```
GarrettEHill, line 600: please add the /* sql */ at the start of the query string to allow for syntax highlighting to make the code easier to read.

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+  SELECT
+    count(DISTINCT "activityReportId")::decimal AS "reportsWithResourcesCount",
```
GarrettEHill, line 602: why are you casting the ID to a decimal?

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    ${totalReportCount}::decimal AS "totalReportsCount",
```
GarrettEHill, line 603: why are you casting to a decimal?

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    CASE WHEN ${totalReportCount} = 0 THEN
+      0
+    ELSE
+      (round(count(DISTINCT "activityReportId")::decimal / ${totalReportCount}::decimal, 4) * 100)::decimal
```
GarrettEHill, line 607: only one of the values needs to be cast to a decimal to get a decimal value after the division, there should be no need to cast the result to a decimal also
```suggestion
      (count(DISTINCT "activityReportId") / ${totalReportCount} * 100)::decimal(4,2)
```

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    END AS "resourcesPct"
+  FROM ${tblNames.createdAroResourcesTempTableName};
+  `;
+  const pctOfReportsWithResources = sequelize.query(pctOfResourcesSql, {
+    type: QueryTypes.SELECT,
+    transaction,
+  });
+
+  // - Number of Reports with ECLKC Resources Pct -
+  const pctOfECKLKCResources = sequelize.query(`
```
GarrettEHill, line 617: please add the /* sql */ at the start of the query string to allow for syntax highlighting to make the code easier to read.

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    WITH eclkc AS (
+      SELECT
+    COUNT(DISTINCT url) AS "eclkcCount"
+    FROM  ${tblNames.createdFlatResourceTempTableName}
+    WHERE url ilike '%eclkc.ohs.acf.hhs.gov%'
```
GarrettEHill, line 622: why not to a = check with the domain?

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    COUNT(DISTINCT url) AS "allCount"
+    FROM ${tblNames.createdFlatResourceTempTableName}
+    )
+    SELECT
+      e."eclkcCount",
+    r."allCount",
+      CASE WHEN
+    r."allCount" = 0
+    THEN 0
+    ELSE round(e."eclkcCount" / r."allCount"::decimal * 100,4)
```
GarrettEHill, line 634:
```suggestion
     ELSE  (e."eclkcCount" / r."allCount"::decimal * 100)::decimal(4,2)
```

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    FROM eclkc e
+    JOIN allres r
+      ON 1=1;
+  `, {
+    type: QueryTypes.SELECT,
+    transaction,
+  });
+
+  // 5.) Date Headers table.
+  const dateHeaders = sequelize.query(`
```
GarrettEHill, line 645: please add the /* sql */ at the start of the query string to allow for syntax highlighting to make the code easier to read.

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    reportIds.push({ id: 0 });
+  }
+  // 2.) Create temp table names.
+  const createdArTempTableName = `Z_temp_resdb_ar__${uuidv4().replaceAll('-', '_')}`;
+  const createdAroResourcesTempTableName = `Z_temp_resdb_aror__${uuidv4().replaceAll('-', '_')}`;
+  const createdResourcesTempTableName = `Z_temp_resdb_res__${uuidv4().replaceAll('-', '_')}`;
+  const createdAroTopicsTempTableName = `Z_temp_resdb_arot__${uuidv4().replaceAll('-', '_')}`;
+  const createdTopicsTempTableName = `Z_temp_resdb_topics__${uuidv4().replaceAll('-', '_')}`;
+  const createdFlatResourceHeadersTempTableName = `Z_temp_resdb_headers__${uuidv4().replaceAll('-', '_')}`; // Date Headers.
+  const createdFlatResourceTempTableName = `Z_temp_resdb_flat__${uuidv4().replaceAll('-', '_')}`; // Main Flat Table.
```
GarrettEHill, line 701: why not use the same uuid for the set of temp tables, they way it will be easier to debug when they all have the matching part of the name?

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+    [
+      resourceUseResult,
+      topicUseResult,
+      numberOfParticipants,
+      numberOfRecipients,
+      pctOfReportsWithResources,
+      pctOfECKLKCResources,
+      dateHeaders,
+    ],
+  );
```
GarrettEHill, line 752: Thank you from doing this in parallel

src/services/dashboards/resource.js
```diff
@@ -1496,7 +1929,7 @@ Expected JSON:
             title: 'Jan-22',
              value: '14',
           },
-          {
+          {79
```
GarrettEHill, line 1932: what is this change?

src/services/dashboards/resource.js
```diff
@@ -1604,3 +2037,114 @@ export async function resourceDashboard(scopes) {
     domainList: generateResourceDomainList(data, true),
   };
 }
+
+export async function rollUpResourceUse(data) {
+  const rolledUpResourceUse = data.resourceUseResult.reduce((accumulator, resource) => {
+    const exists = accumulator.find((r) => r.url === resource.url);
```
GarrettEHill, line 2043: use a map instead of find to increased speed

src/services/dashboards/resource.js
```diff
@@ -1604,3 +2037,114 @@ export async function resourceDashboard(scopes) {
...
+
+  // Sort by total and name or url.
+  // rolledUpResourceUse.sort((r1, r2) => r2.total - r1.total || r1.sortBy.localeCompare(r2.sortBy));
+  rolledUpResourceUse.sort((r1, r2) => r2.total - r1.total || r1.url.localeCompare(r2.url));
+  return rolledUpResourceUse;
+}
+
+export async function rollUpTopicUse(data) {
+  const rolledUpTopicUse = data.topicUseResult.reduce((accumulator, topic) => {
+    const exists = accumulator.find((r) => r.name === topic.name);
```
GarrettEHill, line 2078: use a map instead of find for increased speed

src/services/dashboards/resource.js
```diff
@@ -1604,3 +2037,114 @@ export async function resourceDashboard(scopes) {
...
+  }, []);
+
+  // Loop through the rolled up resources and add a total.
+  rolledUpResourceUse.forEach((resource) => {
+    resource.data.push({ title: 'Total', value: resource.total });
+  });
+
+  // Sort by total and name or url.
+  // rolledUpResourceUse.sort((r1, r2) => r2.total - r1.total || r1.sortBy.localeCompare(r2.sortBy));
+  rolledUpResourceUse.sort((r1, r2) => r2.total - r1.total || r1.url.localeCompare(r2.url));
```
GarrettEHill, line 2072: its not sorting by name

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+  SELECT
+    id,
+    "startDate",
+    "numberOfParticipants",
+    to_char("startDate", 'Mon-YY') AS "rollUpDate",
+    "regionId",
+    "calculatedStatus"
+    INTO TEMP ${tblNames.createdArTempTableName}
+    FROM "ActivityReports" ar
+    WHERE ar."id" IN (${reportIds.map((r) => r.id).join(',')});
```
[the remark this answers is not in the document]
AdamAdHocTeam: The AR filtering is handled using scopes, correct me if I'm wrong but that should be done by default?

src/services/dashboards/resource.js
AdamAdHocTeam: Odd looks like a typo surprised lint didn't catch it.

src/services/dashboards/resource.js
AdamAdHocTeam: For now no, the commented out code is the one we will ultimately use however to match closer to your order ( in the case of top 10) we will just use url for now.

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+      FROM "Resources"
+      WHERE id IN (
+      SELECT DISTINCT "resourceId"
+        FROM ${tblNames.createdAroResourcesTempTableName}
+      );
+
+      -- 4.) Create ARO Topics temp table.
+      SELECT
+      ar.id AS "activityReportId",
+      arot."activityReportObjectiveId", -- We need to group by this incase of multiple aro's.
```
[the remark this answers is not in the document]
AdamAdHocTeam: This is one row per topic.

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+        name
+        INTO TEMP ${tblNames.createdTopicsTempTableName}
+        FROM "Topics"
+        WHERE id IN (
+        SELECT DISTINCT "topicId"
+          FROM ${tblNames.createdAroTopicsTempTableName}
+        );
+
+      -- 6.) Create Flat Resource temp table.
+      SELECT
```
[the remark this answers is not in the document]
AdamAdHocTeam: This table should be 1 record per resource. Are you saying that's not the case?

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+      FROM urlvals
+      GROUP BY url, title
+      ORDER BY SUM("resourceCount")  DESC,
+      url ASC
+      LIMIT 10
+    ),
+    series AS
+    (
+      SELECT * FROM ${tblNames.createdFlatResourceHeadersTempTableName}
+    )
```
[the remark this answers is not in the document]
AdamAdHocTeam: This gives us every month in the range for each row. We either do it here or programmatically on web server mem.

Let me know if you want to talk about other options.

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+        name,
+        SUM("resourceCount") AS "totalCount"
+      FROM topicsperdate
+      GROUP BY name
+      ORDER BY SUM("resourceCount") DESC
+    ),
+    series AS
+    (
+      SELECT * FROM ${tblNames.createdFlatResourceHeadersTempTableName}
+    )
```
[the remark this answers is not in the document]
AdamAdHocTeam: We generate the series once. Then we join our data to it and ensure every resouce or topic has a date.

src/services/dashboards/resource.js
AdamAdHocTeam: So are you saying we don't need to filter active grants?

src/services/dashboards/resource.js
AdamAdHocTeam: Its the resulting count to decimal

src/services/dashboards/resource.js
AdamAdHocTeam: Again its the count.

src/services/dashboards/resource.js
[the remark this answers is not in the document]
GarrettEHill: please verify

src/services/dashboards/resource.js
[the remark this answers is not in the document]
AdamAdHocTeam: Cross join can't be the last join here as we join on dates. So if I have a match for Jan-22 but I don't have a match for Feb-22 I still want it in the output.

src/services/dashboards/resource.js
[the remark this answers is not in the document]
GarrettEHill: for a report with 5 recipients, and an objective with 1 topic, why do you need 5 rows with the same report id and topic. when a count of objectives might be more correct and faster processing later.

src/services/dashboards/resource.js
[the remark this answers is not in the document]
GarrettEHill: it will be multiple as currently written

src/services/dashboards/resource.js
```diff
@@ -330,6 +331,438 @@ const switchToTopicCentric = (input) => {
...
+      SELECT
+          generate_series(
+            date_trunc('month', (SELECT MIN("startDate") FROM ${tblNames.createdFlatResourceTempTableName})),
+            date_trunc('month', (SELECT MAX("startDate") FROM ${tblNames.createdFlatResourceTempTableName})),
+            interval '1 month'
+          )::date AS "date"
+        INTO TEMP ${tblNames.createdFlatResourceHeadersTempTableName};
+  `;
+
+  const transaction = await sequelize.transaction();
```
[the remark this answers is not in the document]
GarrettEHill: the issue you are seeing is that test normaly dont use a transaction. which is why you think you need transactions implemented like this. in production or a running system, all the calls will already be in a transaction as all calls through the route are wrapped in a single transaction. Please put your transaction wrapper in the test file not in the code directly as it is not correct.

src/services/dashboards/resource.js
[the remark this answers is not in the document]
GarrettEHill: use ${tblNames.createdFlatResourceHeadersTempTableName} direclty in the query instead of making an alias of it named series

src/services/dashboards/resource.js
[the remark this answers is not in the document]
GarrettEHill: use ${tblNames.createdFlatResourceHeadersTempTableName} directly in the query instead of making an alias of it named series

src/services/dashboards/resource.js
GarrettEHill: > response is functionally the correct one with regard to your question as stated; the SQL Server is smart enough to realize that if you are using "Group By" and not using any aggregate functions, then what you actually mean is "Distinct" - and therefore it generates an execution plan as if you'd simply used "Distinct."

please just use distinct as that's what is really happening, as there is no agg operations

src/services/dashboards/resource.js
AdamAdHocTeam: Ok I made the change you suggested but had to join recipients.

GarrettEHill: that too, and its way too much joins

src/services/dashboards/resource.js
GarrettEHill: why do you need it as a decimal?

src/services/dashboards/resource.js
GarrettEHill: why do you need it as a decimal?

src/services/dashboards/resource.js
AdamAdHocTeam: just the way im used to checking

src/services/dashboards/resource.js
AdamAdHocTeam: ah its in a comment

src/services/dashboards/resource.js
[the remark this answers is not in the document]
AdamAdHocTeam: image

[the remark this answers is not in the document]
AdamAdHocTeam: I guess what do you mean by complete, this is approved only?

src/services/dashboards/resource.js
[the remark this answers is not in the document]
AdamAdHocTeam: Nice catch its supposed to be objective id not aro id. This should speed things up a good amount.

Thanks! :-)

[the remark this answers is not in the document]
AdamAdHocTeam: Actually there is a comment on that group by I think it was done intentionally.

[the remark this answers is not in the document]
AdamAdHocTeam: I changed this for now we always put it back.

src/lib/cache.test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-commented-out-tests */
```
thewatermethod, line 1: we want this uncommented now, right?

src/services/dashboards/resourceFlat.test.js
```diff
@@ -0,0 +1,857 @@
...
+          totalCount: '2',
+        },
+      ]);
+    });
+  });
+
+  it('resourceTopicUseFlat', async () => {
+    const scopes = await filtersToScopes({ 'region.in': [REGION_ID], 'startDate.win': '2021/01/01-2021/01/31' });
+    let topicUseResult;
+    db.sequelize.transaction(async () => {
```
thewatermethod, line 543: why are these in transactions?

thewatermethod:
https://github.com/HHS/Head-Start-TTADP/assets/3288586/7659ad10-62b9-4b25-8196-51f65d3ba291
- This table doesn't seem to switch between asc/desc sort
- Everything looks good. Faster. Data looks to be correct, although I did not check with every filter.
- Just a couple of test related questions on the code

src/services/dashboards/resourceFlat.test.js
AdamAdHocTeam: This is because the app will always be wrapped in a txn we need to add it for the test.

src/lib/cache.test.js
AdamAdHocTeam: putting it back.

AdamAdHocTeam:
> Screen.Recording.2024-04-10.at.1.57.57.PM.mov 
> * This table doesn't seem to switch between asc/desc sort
> * Everything looks good. Faster. Data looks to be correct, although I did not check with every filter.
> * Just a couple of test related questions on the code

This is very odd it works for me locally

https://github.com/HHS/Head-Start-TTADP/assets/84350609/0774784c-41d3-4b4b-9456-3aea7a65b7e0

thewatermethod:
> This is very odd it works for me locally, it also works on DEV.

Weird, it doesn't seem to be working on dev for me

thewatermethod:
Maybe I am counting something wrong

**Counted in the ARs**
Facilities	1
Ongoing Monitoring and Continuous Improvement	1
Recordkeeping and Reporting	1
Leadership / Governance	3
Program Planning and Services	3

Here is the chart, showing 5 instead of 3 (it's a 2recipient report)

Screenshot 2024-04-11 at 1 25 51 PM

```sql
SELECT
  t.name,
  COUNT(arot."topicId") count
FROM "ActivityReportObjectives" aro
INNER JOIN "ActivityReportObjectiveTopics" arot ON aro."id" = arot."activityReportObjectiveId"
JOIN "Topics" t on arot."topicId" = t."id"
WHERE aro."activityReportId" in (40748,39987,40030)
GROUP BY t."name"
;
```

yields this:
Facilities	1
Leadership / Governance	3
Ongoing Monitoring and Continuous Improvement	1
Program Planning and Services	3
Recordkeeping and Reporting	1


Filters to repro: /dashboards/resources-dashboard?startDate.win=2024%2F02%2F09-2024%2F02%2F09&region.in[]=3

thewatermethod (approved):

GarrettEHill (approved):
My only concern is that the method of aggregation may not be reporting the numbers expected by the client. But what is implemented looks good enough for now.

AdamAdHocTeam:
> My only concern is that the method of aggregation may not be reporting the numbers expected by the client. But what is implemented looks good enough for now.

Can you give an example? Maybe it makes sense to run this by OHS first.

PR #2080, opened 2024-04-05
[TTAHUB-2790] use closed_goal_merge_override feature flag to bypass filter

GarrettEHill:
## Description of change
closed_goal_merge_override feature flag was not bypassing the filtering of goals to merge as it was intended

## How to test
with the feature flag on, closed and open matching curated goals can be merged. 

## Issue(s)

* [TTAHUB-2790](https://jira.acf.gov/browse/TTAHUB-2790)

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/goalSimilarityGroup.ts
+++ b/src/services/goalSimilarityGroup.ts
@@ -46,10 +46,15 @@ interface SimilarityGroup {
   toJSON: () => SimilarityGroup;
 }
 
-export const flattenSimilarityGroupGoals = (group: SimilarityGroup) => ({
+export const flattenSimilarityGroupGoals = (
+  group: SimilarityGroup,
+  userHasFeatureFlag = false,
+) => ({
   ...group.toJSON(),
   goals: group.goals.filter((goal) => {
-    if (goal.goalTemplate && goal.goalTemplate.creationMethod === CREATION_METHOD.CURATED) {
+    if (goal.goalTemplate
+      && goal.goalTemplate.creationMethod === CREATION_METHOD.CURATED
+      && !userHasFeatureFlag) {
       return goal.status !== GOAL_STATUS.CLOSED;
     }
     return true;
@@ -120,7 +125,7 @@ export async function getSimilarityGroupByContainingGoalIds(
             SELECT gs."goalSimilarityGroupId"
             FROM "GoalSimilarityGroupGoals" gs
             WHERE gs."goalId" IN (${goalIds.join(',')})
-            GROUP BY gs."goalSimilarityGroupId" 
+            GROUP BY gs."goalSimilarityGroupId"
             HAVING COUNT(DISTINCT gs."goalId") = ${goalIds.length}
           )`,
         ),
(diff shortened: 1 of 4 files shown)


thewatermethod:
Looks to me like this change means that if the user who first looks at the recipient record has the feature flag, all subsequent users will see the groups as that user did (including closed curated goals, etc)

GarrettEHill:
> ~Looks to me like this change means that if the user who first looks at the recipient record has the feature flag, all subsequent users will see the groups as that user did (including closed curated goals, etc)~ nvm I misread what was happening here, sorry
> 
> There is code to do this already though, when the goal groups are created. It might be better to take a look at what the logic error is there.

It was working before, then some of the logic looks to have changed 2 months ago and now its not working. This make it work again.

src/services/goalSimilarityGroup.ts
```diff
@@ -46,10 +46,15 @@ interface SimilarityGroup {
   toJSON: () => SimilarityGroup;
 }
 
-export const flattenSimilarityGroupGoals = (group: SimilarityGroup) => ({
+export const flattenSimilarityGroupGoals = (
+  group: SimilarityGroup,
+  userHasFeatureFlag = false,
```
thewatermethod, line 51:
```suggestion
```

src/services/goalSimilarityGroup.ts
```diff
@@ -46,10 +46,15 @@ interface SimilarityGroup {
...
+) => ({
   ...group.toJSON(),
   goals: group.goals.filter((goal) => {
-    if (goal.goalTemplate && goal.goalTemplate.creationMethod === CREATION_METHOD.CURATED) {
+    if (goal.goalTemplate
+      && goal.goalTemplate.creationMethod === CREATION_METHOD.CURATED
+      && !userHasFeatureFlag) {
       return goal.status !== GOAL_STATUS.CLOSED;
     }
     return true;
```
thewatermethod, line 60: This entire filter should be removed, and replaced with

```js
goals: group.goals.map((goal) => goal.id)

File: src/services/goalSimilarityGroup.ts [hunk 3]
```diff
@@ -193,7 +198,7 @@ export async function getSimilarityGroupsByRecipientId(
   });
 
   return groups.map(
-    (gg: SimilarityGroup) => flattenSimilarityGroupGoals(gg),
+    (gg: SimilarityGroup) => flattenSimilarityGroupGoals(gg, userHasFeatureFlag),
```
thewatermethod, line 201:
```suggestion
    (gg: SimilarityGroup) => flattenSimilarityGroupGoals(gg),
```

thewatermethod:
> It was working before, then some of the logic looks to have changed 2 months ago and now its not working. This make it work again.

@GarrettEHill 
The logic was not removed but moved. There is a column in the GoalSimilarityGroupGoals table (excluded if not admin.) Goals are flagged as excluded when they groups are created. The bug here is that the filter should have been removed from the GET service.

thewatermethod (changes requested):
https://github.com/HHS/Head-Start-TTADP/pull/2086

PR #2076, opened 2024-04-03
[TTAHUB-2575] Escape fields for XSS

thewatermethod:
## Description of change
Escape the fields called out via pen test

This PR is large because I had to relocate the hooks folder to mitigate OOM issues with Jest on the CI. The vast majority of files were only slightly changed to update imports. Also, the relocation caused test coverage to fall below acceptable levels, so I had to add ~a few~ many tests as well.

## How to test

- Follow the instructions in the pen test tickets
- I self-reviewed the files that were changed as part of the Jira. The rest of the files, nearly 140, are just renames or added tests.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2573
* https://jira.acf.gov/browse/TTAHUB-2575
* https://jira.acf.gov/browse/TTAHUB-2576
* https://jira.acf.gov/browse/TTAHUB-2577

### Every PR

- [ ] Meets issue criteria
- [ ] JIRA ticket status updated
- [ ] Code is meaningfully tested
- [ ] Meets accessibility standards (WCAG 2.1 Levels A, AA)
- [ ] API Documentation updated
- [ ] Boundary diagram updated
- [ ] Logical Data Model updated
- [ ] [Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete
- [ ] Ready to create production PR

### Production Deploy

- [ ] Staging smoke test completed

### After merge/deploy

- [ ] Update JIRA ticket status

PR #2092, opened 2024-04-10
Revert "[TTAHUB-2575] Escape fields for XSS"

thewatermethod:
Reverts HHS/Head-Start-TTADP#2076


AdamAdHocTeam (approved):
Would wait for the OK from the client first.

PR #2101, opened 2024-04-15
[TTAHUB-2683] Use status change service fn in place of `Goal.update({ status })`

nvms:
## Description of change

In this PR we begin using the `changeGoalStatus` helper in place of e.g. `Goal.update({ status })`. It's only a few places, it seems.

## How to test

- Make sure the tests pass and the code makes sense.
- Ensure there are no other places I missed where we change goal status.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2683

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


nvms:
I don't think we want to use the helper here:

https://github.com/HHS/Head-Start-TTADP/blob/jp/2683/useNewHelper/src/tools/importPlanGoals.js/#L47-L55

Right?

src/goalServices/changeGoalStatus.ts
```diff
@@ -6,15 +7,25 @@ interface GoalStatusChangeParams {
...
   context,
+  transaction = null,
 }: GoalStatusChangeParams) {
+  // If userId is null, and we're in CI or a Jest test, just set it to 5.
+  // This let's us simplify the way we write the many tests that end up using this function
+  // one way or another, since mocking httpContext is a pain.
+  if (!userId && (process.env.NODE_ENV === 'test' || process.env.CI)) {
+    // eslint-disable-next-line no-param-reassign
+    userId = 5;
+  }
```
nvms, line 27: I know not everybody will like this, but the alternative (at least the way I see it) is to mock httpContext (because that's what is used to get the userId) in every test that changes the status of a Goal, which would be a lot of additional code, compared to just three lines here. I think it's a reasonable compromise, but open to better solutions if there are any.

src/models/hooks/objective.js
```diff
@@ -201,12 +202,16 @@ const propogateStatusToParentGoal = async (sequelize, instance, options) => {
 
       // and if so, we update it (storing the previous status so we can revert if needed)
       if (atLeastOneInProgress) {
-        await goal.update({
-          status: 'In Progress',
-          previousStatus: 'Not Started',
-        }, {
+        // eslint-disable-next-line global-require
+        const changeGoalStatus = require('../../goalServices/changeGoalStatus').default;
+        const userId = httpContext.get('impersonationUserId') || httpContext.get('loggedUser');
```
GarrettEHill, line 207: question, could the parameter order of changeGoalStatus be changed, so that userId is near the end, then have the httpContext calls used as the default value if no user is supplied. That way it would not need to be included all over the pace and need to be called/looked up all over, when its only needed in the one place for the changeGoalStatus function?

GarrettEHill:
> I don't think we want to use the helper here:
> 
> https://github.com/HHS/Head-Start-TTADP/blob/jp/2683/useNewHelper/src/tools/importPlanGoals.js/#L47-L55
> 
> Right?

No, all status changes to goals should use the new code path.

src/models/hooks/objective.js
nvms: `changeGoalStatus` only has one parameter: an object

src/models/hooks/activityReportObjective.test.js
```diff
@@ -131,4 +131,41 @@ describe('activityReportObjective hooks', () => {
...
+        supportType,
+      });
+
+      const updatedObjective = await Objective.findOne({
+        where: { id: supportObjective.id },
+      });
+
+      expect(updatedObjective.supportType).toEqual(supportType);
+
+      await ActivityReportObjective.destroy({
```
thewatermethod, line 166: could you move the destroy to the afterAll? we probably want it to run even if the test fails for some weird reason

thewatermethod:
There's a "previousStatus" field we still use a couple of places I think. Looks like that isn't updated via the handler? I don't think we've updated any queries to return that anywhere either.

Other than what I pointed out above, everything looks & works good

src/goalServices/changeGoalStatus.test.js
```diff
@@ -54,12 +54,13 @@ describe('changeGoalStatus service', () => {
   });
 
   afterAll(async () => {
-    await db.User.destroy({ where: { id: mockUser.id } });
-    await db.Goal.destroy({ where: { id: goal.id } });
+    await db.Goal.destroy({ where: { id: goal.id }, force: true });
+    await db.GrantNumberLink.destroy({ where: { grantId: grant.id }, force: true });
```
nvms, line 58: `force: true` because we don't just want to set `deletedAt`, we want to actually remove these rows.

`GrantNumberLink` removed before `Grant` because the grant fails to remove due to a constraint.

src/models/hooks/activityReportObjective.test.js
nvms: 7e82c67faf544dfe68e06ee604c823516e2e8135

nvms:
> There's a "previousStatus" field we still use a couple of places I think. Looks like that isn't updated via the handler? I don't think we've updated any queries to return that anywhere either.
> 
> Other than what I pointed out above, everything looks & works good

58402bb9a979352a50f15893e1a0e709b2e9ad33

nvms:
> There's a "previousStatus" field we still use a couple of places I think. Looks like that isn't updated via the handler? I don't think we've updated any queries to return that anywhere either.
> 
> Other than what I pointed out above, everything looks & works good

Was planning on removing all references to that, as well as the other columns, as part of https://jira.acf.gov/browse/TTAHUB-2794 (next PR)

thewatermethod (approved):
Looks good, obv we want CI to pass before merging

src/models/hooks/activityReportObjective.test.js
```diff
@@ -131,4 +131,44 @@ describe('activityReportObjective hooks', () => {
...
+    afterAll(async () => {
+      await Objective.destroy({
+        where: { id: supportObjective.id },
+        force: true,
+      });
+
+      if (aroWithSupportType) {
+        await ActivityReportObjective.destroy({
+          where: { id: aroWithSupportType.id },
+        });
```
thewatermethod, line 155: I know I approved this but does this work? deleting the objective while there is still an ARO attached to it? I think you'd have to do the aro destroy first... you can also lose the if and do ```where: { objectiveId: supportObjective.id }```, I assume

nvms: Oh yeah, good catch: 657a9c117349f7f69b7122eb12b2c572c5e0a6c6

PR #2111, opened 2024-04-23
[TTAHUB-2784] correct onApprovedAR values for goals and obj

hardwarehuman:
## Description of change

This is a one-time migration to determine which Goals and Objectives have incorrect or null 'cached' values for `onApprovedAR`, then set them to the correct values

## How to test

There are before and after values, plus a post-operation select statement that brings together the numbers, which should balance out like so:
```
 order |       description        | matching_values | incorrect_values | should_be__marked_approved_but_isnt | marked_approved_but_shouldnt_be | total
-------+--------------------------+-----------------+------------------+-------------------------------------+---------------------------------+--------
     1 | objective starting stats |          111976 |             5257 |                                   1 |                            5256 | 117233
     2 | objective values changed |                 |                  |                                   1 |                            5256 |   5257
     3 | objective ending stats   |          117233 |                0 |                                   0 |                               0 | 117233
     4 | goal starting stats      |           65745 |             3364 |                                 880 |                            2484 |  69109
     5 | goal values changed      |                 |                  |                                 880 |                            2484 |   3364
     6 | goal ending stats        |           69109 |                0 |                                   0 |                               0 |  69109
```

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2784

### Every PR

- [x] Meets issue criteria

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/migrations/20240420000000-correct-onApprovedAR-values.js
```diff
@@ -0,0 +1,252 @@
...
+        SUM(CASE WHEN on_approved_ar = "onApprovedAR" THEN 1 ELSE 0 END
+        ) matching_values_o,
+        SUM(
+          CASE
+            WHEN on_approved_ar = "onApprovedAR"
+              OR "onApprovedAR" IS NULL
+          THEN 0
+          ELSE 1
+          END
+        ) incorrect_values_o,
```
GarrettEHill, line 57: while this does work it takes a bit more time to read and understand what is being done here. `SUM(CASE)` is also a bit slower then the alternative `count() filter (where)`.

src/migrations/20240420000000-correct-onApprovedAR-values.js
GarrettEHill, line 1: While this does correct the onApprovedAR, The onAR values are also not correct for a highly overlapping subset. Do you think the onAR could also be corrected in the same pass?

src/migrations/20240420000000-correct-onApprovedAR-values.js
hardwarehuman: It was complaining for some reason and I didn't feel like figuring out why. Though now that I think about it I think I know

src/migrations/20240420000000-correct-onApprovedAR-values.js
hardwarehuman: Not as quick as I thought, but done now

src/migrations/20240420000000-correct-onApprovedAR-values.js
```diff
@@ -0,0 +1,331 @@
...
+      --  2. Calculate correct onApprovedAR values for goals
+      DROP TABLE IF EXISTS goals_on_ars;
+      CREATE TEMP TABLE goals_on_ars
+      AS
+      SELECT
+        g.id gid,
+        BOOL_OR(
+          (ar.id IS NOT NULL AND ar."calculatedStatus" = 'approved')
+          OR
+          COALESCE(ooaa.on_approved_ar,FALSE)
```
AdamAdHocTeam, line 36: I haven't gotten to the usage of this table yet. Is there any concern that a goal without an ar would have on_approved_ar set to true with the ooaa table link?

src/migrations/20240420000000-correct-onApprovedAR-values.js
```diff
@@ -0,0 +1,331 @@
...
+        SET "onApprovedAR" = on_approved_ar
+        FROM goals_on_ars
+        WHERE g.id = gid
+          AND ("onApprovedAR" != on_approved_ar OR "onApprovedAR" IS NULL)
+        RETURNING
+          gid,
+          on_approved_ar
+      ) SELECT * FROM updater
+      ;
+      --  8. Update onAR values for goals and save the results
```
AdamAdHocTeam, line 168: Should this be number 10

src/migrations/20240420000000-correct-onApprovedAR-values.js
```diff
@@ -0,0 +1,331 @@
...
+      LEFT JOIN "ActivityReports" ar
+        ON arg."activityReportId" = ar.id
+        AND ar."calculatedStatus" != 'deleted'
+      LEFT JOIN "Objectives" o
+        ON o."goalId" = g.id
+      LEFT JOIN objectives_on_ars ooaa
+        ON ooaa.oid = o.id
+      GROUP BY 1
+      ;
+      --  3. Calculate onApprovedAR stats for objectives
```
AdamAdHocTeam, line 51: Seems you have alot of stats sql, I think this is a good idea. Just be careful that a sql error in your stats table/calc doesn't give you a false positive. Some times I find writing a new smaller query in the analyzer can help me re-think what I already have written.

src/migrations/20240420000000-correct-onApprovedAR-values.js
hardwarehuman: Oops, yep, this is out of order

src/migrations/20240420000000-correct-onApprovedAR-values.js
hardwarehuman: If the Objective is on an AR and is linked to the Goal, then it is inherently on an AR. Basically, there's two ways to link a goal to an AR: directly through activityReportGoals->Goals, or indirectly through activityReportObjectives->Objectives->Goals

src/migrations/20240420000000-correct-onApprovedAR-values.js
hardwarehuman: I'm not sure if I understand what you're saying, but I will note that:
- All the temp tables can be examined separately
- The calculations for the before/afters are performed in a different way from the calculations for the updates to lower the chance that they correlate spuriously

PR #2123, opened 2024-04-30
[NO TICKET] Add a trigger for production PR notifications

kryswisnaskas:
## Description of change
The GitHub Actions for sending production PR notifications did not trigger when PRs were created in an open state, rather than moving from draft to open.

## How to test
No tests needed

## Issue(s)

### Every PR

- [x] Meets issue criteria
- [n/a] UI review complete

### Before merge to main

- [n/a] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/.github/workflows/pr-notifications.yml
+++ b/.github/workflows/pr-notifications.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches:
       - production
-    types: [ready_for_review]
+    types: [opened, ready_for_review]
 
 jobs:
   notify_slack:


thewatermethod:
First: PR is still in draft, but I assume is ready for review?

This is probably good. We have to be aware if we make this change, in the example of the PR today, that it will notify BEFORE it is actually ready for review right? 

What I mean is, the PR that caused this change was created yesterday but wasn't ready for review until today. Therefore it would have posted yesterday and we would have had to explain that, no, please no approve yet.

I don't want to tie this up for too long, but are we fixing something that isn't broken, and the real fix is to create production pull requests as draft PRs?

kryswisnaskas:
> First: PR is still in draft, but I assume is ready for review?
> 
> This is probably good. We have to be aware if we make this change, in the example of the PR today, that it will notify BEFORE it is actually ready for review right?
> 
> What I mean is, the PR that caused this change was created yesterday but wasn't ready for review until today. Therefore it would have posted yesterday and we would have had to explain that, no, please no approve yet.
> 
> I don't want to tie this up for too long, but are we fixing something that isn't broken, and the real fix is to create production pull requests as draft PRs?

Thank you for pointing that out. I completely agree with you. Actually, I intended for the notifications to sync up exactly with when a PR is ready for review.

We could use this PR to straighten out our process: Let’s start opening production PRs as drafts and only switch them to open when they’ve passed smoke tests and are actually ready. That way, the Slack notifications will only go off when everything’s truly ready for a final look.

thewatermethod:
> We could use this PR to straighten out our process: Let’s start opening production PRs as drafts and only switch them to open when they’ve passed smoke tests and are actually ready. That way, the Slack notifications will only go off when everything’s truly ready for a final look.

I like it. So we don't need a code change then, as those notifications work as expected?

kryswisnaskas:
Closing this PR because we do not want to prematurely post a slack notification. Transitioning from a draft state to open will automatically trigger the posting.

PR #2130, opened 2024-05-02
[TTAHUB-2880] merge non-fei goal to fei goal

hardwarehuman:
## Description of change

The user had an old non-FEI goal they wanted merged into their textually dissimilar FEI goal.

## How to test
There's a result set at the end that should look like:
```
 op_order |            operation             | cnt
----------+----------------------------------+-----
        1 | deleted_gfrs                     |   0
        2 | updated_target_objectives        |   0
        3 | relinked_objective_topics        |   0
        4 | relinked_objective_resources     |   0
        5 | relinked_objective_files         |   0
        6 | deleted_objective_topics         |   0
        7 | deleted_objective_resources      |   0
        8 | deleted_objective_files          |   0
        9 | relinked_objectives              |   1
       10 | updated_target_goals             |   1
       11 | relinked_args                    |   2
       12 | deleted_args                     |   0
       13 | deleted_goals                    |   1
       14 | relinked_goal_collaborators      |   0
       15 | deleted_goal_collaborators       |   3
       16 | relinked_objective_collaborators |   0
       17 | deleted_objective_collaborators  |   0
```

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2880

### Every PR

- [x] Meets issue criteria

### Before merge to main

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20240503000000-merge_to_fei_goal.js
+++ b/src/migrations/20240503000000-merge_to_fei_goal.js
@@ -1,0 +1,470 @@
+const {
+  prepMigration,
+} = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+
+      await queryInterface.sequelize.query(`
+      -- A user wants non-FEI goals merged into their FEI goal
+
+      -- PROCESS:
+      -- Delete existing root cause (there should be none for this PR)
+      -- Test to make sure no root causes were actually deleted (this PR only)
+      -- Merge goal collaborators
+      -- Merge objective collaborators
+      -- Update the merge target objectives
+      -- Merge objective topics
+      -- Merge objective resources
+      -- Merge objective files
+      -- Delete duplicate objective topics
+      -- Delete duplicate objective resources
+      -- Delete duplicate objective files
+      -- Relink non-goalmerge Objectives
+      -- Update the target goals record history fields from the donor goals
+      -- Relink non-duplicate ARGs
+      -- Delete duplicate ARGs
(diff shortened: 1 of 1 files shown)


thewatermethod:
I confirmed that this worked via the UI, and I think this is OK. 

One thing to note about these merges is we lose some of the merge data since we aren't following the same pattern as the UI. A couple of examples: we don't create a new goal and assign the old ones a "mapsToParentGoalId". Since the new goal isn't created, it doesn't have a "createdVia" set to "merge"and thus the goal doesn't have the little "merge" badge in the goal cards. Also, since that pattern isn't followed, we might eventually exclude these goals from analytics, or some other weird thing might be missed that I'm not thinking of at the moment.

I guess my point is, if we keep doing these "pseudo-merges" perhaps merge isn't the right word, since we should be managing expectations on what everyone is getting (not exactly a merge).

thewatermethod (approved):

hardwarehuman:
> I confirmed that this worked via the UI, and I think this is OK.
> 
> One thing to note about these merges is we lose some of the merge data since we aren't following the same pattern as the UI. A couple of examples: we don't create a new goal and assign the old ones a "mapsToParentGoalId". Since the new goal isn't created, it doesn't have a "createdVia" set to "merge"and thus the goal doesn't have the little "merge" badge in the goal cards. Also, since that pattern isn't followed, we might eventually exclude these goals from analytics, or some other weird thing might be missed that I'm not thinking of at the moment.
> 
> I guess my point is, if we keep doing these "pseudo-merges" perhaps merge isn't the right word, since we should be managing expectations on what everyone is getting (not exactly a merge).

I think this is a pretty good point, though I think our real aspiration is *not* to do these any more. I'm also unsure if we do or don't want these included in the list of goals that had been merged because they are definitely not going through the normal process in any respect. Regardless, best if we obviate the whole think by switching to the programmatic goal merges via the admin page once we can get it scheduled.

PR #2145, opened 2024-05-10
[TTAHUB-2895] Hide edit button if the user has incorrect permissions

AdamAdHocTeam:
## Description of change

1. If a user doesn't have permission to create reports in a region they shouldn't be able to edit a RTR goal in that region. That being said they should be able to view the goal with read only permissions.
2. If a user is read only they shouldn't be able to change the goal source OR support type when viewing the goal.

_Note: Confirmed with Angela even if the user is an admin but doesn't have the permission hide the option._

## How to test

1. Visit the RTR without the write permission for that region. You should not be able to edit the goal (but should be able to view). Add the permission for that region you now should be able to edit.
2. Viewing a goal with read only permissions shouldn't allow the user to change the goal source OR support type. 

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2895

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/components/GoalCards/GoalCard.js
+++ b/frontend/src/components/GoalCards/GoalCard.js
@@ -17,7 +17,7 @@ import { goalPropTypes } from './constants';
 import colors from '../../colors';
 import SessionObjectiveCard from './SessionObjectiveCard';
 import Tooltip from '../Tooltip';
-import isAdmin, { hasApproveActivityReportInRegion } from '../../permissions';
+import isAdmin, { hasApproveActivityReportInRegion, canEditOrCreateGoals } from '../../permissions';
 import UserContext from '../../UserContext';
 import { deleteGoal } from '../../fetchers/goals';
 import AppLoadingContext from '../../AppLoadingContext';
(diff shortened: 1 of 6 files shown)


thewatermethod:
> Users with READ_REPORTS permission
> should not see the "Edit" menu on goal cards. They should only see "View" menu
> user should not see edit UI for goal source or TTA Support type fields. Currently they have a drop down list rather then read only text

I don't think this meets either of the acceptance criteria.
1) Read only users see no options. They should see view.
2) If the menu is empty, we have a button that does nothing and we shouldn't show it (although I don't think there's actually a case where the button is empty.
3) As you said on slack, reopen goals should be treated the same as edit
4) No changes were made here to the form fields mentioned in the ticket (user should not see edit UI for goal source or TTA Support type fields. Currently they have a drop down list rather then read only text)

AdamAdHocTeam:
> > Users with READ_REPORTS permission
> > should not see the "Edit" menu on goal cards. They should only see "View" menu
> > user should not see edit UI for goal source or TTA Support type fields. Currently they have a drop down list rather then read only text
> 
> I don't think this meets either of the acceptance criteria.
> 
> 1. Read only users see no options. They should see view.
> 2. If the menu is empty, we have a button that does nothing and we shouldn't show it (although I don't think there's actually a case where the button menu is empty.)
> 3. As you said on slack, reopen goals should be treated the same as edit
> 4. No changes were made here to the form fields mentioned in the ticket (user should not see edit UI for goal source or TTA Support type fields. Currently they have a drop down list rather then read only text)

@thewatermethod I think these have all been addressed.

thewatermethod:
@AdamAdHocTeam TTA Support type fields is still editable

thewatermethod (approved):

PR #2147, opened 2024-05-14
[TTAHUB-2530] Remove fields from objective form in RTR

thewatermethod:
## Description of change
- Removed topics, resources, support type, objective status, courses, and files from RTR
- The "view goal" from the RTR is a little different, so I created a new template to stop the logic loops inside the create form from getting even crazier. To do that and keep things consistent, some bits of display were moved into reusable components
- Cleaned up the route on the backend for this page. We no longer care about "ObjectiveTopics", "ObectiveResources", etc, only their "ActivityReportObjectivesTopics" and the like counterparts. Moved it to Typescript to make refactors a little easier in the future.
- Removed an unused route on the backend
- Note that other PRs within this sprint will build on this work, altering the AR fields and cleaning up the database schema to remove unused tables.

## How to test
- Create a new goal on the RTR, with an objective. Confirm that the  topics, resources, support type, objective status, courses, and files fields aren't there.
- Edit that goal. If the objective isn't used on an activity report, the text should still be editable.
- Use that goal and objective on a report. Once the objective is on a report, the text is no longer editable.
- Approve the report and close the goal. Go the RTR and confirm that the view option is present. View the goal and confirm that it matches the design.
- Test the same cases with FEI goals and multirecipient reports.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2530

### Every PR

- [ ] Meets issue criteria
- [ ] JIRA ticket status updated
- [ ] Code is meaningfully tested
- [ ] Meets accessibility standards (WCAG 2.1 Levels A, AA)
- [ ] API Documentation updated
- [ ] Boundary diagram updated
- [ ] Logical Data Model updated
- [ ] [Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete
- [ ] Ready to create production PR

### Production Deploy

- [ ] Staging smoke test completed

### After merge/deploy

- [ ] Update JIRA ticket status

PR #2152, opened 2024-05-20
[TTAHUB-2260] Activity report updates for reduced objective fields 

thewatermethod:
# Stop, potential reviewer! It's recommended by the author that you review #2147 first

## Description of change
- Objective text (for existing objectives) is read-only
- Support type is always editable
- Goal source is editable until goal is on approved AR
- Remove all the code that was displaying previously used Objective topics, etc, as it is dead code

## How to test
- Create an objective on the RTR. Add it to an AR, and confirm that the text is read-only on the AR. Return to the RTR and confirm that it is still read-only, as it has been added to an AR.
- Create an objective on an AR. The text of that objective should be editable on the AR if it was created there
- On the AR, support type is always editable.
- Confirm that goal source is read-only if the goal is on an approved AR. Otherwise, it is editable from a draft AR and from the RTR, and changes in one place are reflected in the other.
- Confirm that for an other entity objective, the objective is editable if it was created on that report, and that support type is always editable.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2260

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


frontend/src/components/Navigator/ActivityReportNavigator.js
```diff
@@ -405,7 +405,7 @@ const ActivityReportNavigator = ({
       };
 
       if (allowUpdateFormData) {
-        updateFormData(data, true);
+        updateFormData(data, false);
```
thewatermethod, line 408: This fixes a bug where if you "save draft" with a goal filled out (but before you've saved and continued) and then leave the report, the goal will disappear. I believe this is the bug that C'era showed us a video of.

frontend/src/components/ObjectiveCourseSelect.js
```diff
@@ -36,10 +36,6 @@ export default function ObjectiveCourseSelect({
     fetchCourses();
   }, []);
 
-  if (!options || options.length === 0) {
```
thewatermethod, line 39: We no longer need this check

docs/logical_data_model.puml
```diff
@@ -137,6 +137,7 @@ class ActivityReportObjectives{
  arOrder : integer : 1
  closeSuspendContext : text
  closeSuspendReason : enum
+ objectiveCreatedHere : boolean
```
GarrettEHill, line 140: Could we have a source column on the Objective instead?

src/migrations/20240516140827-add-created-here-column-to-activity-report-objectives.js
```diff
@@ -0,0 +1,23 @@
+const { prepMigration } = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      await queryInterface.addColumn('ActivityReportObjectives', 'objectiveCreatedHere', {
```
GarrettEHill, line 9: I think this servers the same concept as the source column on the Goals table, but with less flexibility. I think we should add a source column to the Objectives table, it can be back populated with historical data via the Audit log.

docs/logical_data_model.puml
thewatermethod: It needs to track to a specific report.

src/migrations/20240516140827-add-created-here-column-to-activity-report-objectives.js
thewatermethod: We need to know which specific report the objective was created on for the UI, not that it was created on a report

PR #2156, opened 2024-05-21
[TTAHUB-2560] dedupe ARG and create timeseries functions, restore missing ARGFRs

hardwarehuman:
## Description of change

The first migration creates and runs a stored procedure for merging duplicate ARGs. It can and probably should be rerun in the future, but for now it's just here so that when we add missing Activity Report Goal Field Responses, there's only one target Activity Report Goal.

The second migration creates a stored procedure that accepts any audited table name (as a string literal) and uses the audit log to create an expanded version of that table containing one timeseries record for each historical state of each record in the audited table. Each record has the data as it was in that version, plus two additional columns `timeband_start` and `timeband_end` that contain the timestamps for the start and end of the band of time that the original record was in that state

The utility of this is demonstrated by the simplicity of the final query, where we join the activity report to the goal responses as they looked at the time the report was approved just by adding `ar."approvedAt" BETWEEN timeband_start AND timeband_end`

## How to test

There's a lot of different things to check here. One thing to try is making updates to your local database and then using `SELECT create_timeseries_from_audit_log('TableYouUpdated');` to generate `TableYouUpdated_timeseries` and then inspecting to see that it represented your updates faithfully.

also try running
```
SELECT DISTINCT
  arg."activityReportId" ar_id,
  arg."goalId" goal_id,
  arfr.response root_cause_on_ar,
  gfr.response root_cause_on_goal,
  ar."calculatedStatus" ar_status
FROM "Goals" g
JOIN "ActivityReportGoals" arg
  ON g.id = arg."goalId"
JOIN "ActivityReports" ar
  ON arg."activityReportId" = ar.id
LEFT JOIN "ActivityReportGoalFieldResponses" arfr
  ON arg.id = arfr."activityReportGoalId"
LEFT JOIN "GoalFieldResponses" gfr
  ON g.id = gfr."goalId"
WHERE g."goalTemplateId" = 19017
  AND ar."createdAt" BETWEEN '2023-04-28' AND '2024-02-01' ;
```
which in the original Jira shows null `root_cause_on_ar` values where `root_cause_on_goal` is not null. There should be no more of those in that time period or at any other.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2560
* https://jira.acf.gov/browse/TTAHUB-2789

### Every PR

- [ ] Meets issue criteria

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


GarrettEHill (approved):

src/migrations/20240520000000-merge_duplicate_args.js
```diff
@@ -0,0 +1,232 @@
...
+      AS
+      WITH updater AS (
+        DELETE FROM "ActivityReportGoalFieldResponses"
+        USING arg_merges
+        WHERE "activityReportGoalId" = donor_arg
+        RETURNING
+          id argfrid,
+          donor_arg
+      ) SELECT * FROM updater
+      ;
```
AdamAdHocTeam, line 130: Can you double check this delete. I created a duplicate response case and after running the delete I had no responses for that arg. I would think we would want to keep one of them.

Inserted duplicate records like this: 
  insert into "ActivityReportGoalFieldResponses" (id, "activityReportGoalId", "goalTemplateFieldPromptId", response, "createdAt", "updatedAt")
	  values 
	  (1584, 12355, 1, '{Workforce}','2024-05-17 12:35:45.871-04','2024-05-17 12:35:45.871-04' )
	  
	  	  insert into "ActivityReportGoalFieldResponses" (id, "activityReportGoalId", "goalTemplateFieldPromptId", response, "createdAt", "updatedAt")
	  values 
	  (1585, 12354, 1, '{Workforce}','2024-05-17 12:35:45.871-04','2024-05-17 12:35:45.871-04' )

src/migrations/20240520000000-merge_duplicate_args.js
```diff
@@ -0,0 +1,232 @@
...
+        JOIN unmatched u
+          ON u.donor_arg = am.donor_arg
+        WHERE argr.id = u.argrid
+        RETURNING
+          id argrid,
+          am.donor_arg original_arg
+      ) SELECT * FROM updater
+      ;
+      -- Delete duplicate objective ARGRs
+      DROP TABLE IF EXISTS deleted_argrs;
```
AdamAdHocTeam, line 183: Would also do a quick double check here inserting duplicate resource records for the same ar and goal, just making sure we keep one of them.

src/migrations/20240520000000-merge_duplicate_args.js
```diff
@@ -0,0 +1,232 @@
...
+        USING arg_merges
+        WHERE id = donor_arg
+          AND target_arg != donor_arg
+        RETURNING
+          donor_arg
+      ) SELECT * FROM updater
+      ;
+
+      END
+      $$
```
AdamAdHocTeam, line 211: Probably works from the yarn command but the $$ gives me a weird error when run in query analyzer.

hardwarehuman: The `END $$` just terminates `dedupe_args()`, so if you were stepping through the contents of the function manually to test (thank you!) then those won't make sense and will generate errors.

src/migrations/20240520000001-create-timeseries-and-root-causes.js
```diff
@@ -0,0 +1,352 @@
...
+      qry := format('
+      DROP TABLE IF EXISTS clist;
+      CREATE TEMP TABLE clist
+      AS
+      SELECT
+        column_name cname
+        ,ordinal_position cnum
+        ,data_type  ctype
+      FROM information_schema.columns ic
+      WHERE table_schema = %L
```
AdamAdHocTeam, line 62: Finding my self scratching my head on some of this stuff. Can we add some comments on some of the more cryptic lines. Even a 1 line comment would be helpful.

PR #2166, opened 2024-05-28
[TTAHUB-2986]Correct Spanish coursenames with unknown characters

hardwarehuman:
## Description of change

Spanish iPD course names with accent marks were corrupted on import. This finds and replaces them with the correct UTF-8 versions

## How to test

After the migration runs, there should be no iPD courses with unprintable characters. There should also be 10 records with `mapsTo` and `deletedAt` values set.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2986

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20240529000000-correct-spanish-course-names.js
+++ b/src/migrations/20240529000000-correct-spanish-course-names.js
@@ -1,0 +1,90 @@
+const {
+  prepMigration,
+} = require('../lib/migration');
+
+module.exports = {
+  up: async (queryInterface) => queryInterface.sequelize.transaction(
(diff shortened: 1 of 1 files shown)


thewatermethod:
This nicely corrects the course names. As far as I can tell though, we don't use the mapsTo in any queries for the courses, so any activity reports with these courses would be out of luck. Prod DB backup from Friday suggests there aren't any, so maybe we just got lucky and hurry this into production.

PR #2171, opened 2024-05-29
[TTAHUB-3022] Backup production DB on each deploy

GarrettEHill:
## Description of change

Add step in circleci to auto-backup production as part of the deployment process, So that any failed deployment can be reverted back to a known good state.

## How to test
It runs the same DB backup function that the scheduled backup does, just as part of the production deployment job. Only the production deployment job.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3022

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1069,6 +1069,20 @@ workflows:
       - dynamic_security_scan:
           requires:
             - build_and_lint
+      - backup_upload_production:
+          requires:
+            - test_backend
+            - test_frontend
+            - test_e2e
+            - test_api
+            - test_similarity_api
+            - test_utils
+            - cucumber_test
+            - dynamic_security_scan
+          filters:
+            branches:
+              only:
+                - << pipeline.parameters.prod_git_branch >>
       - deploy:
           requires:
             - test_backend


kcirtapfromspace:
What is the retention (policy) around backups? 

is it like 30 days 1 in 1 out?

GarrettEHill:
@kcirtapfromspace  We currently do not have a deletion schedule. A JIRA should be made for that

hardwarehuman (approved):
I think a YAML comment to note that there's currently no backup management might be good, but I'm also not really sure where we would put that. But I would want something in there to note that all of this was written assuming that we would fix this eventually and is thus currently a time bomb with a long and mostly predictable fuze. Maybe just creating the Jira is all we should really because commenting it is basically burying a postit note.

thewatermethod (approved):
As part of the resolution of this ticket, we should (as suggested above) write a Jira to document the need for a backup management policy and add a section to the main README explaining the backups, both how to run them on demand and that they run each production deploy, and with a "Retention policy" subheading that says "TBD."

nvms (approved):
> So that any failed deployment can be reverted back to a known good state

We talked in Slack about improving our documentation around restoring these backups in the context of cloud.gov, the various ways to do it, etc. I don't think it needs to be a part of this PR, but something we should keep in mind. The process you outlined in the RDS upgrade you handled last week was pretty involved and not documented anywhere outside of that Slack post (as far as I can tell).

PR #2177, opened 2024-05-30
Attempt to deflakify a bunch of cleanup within the backend tests

thewatermethod:
## Description of change

General cleanup: checked each of these test folders for errors/hanging runs, as well as periodic re-runs/DB spot checks. Added cleanup where necessary, as well as removing unused variables, unneeded awaits, stuff like that.

Deleted one test that was just a mess as far as cleanup; it was a good test, so I will work on rewriting it in the meantime

Not going to cure our flaky test problem (you can see that from the CI history on this branch) but hopefully helps a little

### Stuff I noticed/gotchas to be aware of

- Grant.destroys all need to run hooks or the GrantNumberLinks prevent delete. 
- Probably all destroys should run hooks by default when writing tests. There may be times when we don't want that, but I think for the most part it's a good idea
- Awareness of paranoid models. I didn't realize Topics were paranoid, for example, although they were mostly disposed of correctly. For those models: ```force: true``` gets around the soft delete. If they are already soft-deleted though, you need to remove the default scopes ```paranoid: true``` does it, as well as ```Model.unscoped()``` I think
- the helpers in testUtils are great but I did run into an issue where the destroyReport function was removing region 5 - it takes a real scorched earth approach to cleanup.

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/goalServices/getGoalsMissingDataForActivityReportSubmission.test.js
```diff
@@ -79,37 +79,43 @@ describe('getGoalsMissingDataForActivityReportSubmission', () => {
       where: {
         goalId: goalOne.id,
       },
+      individualHooks: true,
```
AdamAdHocTeam, line 82: Is there any worry that these hooks could cause unforeseen issues? Almost wondering if we want to wrap in try catch with the catch failing the test.

AdamAdHocTeam (approved):
Overall this is a improvement. 

I almost wonder if we should fail the test via try catch in the after all if it ends up in the catch for any reason (on these tests).

PR #2223, opened 2024-06-21
[TTAHUB-3027] Fix bad created vias and remove path from whence they sprang

thewatermethod:
## Description of change
- Remove overly complicated logic from goal form around setting createdVia. We only need set that if a goal is created there, there is no case for updating existing createdVias there.
- Fix two erroneously set createdVias. These were identified via the audit log.

## How to test
Confirm that an existing goal does not have a createdVia set when edited on the RTR, including a merged goal
Confirm that the migration on recent prod data sets those two goals to a createdVia of "merge" (their original value)

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3027

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/goalServices/goals.js
```diff
@@ -504,10 +505,6 @@ export async function createOrUpdateGoals(goals) {
       if (newGoal.status !== status) {
         newGoal.set({ status });
       }
-
-      if (!newGoal.createdVia || newGoal.createdVia !== createdVia) {
```
AdamAdHocTeam, line 508: Did we look into why we added this?

thewatermethod: We added it to account for goals that didn't have the createdVia yet, since we didn't write a big migration when we added the column. At this point, I feel that it is safe to discard.

thewatermethod: Specifically, the smartsheet imported goals didn't have createdVias

AdamAdHocTeam (approved):
Looks good was able to edit the a goal and have it keep its createdVia also tested case where its not set.

Script also updates values to 'merge'.

GarrettEHill (approved):

PR #2230, opened 2024-06-25
[TTAHUB-2886] Display FEI root cause on goal card

AdamAdHocTeam:
## Description of change

For FEI goals on the RTR we want to display the FEI root causes if applicable.

## How to test

- Review the code.
- Ensure we only show root causes for FEI goals.
- Ensure we show blank root causes if the FEI goal is missing them.
- Make sure we properly adjust the view on smaller browser displays.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2886

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/recipient.js
+++ b/src/services/recipient.js
@@ -514,6 +514,7 @@ export async function getGoalsByActivityRecipient(
 ) {
   // Get the GoalTemplateFieldPrompts where title is 'FEI root cause'.
   const feiRootCauseFieldPrompt = await GoalTemplateFieldPrompt.findOne({
+    attributes: ['goalTemplateId'],
     where: {
       title: 'FEI root cause',
     },
@@ -597,6 +598,7 @@ export async function getGoalsByActivityRecipient(
           ELSE 7 END`),
       'status_sort'],
       [sequelize.literal(`CASE WHEN "Goal"."id" IN (${sanitizedIds}) THEN 1 ELSE 2 END`), 'merged_id'],
+      [sequelize.literal(`"Goal"."goalTemplateId" = ${feiRootCauseFieldPrompt.goalTemplateId}`), 'isFei'],
     ],
     where: goalWhere,
     include: [
@@ -877,7 +879,7 @@ export async function getGoalsByActivityRecipient(
       onAR: current.onAR,
       sessionObjectives: [],
       responses: current.responses,
-      isFei: current.goalTemplateId === feiRootCauseFieldPrompt.goalTemplateId,
+      isFei: current.isFei,
     };
 
     goalToAdd.collaborators.push(...createCollaborators(current));
--- a/src/services/recipient.js
+++ b/src/services/recipient.js
@@ -26,6 +26,7 @@ import {
   ActivityReportCollaborator,
   ActivityReportApprover,
   ActivityReportObjective,
+  GoalTemplateFieldPrompt,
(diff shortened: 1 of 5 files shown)


src/services/recipient.js
```diff
@@ -511,6 +512,13 @@ export async function getGoalsByActivityRecipient(
     ...filters
   },
 ) {
+  // Get the GoalTemplateFieldPrompts where title is 'FEI root cause'.
+  const feiRootCauseFieldPrompt = await GoalTemplateFieldPrompt.findOne({
+    where: {
+      title: 'FEI root cause',
+    },
+  });
```
GarrettEHill, line 520: since you only use the column goalTemplateId, please filter result set to only that column

src/services/recipient.js
```diff
@@ -868,6 +876,8 @@ export async function getGoalsByActivityRecipient(
       collaborators: [],
       onAR: current.onAR,
       sessionObjectives: [],
+      responses: current.responses,
+      isFei: current.goalTemplateId === feiRootCauseFieldPrompt.goalTemplateId,
```
GarrettEHill, line 880: why not have this defined in the call done on line 576, there is no need to have it done here. with the knowledge that there will soon be other standard goals, that may each have some custom UI work. I think we need to choose a better design on how to handle the special goals.

GarrettEHill (changes requested):

src/services/recipient.js
AdamAdHocTeam: Do you mean something like this: 

      [sequelize.literal(`"Goal"."goalTemplateId" = ${feiRootCauseFieldPrompt.goalTemplateId}`), 'isFei'],

src/services/recipient.js
```diff
@@ -652,7 +662,7 @@ export async function getGoalsByActivityRecipient(
         model: GoalFieldResponse,
         as: 'responses',
         required: false,
-        attributes: ['response', 'goalId'],
+        attributes: ['response', 'goalId', 'goalTemplateFieldPromptId'],
```
GarrettEHill, line 665: now that the isFei is in the command, is this needed here?

GarrettEHill (approved):

src/services/recipient.js
AdamAdHocTeam: removed.

src/services/recipient.js
```diff
@@ -511,6 +512,14 @@ export async function getGoalsByActivityRecipient(
     ...filters
   },
 ) {
+  // Get the GoalTemplateFieldPrompts where title is 'FEI root cause'.
+  const feiRootCauseFieldPrompt = await GoalTemplateFieldPrompt.findOne({
```
thewatermethod, line 516: can we cache this somehow? seems bad to do this query this every single time we make this request, considering the result will always be the same

thewatermethod (approved):

src/services/recipient.js
AdamAdHocTeam: What are you thinking for cache?

thewatermethod: redis, or in memory

thewatermethod (approved):

PR #2268, opened 2024-07-12
[TTAHUB-3083] Integrate programs retrieve into normal sequelize pattern

AdamAdHocTeam:
## Description of change

We had put in place a temporary fix for grant programs back in 2022. This ticket revisits it and removes the un needed function.

## How to test
- Review code.
- Create a report with multi recipients with multiple programs per recipient.
- Make sure programs display correctly at every step of the approval process.
- Make sure the programs display correctly viewing approved reports and exporting to CSV.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3083

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/activityReports.js
+++ b/src/services/activityReports.js
@@ -299,30 +299,18 @@ export function activityReportByLegacyId(legacyId) {
   });
 }
 
-export async function populateRecipientInfo(activityRecipients, grantPrograms) {
-  /*
-      Hopefully this code is temporary until we figure
-      out why joining programs to grants causes issues.
-  */
+export function populateRecipientInfo(activityRecipients) {
   return activityRecipients.map((recipient) => {
-    if (recipient.grant && grantPrograms.length) {
-      const programsToAssign = grantPrograms.filter((p) => p.grantId === recipient.grantId);
-      // Programs.
-      Object.defineProperty(
-        recipient.grant,
-        'programs',
-        {
-          value: programsToAssign,
-          enumerable: true,
-        },
-      );
+    if (recipient.grant) {
+      const programsToAssign = (recipient.grant.programs || []).map((p) => p.programType);
 
       // Program Types.
       const programTypes = programsToAssign && programsToAssign.length > 0
         ? [...new Set(
-          programsToAssign.filter((p) => (p.programType))
-            .map((p) => (p.programType)).sort(),
+          programsToAssign.filter((p) => (p))
+            .map((p) => (p)).sort(),
         )] : [];
+
       Object.defineProperty(
         recipient.grant,
         'programTypes',
@@ -345,17 +333,11 @@ export async function populateRecipientInfo(activityRecipients, grantPrograms) {
       );
 
       // Name.
-      let nameValue;
-      if (recipient.grant.recipient) {
-        nameValue = `${recipient.grant.recipient.name} - ${recipient.grant.numberWithProgramTypes}`;
-      } else {
-        nameValue = `${recipient.grant.numberWithProgramTypes}`;
-      }
       Object.defineProperty(
         recipient.grant,
         'name',
         {
-          value: nameValue,
+          value: recipient.name,
           enumerable: true,
         },
       );
@@ -364,7 +346,7 @@ export async function populateRecipientInfo(activityRecipients, grantPrograms) {
         recipient,
         'name',
         {
-          value: nameValue,
+          value: recipient.name,
           enumerable: true,
         },
       );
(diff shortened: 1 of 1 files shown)


src/services/activityReports.js
```diff
@@ -299,30 +299,18 @@ export function activityReportByLegacyId(legacyId) {
...
+    if (recipient.grant) {
+      const programsToAssign = (recipient.grant.programs || []).map((p) => p.programType);
 
       // Program Types.
       const programTypes = programsToAssign && programsToAssign.length > 0
         ? [...new Set(
-          programsToAssign.filter((p) => (p.programType))
-            .map((p) => (p.programType)).sort(),
+          programsToAssign.filter((p) => (p))
+            .map((p) => (p)).sort(),
```
GarrettEHill, line 311: does this do anything: `.map((p) => (p))`, when you have already filtered it?

src/services/activityReports.js
```diff
@@ -345,17 +333,11 @@ export async function populateRecipientInfo(activityRecipients, grantPrograms) {
...
-        nameValue = `${recipient.grant.recipient.name} - ${recipient.grant.numberWithProgramTypes}`;
-      } else {
-        nameValue = `${recipient.grant.numberWithProgramTypes}`;
-      }
       Object.defineProperty(
         recipient.grant,
         'name',
         {
-          value: nameValue,
+          value: recipient.name,
```
GarrettEHill, line 340: Are you sure line 340 and 349 should be the same value? one is the recipient the other is the recipient.grant. It would make more sense if the recipient.grant was the grant number.

src/services/activityReports.js
GarrettEHill, line 1: please confirm that there is no need to update the tests

src/services/activityReports.js
```diff
@@ -364,7 +346,7 @@ export async function populateRecipientInfo(activityRecipients, grantPrograms) {
         recipient,
         'name',
         {
-          value: nameValue,
+          value: recipient.name,
```
thewatermethod, line 349: hmm why do we need to set the recipient's name to what it already is?

thewatermethod:
Testing locally, I think you could just lose the function altogether and just rely on _existing_ virtual fields, right? The below gave me something that I think has everything you originally had (although we may need more attributes on the frontend).

All I did was specify the virtual fields on the grant. Which you do not even have to do by default. Am I missing something? I also didn't check the CSV part yet

```js
  const recipients = await ActivityRecipient.findAll({
    where: {
      activityReportId: arId,
    },
    attributes: [
      'id',
      'name',
      'activityRecipientId',
      'grantId',
      'otherEntityId',
    ],
    include: [
      {
        attributes: [
          'id',
          'number',
          'numberWithProgramTypes',
          'name',
        ],
        model: Grant,
        as: 'grant',
        include: [
          {
            model: Program,
            as: 'programs',
            attributes: ['programType'],
          },
          {
            model: Recipient,
            as: 'recipient',
            attributes: ['name'],
          },
        ],
      },
    ],
  });
  ```

File: src/services/activityReports.js (hunk 1 above)
Comment (AdamAdHocTeam) [reply src/services/activityReports.js:311]: I think it checks for a value and sorts, not sure why I had it there before

File: src/services/activityReports.js
Comment (AdamAdHocTeam) [reply src/services/activityReports.js:1]: test failed after change so I dont think so.

File: src/services/activityReports.js (hunk 3 above)
Comment (AdamAdHocTeam) [reply src/services/activityReports.js:349]: Will remove.

**Comment by AdamAdHocTeam** [comment]
> Testing locally, I think you could just lose the function altogether and just rely on _existing_ virtual fields, right? The below gave me something that I think has everything you originally had (although we may need more attributes on the frontend).
> 
> All I did was specify the virtual fields on the grant. Which you do not even have to do by default. Am I missing something? I also didn't check the CSV part yet
> 
> ```js
>   const recipients = await ActivityRecipient.findAll({
>     where: {
>       activityReportId: arId,
>     },
>     attributes: [
>       'id',
>       'name',
>       'activityRecipientId',
>       'grantId',
>       'otherEntityId',
>     ],
>     include: [
>       {
>         attributes: [
>           'id',
>           'number',
>           'numberWithProgramTypes',
>           'name',
>         ],
>         model: Grant,
>         as: 'grant',
>         include: [
>           {
>             model: Program,
>             as: 'programs',
>             attributes: ['programType'],
>           },
>           {
>             model: Recipient,
>             as: 'recipient',
>             attributes: ['name'],
>           },
>         ],
>       },
>     ],
>   });
> ```

Good point let me try it.

File: src/services/activityReports.js [hunk 4]
```diff
@@ -1229,6 +1145,18 @@ async function getDownloadableActivityReports(where, separate = true) {
           {
             model: Grant,
```
GarrettEHill, line 1146: Does this need all the columns from the Grants table?

GarrettEHill (approved):

src/services/activityReports.js
AdamAdHocTeam: Not sure tbh but to play it safe for now probably.

thewatermethod (approved):

PR #2284, opened 2024-07-22
[TTAHUB-3096] Include inactive grants that fall within 60 day inactivation date

AdamAdHocTeam:
## Description of change

This change ensures that we only show inactive grants if the inactivation date is 60 days from either todays date for new reports. Or 60 days from the start date if an existing report. In Jon's previous PR he ensures that any grant that was linked via the activity recipient's table will still show in the grant list regardless of date.

## How to test

- Review the code.
- Make sure for a new report we show inactive grants within 60 days from today's date.
- Make sure for existing reports with and without a start date we show inactive grants respectively.
- Make sure Jon's fix in the previous PR still works as expected (see above description).

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3096

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/activityReports.js
+++ b/src/services/activityReports.js
@@ -1042,21 +1042,7 @@ export async function setStatus(report, status) {
  * @returns {*} Grants and Other entities
  */
 export async function possibleRecipients(regionId, activityReportId = null) {
-  // If we have an activityReportId get the startDate of the report.
-  // We have to consider that we have a report with no ActivityRecipients (ie no way to join).
-  const reportWithStartDate = activityReportId ? await ActivityReport.findOne({
-    attributes: ['startDate'],
-    where: {
-      id: activityReportId,
-    },
-  }) : null;
-
-  // If we have a valid startDate subtract 61 days from it.
-  let startDate = null;
-  if (reportWithStartDate && reportWithStartDate.startDate) {
-    startDate = moment(reportWithStartDate.startDate).subtract(61, 'days').format('YYYY-MM-DD');
-  }
-
+  const inactiveDayDuration = 61;
   const grants = await Recipient.findAll({
     attributes: [
       'id',
@@ -1097,8 +1083,13 @@ export async function possibleRecipients(regionId, activityReportId = null) {
         { ...(activityReportId ? { '$grants.activityRecipients.activityReportId$': activityReportId } : {}) },
         {
           '$grants.inactivationDate$': {
-            [Op.gte]:
-              sequelize.literal(`${!startDate ? `date_trunc('day', NOW()) - interval '61 days'` : `'${startDate}'`}`),
+            [Op.gte]: sequelize.literal(`
+          CASE
+            WHEN ${activityReportId ? 'true' : 'false'}
+            THEN (SELECT COALESCE("startDate", NOW()) - INTERVAL '${inactiveDayDuration} days' FROM "ActivityReports" WHERE "id" = ${activityReportId})
+            ELSE date_trunc('day', NOW()) - interval '${inactiveDayDuration} days'
+          END
+        `),
           },
         },
       ],
--- a/src/services/activityReports.js
+++ b/src/services/activityReports.js
@@ -1043,6 +1043,20 @@ export async function setStatus(report, status) {
  * @returns {*} Grants and Other entities
  */
 export async function possibleRecipients(regionId, activityReportId = null) {
+  // If we have an activityReportId get the startDate of the report.
+  // We have to consider that we have a report with no ActivityRecipients (ie no way to join).
+  let startDate = activityReportId ? await ActivityReport.findOne({
+    attributes: ['startDate'],
+    where: {
+      id: activityReportId,
+    },
+  }) : null;
+
+  // If we have a valid startDate subtract 61 days from it.
+  if (startDate) {
+    startDate = moment(startDate.startDate).subtract(61, 'days').format('YYYY-MM-DD');
+  }
+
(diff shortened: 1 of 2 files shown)


src/services/activityReports.js
```diff
@@ -1043,6 +1042,21 @@ export async function setStatus(report, status) {
...
+    where: {
+      id: activityReportId,
+    },
+  }) : null;
+
+  // If we have a valid startDate subtract 61 days from it.
+  let startDate = null;
+  if (reportWithStartDate && reportWithStartDate.startDate) {
+    startDate = moment(reportWithStartDate.startDate).subtract(61, 'days').format('YYYY-MM-DD');
+  }
```
GarrettEHill, line 1058: why not do this all in sql 
```suggestion
    const reportWithStartDate = activityReportId ? await ActivityReport.findOne({
      attributes: [
        'startDate',
        [sequelize.literal(`"startDate" - INTERVAL '61 days'`), 'adjustedStartDate']
      ],
      where: {
        id: activityReportId,
      },
    }) : null;

  // If we have a valid startDate subtract 61 days from it.
  const startDate = reportWithStartDate && reportWithStartDate.startDate
      ? reportWithStartDate.get('adjustedStartDate')
      :  null;
```

AdamAdHocTeam: I did initially we don't have a link to that report. Right now reports are joined through ActivityRecipients table we need to handle the case were a inactive grant is NOT linked to an AR.

src/services/activityReports.js
```diff
@@ -1080,7 +1094,13 @@ export async function possibleRecipients(regionId, activityReportId = null) {
       '$grants.regionId$': regionId,
       [Op.or]: [
         { '$grants.status$': 'Active' },
-        { '$grants->activityRecipients.activityReportId$': activityReportId },
+        { ...(activityReportId ? { '$grants.activityRecipients.activityReportId$': activityReportId } : {}) },
+        {
+          '$grants.inactivationDate$': {
+            [Op.gte]:
+              sequelize.literal(`${!startDate ? `date_trunc('day', NOW()) - interval '61 days'` : `'${startDate}'`}`),
```
GarrettEHill, line 1101:
```suggestion
           [Op.gte]: sequelize.literal(`
        CASE
          WHEN ${activityReportId ? 'true' : 'false'}
          THEN (SELECT COALESCE("startDate", NOW()) - INTERVAL '61 days' FROM "ActivityReports" WHERE "id" = ${activityReportId})
          ELSE date_trunc('day', NOW()) - interval '61 days'
        END
      `)
```

src/services/activityReports.js
GarrettEHill: just remove this whole block

hardwarehuman:
Your link in the PR description "Issue(s)" section is to the wrong Jira

src/services/activityReports.js
```diff
@@ -1043,6 +1042,21 @@ export async function setStatus(report, status) {
...
+    attributes: ['startDate'],
+    where: {
+      id: activityReportId,
+    },
+  }) : null;
+
+  // If we have a valid startDate subtract 61 days from it.
+  let startDate = null;
+  if (reportWithStartDate && reportWithStartDate.startDate) {
+    startDate = moment(reportWithStartDate.startDate).subtract(61, 'days').format('YYYY-MM-DD');
```
hardwarehuman, line 1057: I don't think this is correct business logic. If we have the actual startDate for the AR, then we just want to get the grants active on that date. It's when we do _not_ have a startDate that we hunt back 61 days from the day the AR record is opened

AdamAdHocTeam: This code has been removed. We just added a select in the case. See latest updtes.

hardwarehuman (changes requested):
I believe the 60/61 day adjustment should only apply to ARs that don't have a startDate and thus could easily have been in service of a grant that inactivated between when the activity occurred and when the AR is written. The current logic would help people create ARs attached to grants that expired before the actual activity itself took place. I don't think that was what the Jira meant.

src/services/activityReports.js
```diff
@@ -1080,7 +1080,18 @@ export async function possibleRecipients(regionId, activityReportId = null) {
...
       [Op.or]: [
         { '$grants.status$': 'Active' },
-        { '$grants->activityRecipients.activityReportId$': activityReportId },
+        { ...(activityReportId ? { '$grants.activityRecipients.activityReportId$': activityReportId } : {}) },
+        {
+          '$grants.inactivationDate$': {
+            [Op.gte]: sequelize.literal(`
+          CASE
+            WHEN ${activityReportId ? 'true' : 'false'}
+            THEN (SELECT COALESCE("startDate", NOW()) - INTERVAL '${inactiveDayDuration} days' FROM "ActivityReports" WHERE "id" = ${activityReportId})
```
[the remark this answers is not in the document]
GarrettEHill: if start date is null then it acts just like the else case with the original code, yours does not do the date adjustment at all if start date is populated.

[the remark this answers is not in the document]
GarrettEHill: I guess I see your logic, then yea. yours might be a better solution.

hardwarehuman (approved):
This looks right to me now

GarrettEHill (approved):

nvms (approved):
I think I tested all the cases. Looks good!

PR #2290, opened 2024-07-26
Update expected imports given new CSV

thewatermethod:
## Description of change
The jira ticket has great, detailed information on the change but essentially it was four-fold: 

- Import and populate the POC in the form
- Fix the audience field so it populates the form
- Fix the event duration field so that it populates the "training type" field in the form
- Add clear specifications to the admin import page so we have a reference for expected headings and content
- Automatically populate the collaborators based on the National Center column in the import

## How to test
Confirm using the XSLX file attached to the ticket that the fields are imported correctly. You may need to massage the data slightly, which I expect an engineer will do when the training reports are imported.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3240

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/config/config.js
+++ b/config/config.js
@@ -4,6 +4,8 @@ const singleLineLogger = (
   queryString,
 ) => console.log(queryString.replace(/\n/g, '\\n')); // eslint-disable-line no-console
 
+const suppressSuccessMessage = process.env.SUPPRESS_SUCCESS_MESSAGE === 'true';
+
 const connectionValidation = async (connection) => {
   try {
     /*
@@ -28,8 +30,11 @@ const connectionValidation = async (connection) => {
     };
 
     const result = await connection.query(queryConfig);
-    // eslint-disable-next-line no-console
-    console.info('Connection validated successfully');
+
+    if (!suppressSuccessMessage) {
+      // eslint-disable-next-line no-console
+      console.info('Connection validated successfully');
+    }
     return !!result;
   } catch (error) {
     // eslint-disable-next-line no-console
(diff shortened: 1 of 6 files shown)


config/config.js
```diff
@@ -4,6 +4,8 @@ const singleLineLogger = (
   queryString,
 ) => console.log(queryString.replace(/\n/g, '\\n')); // eslint-disable-line no-console
 
+const suppressSuccessMessage = process.env.SUPPRESS_SUCCESS_MESSAGE === 'true';
```
thewatermethod, line 7: This is just a convenience for testing

src/services/event.ts
```diff
@@ -496,15 +502,17 @@ export async function findAllEvents(): Promise<EventShape[]> {
 const splitPipe = (str: string) => str.split('\n').map((s) => s.trim()).filter(Boolean);
 
 const mappings: Record<string, string> = {
-  Audience: 'audience',
-  Creator: 'creator',
-  'Edit Title': 'eventName',
+  Audience: 'eventIntendedAudience',
```
GarrettEHill, line 505: with the changes to the targets of the remapping, I think this JIRA needs to include a migration file to remap the current existing objects to match the new names/structure.

thewatermethod: All the changed targets are corrections to fields that were named one way in this file and one way in the form; that is to say, eventIntendedAudience & eventDuration are already populated in existing reports. This change simply correctly imports them to begin with.

GarrettEHill: but the data is in the blob under the old name for the old records, right?

thewatermethod: The historical blobs have correct values under the correct (what we've updated to here) names.

GarrettEHill (approved):

PR #2299, opened 2024-08-03
[TTAHUB-3061] Generate processed dataset to remove PII in CI

GarrettEHill:
## Description of change
Full automation of pii free backup of production data.
Updated format of backup to support stream based creation and use.
Updated pii processing to be more time and space efficient by leveraging SQL operations more.
Clean pii from tr reports and monitoring data.

## How to test
The process takes about 1 hour can can be manually triggered by running manual-full-process as true

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3061

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/lib/s3.test.js
```diff
@@ -38,173 +52,273 @@ const VCAP_SERVICES = {
...
+    });
+
+    it('throws an error if promise rejects', async () => {
+      mockS3.deleteObject.mockImplementation(
+        () => ({ promise: () => Promise.reject(anotherFakeError) }),
+      );
+      const got = deleteFileFromS3(Key, undefined, mockS3);
+      await expect(got).rejects.toBe(anotherFakeError);
+      expect(mockS3.deleteObject).toHaveBeenCalledWith({ Bucket, Key });
+    });
```
thewatermethod, line 275: Is there some reason why you are passing undefined as bucket name? it doesn't seem to be part of the test, and passing it as undefined means it defaults to the destructured value from the s3config function, so line 274 will not pass. I'm just not clear as to what we are trying to test here.

automation/configs/process.yml
```diff
@@ -0,0 +1,9 @@
+instances: 1
```
AdamAdHocTeam, line 1: Can you explain a little what all these yml files are for and how they are being utilized.

I'm assuming they spin up to do some processing then get taken down. Is there any worry of shared resources being streched too thin?

GarrettEHill: Each is a different configuration for the automation app. It can only use one at a time. and each is only on for just long enough to do its assigned task.

no, that is why they are only on for just long enough to do their task then get shut down.

nvms (approved):

PR #2314, opened 2024-08-15
[TTAHUB-3288] Remove ghost goals from fake FEI SSDI

GarrettEHill:
## Description of change
remove ghost goals from fake FEI SSDI

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3288

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/queries/fake-fei-goals.sql
+++ b/src/queries/fake-fei-goals.sql
@@ -19,6 +19,49 @@
 * SELECT SET_CONFIG('ssdi.regionIds','[11]',TRUE);
 */
 CREATE EXTENSION IF NOT EXISTS pg_trgm;
+------------------+------------
+WITH gnames AS (
+SELECT DISTINCT g.name gname
+FROM "Goals" g
+JOIN "Grants" gr
+  ON g."grantId" = gr.id
+WHERE g."deletedAt" IS NULL
+  AND gr.deleted = FALSE
+),
+oldest_name_instances AS (
+SELECT
+  gname,
+  MIN(id) firstgoalid
+FROM gnames
+JOIN "Goals"
+  ON gname = name
+GROUP BY 1
+),
+name_creations AS (
+SELECT
+  gname,
+  firstgoalid,
+  "createdVia" first_createdvia,
+  MIN(zag.id) firstzagid
+FROM oldest_name_instances
+JOIN "Goals" g
+  ON g.id = firstgoalid
+LEFT JOIN "ZALGoals" zag
+  ON data_id = firstgoalid
+  AND new_row_data->>'name' = gname
+GROUP BY 1,2,3
+),
+name_creators AS (
+SELECT
+  gname,
+  first_createdvia,
+  u.name creator_name
+FROM name_creations
+LEFT JOIN "ZALGoals" zag
+  ON firstzagid = zag.id
+LEFT JOIN "Users" u
+  ON zag.dml_as = u.id
+)
 SELECT
   r.name,
   r.uei,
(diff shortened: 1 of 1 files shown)


AdamAdHocTeam (commented):
@GarrettEHill can you please provide a use a case in the "How to test" section.

src/queries/fake-fei-goals.sql
```diff
@@ -89,4 +89,8 @@ AND (NULLIF(current_setting('ssdi.status', true), '') IS NULL
           FROM json_array_elements_text(COALESCE(NULLIF(current_setting('ssdi.status', true), ''),'[]')::json) AS value
       ))
 GROUP BY 1,2,3,4,5,6,7,8,11,12
+-- Ghost goal filter
+HAVING NOT (g."createdVia" = 'activityReport'
+	AND COUNT(DISTINCT a.id) FILTER (WHERE a."calculatedStatus" = 'approved') = 0
```
AdamAdHocTeam, line 94: I didn't drill to much into the joins here, but why are we doing a count here instead of a simple != ?

hardwarehuman (approved):
Looks good

src/queries/fake-fei-goals.sql
hardwarehuman: It has to aggregate so that it doesn't delete goals where some related ARs are approved and some aren't.

src/queries/fake-fei-goals.sql
```diff
@@ -19,6 +19,49 @@
...
+SELECT
+  gname,
+  first_createdvia,
+  u.name creator_name
+FROM name_creations
+LEFT JOIN "ZALGoals" zag
+  ON firstzagid = zag.id
+LEFT JOIN "Users" u
+  ON zag.dml_as = u.id
+)
```
GarrettEHill, line 64: I don't think this is a valid way to do this as most of the time the goal is the result of a copy-paste with a spelling correction. Thus the first person to do it has no correlation to others who performed the same action. This is especially notable as the same incorrect value is being used across multiple regions at roughly the same time frame. So this method of finding the first user leads to false blame. Instead lets point to who attached the goal test to the grant which is what is displayed in the UI. using something like the following:
```LEFT JOIN "GoalCollaborators" gc
ON g.id = gc."goalId"
LEFT JOIN "CollaboratorTypes" ct
ON gc."collaboratorTypeId" = ct.id
AND ct.name = 'Creator'
LEFT JOIN "ValidFor" vf
ON ct."validForId" = vf.id
AND vf.name = 'Goals'
LEFT JOIN "Users" u
ON gc."userId" = u.id```

Comment (hardwarehuman) [reply src/queries/fake-fei-goals.sql:64]: They decided they want a completely different thing in there anyway.

File: src/queries/fake-fei-goals.sql [hunk 3]
```diff
@@ -27,15 +27,16 @@ SELECT
...
+  g."createdAt" "goal creation time",
   COUNT(DISTINCT a.id) FILTER (WHERE a."calculatedStatus" = 'approved') "approved reports",
   COUNT(DISTINCT a.id) FILTER (WHERE a."calculatedStatus" IN ('draft', 'submitted')) "pending reports",
   GREATEST(
     similarity(gt."templateName", g.name),
     similarity(gt."templateName", LEFT(g.name, LENGTH(gt."templateName"))),
     similarity(gt."templateName", RIGHT(g.name, LENGTH(gt."templateName")))
   ) similarity,
-  g.name
+  STRINGAGG(DISTINCT u.name, ',') "goal user list",
```
hardwarehuman, line 38:
```suggestion
  STRINGAGG(DISTINCT u.name, ';') "goal user list",
```

hardwarehuman: Safer in case on commas in names like "Lastname, Firstname"

PR #2345, opened 2024-08-30
[TTAHUB-3113] merge duped AROs

hardwarehuman:
## Description of change

This merges all the duplicate AROs we've made in history using almost the same logic as for a previous support request in https://jira.acf.gov/browse/TTAHUB-3309

It also drop the current non-unique index and replaces it with a _unique_ index for `goalId` and `activityReportId`

## How to test

There's not an easy way to see in on the frontend, but the results should look similar to 
```
            table_name            | updated_count | inserted_count | deleted_count
----------------------------------+---------------+----------------+---------------
 ActivityReportObjectiveTopics    |          1412 |             73 |          2189
 ActivityReportObjectiveCourses   |             8 |              0 |             2
 ActivityReportObjectiveFiles     |          1381 |            890 |          5318
 ActivityReportObjectiveResources |          1467 |            192 |          2827
 ActivityReportObjectives         |           596 |              0 |          2517
 ```
 Also you can run `SELECT arid, oid FROM (SELECT "activityReportId" arid, "objectiveId" oid, COUNT(*) cnt FROM "ActivityReportObjectives" GROUP BY 1,2) a WHERE cnt > 1;` to confirm there are no duplicates left

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3113

### Every PR

- [ ] Meets issue criteria

### Before merge to main

### Production Deploy

### After merge/deploy


thewatermethod:
> This doesn't follow up by applying the constraint on the table that would have prevented this, but perhaps it should? 

I think it definitely should, but just my opinion

kryswisnaskas (commented):
Nice. I verified that the SQL query after the migration doesn't return any duplicates in the ARO table.

> This doesn't follow up by applying the constraint on the table that would have prevented this, but perhaps it should? Currently it's left for the future, but that could be added if reviewers think we should also do that here.

It would be helpful to add a constraint here to prevent duplicate records. Do you anticipate a significant scope change if we include adding this constraint?

GarrettEHill:
I think the consensus is in, apply the constraint.

hardwarehuman:
@kryswisnaskas @GarrettEHill @thewatermethod Unfortunately adding the constraint blows up a bunch of tests, so I'm not sure if we want to work through that now or revert the constraint

kryswisnaskas:
> Unfortunately adding the constraint blows up a bunch of tests, so I'm not sure if we want to work through that now or revert the constraint

Thanks for catching this! Let’s revert the constraint for now and run the migration to clean up the bad data. We can then open another ticket/PR to sort work the tests and get the constraint back in place. How does that sound?

kryswisnaskas (approved):

PR #2362, opened 2024-09-11
[TTAHUB-3388] Deduplicate programs from HSES

kryswisnaskas:
## Description of change
It has been found that we've been getting duplicate program records from HSES. These duplicates prevent the data updates. The issue is being investigated by HSES, but in the meantime, this PR provides a fix on our side to deduplicate the program records before sending them to the database.

## How to test
Look at the code. 

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3388

### Every PR

- [x] Meets issue criteria
- [n/a] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/lib/updateGrantsRecipients.js
```diff
@@ -400,8 +400,13 @@ export async function processFiles(hashSumHex) {
 
       await updateCDIGrantsWithOldGrantData(cdiGrantsToLink);
 
+      // Deduplicate based on 'id'
+      const uniqueProgramsForDb = Array.from(
+        new Map(programsForDb.map((item) => [item.id, item])).values(),
```
thewatermethod, line 405: In this case, given a program type array like:
```js
[{ id: 1, name: "EHS"}, { id: 1, name: "HS"}, { id: 2, name: "HS"}, { id: 3, name: "HS"}]
```

This code will return 
```js
[{ id: 1, name: "HS"}, { id: 2, name: "HS"}, { id: 3, name: "HS"}]
```
meaning the last item with that id in the array is the one recorded. 

Is there any chance of those values diverging or are the duplicates returned from HSES always identical?

kryswisnaskas: We use their ids that need to be unique, so there shouldn't be any different records with the same `id`.  Good thought though.

The current duplicates are identical with respect to the data that we extract. The full HSES records only differ in their representation of null values for certain address fields. My guess is that's the reason they are even included in the feed. HSES is looking into it, so hopefully, it will be fixed on their end soon as well.

thewatermethod (approved):

src/lib/updateGrantsRecipients.js
kryswisnaskas: Having the duplicate records blocks our data updates though.

PR #2383, opened 2024-09-23
[TTAHUB-2589] Replace dependency on xml2json

kryswisnaskas:
## Description of change
`xml2json` has been removed in favor of `xml2js`. `topo `and `joi `have been replaced by `@hapi/topo` and `@hapi/joi`. 

## How to test
Verify tests pass.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2589

### Every PR

- [x] Meets issue criteria
- [n/a] UI review complete

### Before merge to main

- [n/a] OHS demo complete

### Production Deploy

### After merge/deploy


src/lib/updateGrantsRecipients.js
```diff
@@ -18,6 +18,12 @@ import { GRANT_PERSONNEL_ROLES } from '../constants';
 
 const fs = require('mz/fs');
 
+const parser = new xml2js.Parser({
+  explicitArray: false,
+  explicitCharkey: false,
+  trim: true,
+});
+
```
GarrettEHill, line 26:
```suggestion
const parser  = async (xmlData) => {
  const parseXml= new xml2js.Parser({
    explicitArray: false,
    explicitCharkey: false,
    trim: true,
  });

  try {
    const result = await parseXml.parseStringPromise(xmlData);
    return result; // handle the result here
  } catch (error) {
    console.error("Error parsing XML:", error);
    throw error; // rethrow or handle error appropriately
  }
};
```

GarrettEHill: Additionally, memory load still could be an issue, to avoid this using our src/lib/semaphore.js could limit concurrency to aid in controlling max memory utilization. See src/models/hooks/genericLink.js for how its used.

GarrettEHill: another way to go, would be to use the src/lib/stream/xml.ts lib we also have implmented which uses SAX for xml parsing

kryswisnaskas: > another way to go, would be to use the src/lib/stream/xml.ts lib we also have implmented which uses SAX for xml parsing

Good point on the SAX parsing approach! I think for now, given the size of the XML we're working with, xml2js is handling things well without hitting memory limits, and it keeps things simpler. If we encounter larger XML data down the road, I'd be open to revisiting the SAX approach. Would love to keep things lightweight unless we hit performance bottlenecks.

kryswisnaskas: > another way to go, would be to use the src/lib/stream/xml.ts lib we also have implmented which uses SAX for xml parsing

Hey, thanks for the suggestion! I get the concern about memory load, but for now, I think we're in a good spot, and adding semaphore.js might add more complexity than we need right now. Let's keep an eye on things, and if we start seeing memory issues, we can circle back to it.

Also, I know you've worked a lot with concurrency. Maybe we could look at simplifying this pattern going forward? Would be great to figure out how to keep things lean while still handling performance.

kryswisnaskas: > 
I see where you’re coming from with the separate` try/catch`, but since we already have error handling at a higher level, I was thinking this might add some unnecessary complexity here. Are you seeing a specific scenario where handling the error locally makes more sense than bubbling it up to the outer block?

GarrettEHill: I think we should add the try catch

thewatermethod (approved):

src/lib/updateGrantsRecipients.js
GarrettEHill: Many times having it only caught in the outer block makes it hard to locate the actual cause/location

kryswisnaskas: Good point, Garrett. I see how having a try/catch in this block could make it easier to trace where the error comes from.

That said, my goal was to keep the solution as lean as possible to avoid redundancy since we're already handling errors higher up. How about we keep it as is for now, and if we start running into issues with debugging, we can add more granular try/catch blocks later? That way, we keep things simple but can still adjust if needed

PR #2398, opened 2024-10-04
[TTAHUB-2979] Update GroupGrants when a grant is replaced

nvms:
## Description of change

When a grant is replaced, update any `GroupGrant` where the `grantId` is the id of the grant being replaced, to the value of the replacing grant.

When more than one grant replaces a single grant that is referenced by a group, a GroupGrant is created for each new replacement grant. Each newly created GroupGrant will have the same groupId.

## How to test

- Validate the logic in both the tests and the processFiles method.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-2979

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/lib/updateGrantsRecipients.js
```diff
@@ -432,6 +436,32 @@ export async function processFiles(hashSumHex) {
...
+      const replacements = await GrantReplacements.findAll({
+        where: { updatedAt: { [Op.gte]: HOUR_AGO } },
+        attributes: ['replacedGrantId', 'replacingGrantId'],
+        raw: true,
+      });
+
+      const affectedGroupGrants = await GroupGrant.findAll({
+        where: { grantId: uniq(replacements.map((g) => g.replacedGrantId)) },
+        attributes: ['id', 'grantId', 'groupId'],
+        raw: true,
```
thewatermethod, line 452: it's not a criticism but a question: I'm curious as to why we're transforming these to raw JSON, given it costs some cycles and I can't see why?

thewatermethod (approved):

src/lib/updateGrantsRecipients.js
nvms: hm yeah, I don't think I have a good reason for doing this! I probably set it for less verbose output when debugging something and forgot to pull it out :)

removed here: 587675310

AdamAdHocTeam (approved):
Looks good (although I still have to look at the other PR).

One question do we need to worry about a grant that is marked 'Inactive' and if so should it be removed from the group?

src/lib/updateGrantsRecipients.js
```diff
@@ -432,6 +436,30 @@ export async function processFiles(hashSumHex) {
...
+      });
+
+      for (const g of affectedGroupGrants) {
+        if (replacements.some((r) => r.replacedGrantId === g.grantId)) {
+          await GroupGrant.destroy({ where: { id: g.id } });
+          for (const replacement of replacements.filter((r) => r.replacedGrantId === g.grantId)) {
+            await GroupGrant.create({ groupId: g.groupId, grantId: replacement.replacingGrantId });
+          }
+        }
+      }
```
GarrettEHill, line 460: I think there is an issue here. if two grants become one grant, both old values will be destroyed, which is correct. But two copies of the replacing grant will try to be added. The create needs to be a separate deduped loop.

PR #2419, opened 2024-10-18
[TTAHUB-3502] Unwind goals on the RTR and Similarity API

AdamAdHocTeam:
## Description of change

This PR does three things: 
- One goal card per grant number, per role on RTR
- Break up goal similarity suggestions by grant
- Update similarity logic to only recommend similar goal text from the same grant

Note:  Per OHS objective roll up will remain as is for now (text and status).

## How to test

- Create a multi recipient/grant AR and approve.
- Make sure the Goal shows as expected on the RTR, one card per goal.
- Find an existing multi grant goal on the RTR make sure it shows one card per goal (and can be edited accordingly).
- Click into a similarity group make sure that all similar goal suggestions are for the same grant.
- Perform a merge and assert the result is expected based on selected goals to merge.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3502

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/services/recipient.js
```diff
@@ -805,39 +804,13 @@ export async function getGoalsByActivityRecipient(
     return goal;
   });
 
+  // reduce
   const r = sorted.reduce((previous, current) => {
```
thewatermethod, line 808: This should be a map now, not a reduce, since it is transforming an array of objects into an array of the same length/modified objects. Additionally, since there is a map called right before this reduce, you could combine the two iterations and gain a little performance.

thewatermethod (approved):
Tested with the steps and everything looks great. Made a code suggestion as well.

src/services/recipient.js
thewatermethod: Thinking a little more about this, I think you could now do the sorting in SQL but I could be wrong. Probably stuck doing the limit/offset in JavaScript though.

kryswisnaskas:
I see unrolled goals now for multiple grants 👍 That part seems to be working as designed.

Regarding merging, are we allowing the merging of unrolled goals? I ask because I found that the application does allow the merge; however, it's confusing. The output of the merge simply results in two goals with different IDs, giving the impression that nothing was accomplished.

image

AdamAdHocTeam:
> I see unrolled goals now for multiple grants 👍 That part seems to be working as designed.
> 
> Regarding merging, are we allowing the merging of unrolled goals? I ask because I found that the application does allow the merge; however, it's confusing. The output of the merge simply results in two goals with different IDs, giving the impression that nothing was accomplished.
> 
> image

@kryswisnaskas  as we discussed although this looks weird it is actually correct given the goals being merged. I attempted to implement a conditional alert based on mixed grants but seems like it will require more effort than initially thought. I would suggest simply just let them know of this case for now. Or create a different ticket to make the change correctly. Another option might be to simply put the message on the goal merge wizard.

PR #2462, opened 2024-11-08
[TTAHUB-3513] Add goal template for monitor goal so we have an id to use in CRON

AdamAdHocTeam:
## Description of change

This simply add a goal template for the montior goal with the text from the jira. The idea here is once we have an ID it will never change and we can use it in the CRON job query.

## How to test

- Review the migration.
- Make sure the migration works in both directions.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3513

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20241108102623-GrantRelationshipToActive.js
+++ b/src/migrations/20241108102623-GrantRelationshipToActive.js
@@ -1,0 +1,46 @@
+const { prepMigration } = require('../lib/migration');
+
+const goalText = 'The recipient will develop and implement a CAP/QIP to address monitoring findings.';
+module.exports = {
+  up: async (queryInterface) => queryInterface.sequelize.transaction(
+    async (transaction) => {
+      await prepMigration(queryInterface, transaction, __filename);
+      // Add monitor goal template.
+      await queryInterface.sequelize.query(
+        `INSERT INTO "GoalTemplates" (
+          hash,
+          "templateName",
+          "regionId",
+          "creationMethod",
+          "createdAt",
+          "updatedAt",
+          "lastUsed",
+          "templateNameModifiedAt"
+        ) Values (
+          MD5(TRIM('${goalText}')),
+          '${goalText}',
+          null,
+          'Curated'::"enum_GoalTemplates_creationMethod",
+          current_timestamp,
+          current_timestamp,
+          NULL,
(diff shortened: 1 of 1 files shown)


src/migrations/20241108102623-GrantRelationshipToActive.js
```diff
@@ -0,0 +1,46 @@
+const { prepMigration } = require('../lib/migration');
+
+const goalText = 'The recipient will develop and implement a CAP/QIP to address monitoring findings.';
```
GarrettEHill, line 3: Its missing the prefix, something like "(Monitoring)"

nvms (approved):

src/migrations/20241108102623-GrantRelationshipToActive.js
AdamAdHocTeam: Sure it wasn't in the ticket, but I will add it.

src/migrations/20241108102623-GrantRelationshipToActive.js
```diff
@@ -0,0 +1,46 @@
+const { prepMigration } = require('../lib/migration');
+
+const goalText = '(Monitoring) The recipient will develop and implement a CAP/QIP to address monitoring findings.';
```
GarrettEHill, line 3: That is not the posted standard goal
```suggestion
const goalText = '(Monitoring) The recipient will develop and implement a QIP/CAP to address monitoring findings.';
```

AdamAdHocTeam: This was copied directly from the JIRA ticket.

AdamAdHocTeam:
> We understand that by merging this into main, this goal is now selectable by every user of the app? I thought this was supposed to be only for specific recipients

Per @GarrettEHill we need this in ahead of the cron job to ensure we have a template id to use in the job.

thewatermethod:
That doesn't resolve the issue I brought up though, does it? It seems like if we need this goal in production now, we need to make additional provisions short of just adding it to the database. Unless I'm wrong about access to it.

PR #2468, opened 2024-11-12
[TTAHUB-3460] change Programs date cols to datetype

hardwarehuman:
## Description of change

`Programs.startDate` and `Programs.endDate` are currently VARCHARs rather than actual dates. This fixes that.

## How to test

Both the up and down should change the data types.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3460

### Every PR

- [ ] Meets issue criteria

### Before merge to main

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20241112000001-make-program-dates-datetypes.js
+++ b/src/migrations/20241112000001-make-program-dates-datetypes.js
@@ -6,8 +6,8 @@ module.exports = {
     async (transaction) => {
       await prepMigration(queryInterface, transaction, __filename);
       // Add monitor goal template.
-      await queryInterface.sequelize.query(`
-        -- change Programs startDate and endDate types to date
+      await queryInterface.sequelize.query(
+        `-- change Programs startDate and endDate types to date
         ALTER TABLE "Programs" ALTER COLUMN "startDate" TYPE date
         USING (NULLIF("startDate", '')::date)
         ;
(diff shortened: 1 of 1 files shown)


thewatermethod:
I don't think this change is sufficient - you **at least** also need to change it in the model, and possibly in other places where we use the program date

src/models/program.js
```diff
@@ -22,9 +23,14 @@ export default (sequelize, DataTypes) => {
       allowNull: false,
     },
     programType: DataTypes.STRING,
-    startYear: DataTypes.STRING,
```
GarrettEHill, line 25: did you intend to drop the start year column

GarrettEHill (commented):
You seem to also be missing the update to the LDM so you might need to run migrate to generate them.

src/models/program.js
hardwarehuman: Nope!

src/services/recipient.test.js
```diff
@@ -164,8 +164,8 @@ describe('Recipient DB service', () => {
         programType: 'EHS',
         startYear: 'Aeons ago',
         status: 'active',
-        startDate: 'today',
```
AdamAdHocTeam, line 167: Was there any case you found were this was expecting a string value like 'today'? Or is this just bad test data?

src/migrations/20241114000001-make-program-dates-datetypes.js
```diff
@@ -0,0 +1,39 @@
+const { prepMigration } = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  up: async (queryInterface) => queryInterface.sequelize.transaction(
+    async (transaction) => {
+      await prepMigration(queryInterface, transaction, __filename);
+      // Add monitor goal template.
```
nvms, line 8:
```suggestion
```

nvms (approved):

PR #2469, opened 2024-11-12
[TTAHUB-3513] Create 'system created' created via

AdamAdHocTeam:
## Issue(s)

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/migrations/20241112105155-GrantRelationshipToActive.js
```diff
@@ -0,0 +1,34 @@
+const { prepMigration, dropAndRecreateEnum } = require('../lib/migration');
```
thewatermethod, line 1: Perhaps we should change the name of this file

thewatermethod (approved):

src/constants.js
```diff
@@ -38,9 +38,11 @@ const SEARCH_RESULTS_PER_PAGE = 23;
 
 const AUTOMATIC_CREATION = 'Automatic';
 const CURATED_CREATION = 'Curated';
+const SYSTEM_GENERATED = 'System Generated';
```
GarrettEHill, line 41: Why this is an idea, the creationMethod column is for where the template came from, not how it is used. I think a new column would better serve what your trying to do here.

Maybe something like canUserInstantiate would be better. For all current rows its true and for this one row its false.

AdamAdHocTeam: its temporary for now to prevent using the new goal template.

PR #2470, opened 2024-11-12
[TTAHUB-3513] Al ttahub 3513 monitoring cron job

AdamAdHocTeam:
## Description of change

This runs a CRON job after the monitoring data is brought in via a post processing step. 

**NOTE: Before merging  we MUST update the start cut off date.**

## How to test

- Review the code.
- Manually run the job via the CLI and ensure we are creating the correct goals.

Criteria to keep in mind for when testing if a Monitoring Goal should be created:
1. **MonitoringReviews.reportDeliveryDate** is between the CUTOFF DATE (TBD) and TODAY
1.  **MonitoringReviewStatuses.name** = 'Complete'
2.  **MonitoringFindingStatuses."name"** = 'Active'
3. The **MonitoringReviews.reviewType** is one of the following:  'AIAN-DEF', 'RAN', 'Follow-up', 'FA-1', 'FA1-FR', 'FA-2', 'FA2-CR', 'Special'
4. There is a case were if the above criteria is met and there is currently a CLOSED monitoring goal for the grant we should set its status to 'Not started'.
5. Verify all fields on the newly created goals are correct

By running the first part of the query via a query analyzer you can get a count of how man goals you expect to have created.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3513

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/tools/createMonitoringGoals.js
```diff
@@ -0,0 +1,154 @@
+import {
+  sequelize,
+  GoalTemplate,
+} from '../models';
+import { auditLogger } from '../logger';
+
+const createMonitoringGoals = async () => {
+  const cutOffDate = '2024-10-01';
+  const monitoringTemplateName = '(Monitoring) The recipient will develop and implement a QIP/CAP to address monitoring findings.';
```
GarrettEHill, line 9: This is a bad design to have the goals text hard coded.

src/migrations/20241115143738-AddMonitoringEnum.js
```diff
@@ -0,0 +1,19 @@
...
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      return Promise.all(Object.values(GOAL_CREATED_VIA).map((action) => queryInterface.sequelize.query(`
+         ALTER TYPE "enum_Goals_createdVia" ADD VALUE IF NOT EXISTS '${action}';
+      `)));
+    });
+  },
```
GarrettEHill, line 14: THIS will fail in testing when GOAL_CREATED_VIA is updated in the future and will require updates to this migration file to fix. Constant values should be defined within the migration to avoid this.

GarrettEHill (changes requested):

src/tools/createMonitoringGoals.js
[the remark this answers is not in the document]
AdamAdHocTeam: Historically this is the location we have put jobs in the past including things like the importSystem.ts, populateLegacyResourceTitles.js, and updateCompletedEventReportPilots.js.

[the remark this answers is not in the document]
AdamAdHocTeam: If  there is no major reason to move this I think its fine atm.

src/tools/createMonitoringGoals.js
AdamAdHocTeam: updated.

GarrettEHill: That will fail in testing.

I have set you the following a few times now on how to get a dynamically populated unique value to use. It would need to be placed in a migration file and then the alias could be used to find them distinctly regardless of the id.

```
ALTER TABLE "GoalTemplates"
ADD COLUMN standard TEXT GENERATED ALWAYS AS (
    CASE
        WHEN "creationMethod" = 'Curated' THEN substring("templateName" from '(?:^[(]([^)]+)[)])')
        ELSE NULL
    END
) STORED;

CREATE UNIQUE INDEX unique_standard_non_null
ON "GoalTemplates"(standard)
WHERE standard IS NOT NULL;
```

Then add the new column to the model. It will be populated only on the curated goals and be the value within the ()

src/migrations/20241115143738-AddMonitoringEnum.js
GarrettEHill: The test drop and recreate the DB, when we have a new value in GOAL_CREATED_VIA  for the next feature added. your migration will be the source of an error that will be hard to track down. I have had to go back and fix about a dozen other migration files for this same structure being used. Please change it.

docker-compose.yml
```diff
@@ -101,7 +101,4 @@ services:
     command: ${ITAMS_MD_USERNAME:-tta_ro}:${ITAMS_MD_PASSWORD:-password}:1001
 volumes:
   dbdata: {}
-  minio-data: {}
-  yarn-cache: {}
-  node_modules-backend: {}
-  node_modules-frontend: {}
\ No newline at end of file
+  minio-data: {}
```
GarrettEHill, line 104: Remove this from the branch

src/migrations/20241115143738-AddMonitoringEnum.js
```diff
@@ -0,0 +1,19 @@
...
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      return Promise.all(Object.values(GOAL_CREATED_VIA).map((action) => queryInterface.sequelize.query(`
+         ALTER TYPE "enum_Goals_createdVia" ADD VALUE IF NOT EXISTS '${action}';
+      `)));
```
thewatermethod, line 12:
```suggestion
      return queryInterface.sequelize.query(`
         ALTER TYPE "enum_Goals_createdVia" ADD VALUE IF NOT EXISTS 'monitoring';
      `);
```

docker-compose.yml
AdamAdHocTeam: done.

src/migrations/20241115143738-AddMonitoringEnum.js
AdamAdHocTeam: Ok I think what your saying here is instead of recreating all the enum's simply add a new one. This was done below per Matt's comment.

src/tools/createMonitoringGoals.js
AdamAdHocTeam: I'm not sure I fully understand why the other way wouldn't have worked. I have added the suggested migration to create the 'standard' column and updated the code to use it.

docker-compose.yml
AdamAdHocTeam: done

src/tools/createMonitoringGoals.js
```diff
@@ -0,0 +1,153 @@
+import {
+  sequelize,
+  GoalTemplate,
+} from '../models';
+import { auditLogger } from '../logger';
+
+const createMonitoringGoals = async () => {
+  const cutOffDate = '2024-10-01'; // TODO: Set this before we deploy to prod.
```
nvms, line 8: Set this to what before we deploy to prod?

AdamAdHocTeam: It will be the day we deploy.

src/migrations/20241115203616-add-post-processing-to-import.js
```diff
@@ -0,0 +1,42 @@
...
+          type: Sequelize.JSONB,
+          allowNull: true,
+        },
+        { transaction },
+      );
+
+      // Update Imports set the postProcessingActions column to the object.
+      await queryInterface.sequelize.query(/* sql */`
+            UPDATE "Imports"
+            SET "postProcessingActions" = '{"name": "Monitoring Goal CRON job", "function": "createMonitoringGoals"}'
```
nvms, line 29: In `postProcess.test.js`, all of the mock records that are created define `postProcessingActions` as an array of objects, shaped like:

```ts
[
  { name: string; function: string; }
]
```

That's good. But...

Your migration above defines `postProcessingActions` as an object, not an array. So, while the tests may pass because the mocks provide good data, it seems to me that the below code (taken from `postProcess.ts`) will fail because objects do not have a `.map` method on their prototype:

```ts
    // Loop and execute post processing actions.
    await Promise.all(importRecord.postProcessingActions.map(async (action) => {
      switch (action.function) {
```

src/tools/createMonitoringGoals.test.js
```diff
@@ -0,0 +1,1780 @@
...
+    ] = grants;
+
+    // Create an inactive grant that has 'cdi' true that points to inactiveGrantThatHasBeenReplacedByActiveGrant9.
+    const cdiGrant = await Grant.create({
+      id: faker.datatype.number({ min: 9999 }),
+      number: faker.datatype.string(4),
+      recipientId: recipient.id,
+      regionId: 1,
+      startDate: new Date(Date.now() - 24 * 60 * 60 * 1000), // Set to yesterday's date
+      endDate: new Date(Date.now() - 24 * 60 * 60 * 1000), // Set to yesterday's date
```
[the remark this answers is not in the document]
AdamAdHocTeam: This seems to require I first declare yesterday variable, will keep as is for now.

src/migrations/20241115203616-add-post-processing-to-import.js
AdamAdHocTeam: I wrapped the postProcess value in [] array.

nvms (approved):

PR #2541, opened 2024-12-12
[TTAHUB-3533] Non monitoring grant warning

AdamAdHocTeam:
## Description of change

This displays an alert on the Goals and Objectives page of the AR when one or more of the grants selected doesn't have an associated monitoring goal.

image

## How to test

- Review the code
- Make sure the alert shows correctly with the appropriate text
- Make sure the alert isnt showing when it shouldn't
- Make sure the alert goes away when its no longer valid

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3533

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


thewatermethod:
I've only looked at the code but I don't think this meets the AC.

From the Jira:

When the OHS standard monitoring goal is the **only goal** on the activity report, do all grants listed in the AR summary have the goal? Invalid if: One or more grants does NOT have the OHS standard monitoring goal.

Correct me if I'm wrong but this just shows if you have a monitoring goal and a grant without a monitoring goal, it doesn't check if we have other goals, which is the case this is meant to solve.

thewatermethod:
It also seems to be missing this AC:

_If the user does not correct the warning issues, they will get a similar error alert on submission of the report, and cannot proceed until they make the correction._

AdamAdHocTeam:
> It also seems to be missing this AC:
> 
> _If the user does not correct the warning issues, they will get a similar error alert on submission of the report, and cannot proceed until they make the correction._

I thought that was this ticket: https://jira.acf.gov/browse/TTAHUB-3627

thewatermethod:
> I thought that was this ticket: https://jira.acf.gov/browse/TTAHUB-3627

According to the AC at least, these two tickets are are validating different cases

thewatermethod:
Screenshot 2024-12-16 at 9 18 49 AM

This report has two grants; one has the goal, one does not, but the alert lists them both

thewatermethod:
> So I think this is correct you are seeing it because they have no matching citation's with that grant id.

Are you sure that is how it is supposed to work? Reading the language of the alert, I think the check is supposed to be see whether the grants have the monitoring goal

AdamAdHocTeam:
> > So I think this is correct you are seeing it because they have no matching citation's with that grant id.
> 
> Are you sure that is how it is supposed to work? Reading the language of the alert, I think the check is supposed to be see whether the grants have the monitoring goal

Right now its a combo of both things:
1. must have a monitoring goal with the linked template
2. must have valid citations

thewatermethod (approved):

thewatermethod (approved):

PR #2589, opened 2025-01-09
[TTAHUB-3757] Close monitoring goals in CRON when there are no active citations or un-approved reports

AdamAdHocTeam:
## Description of change

This change adds SQL to the create monitoring job to set goals to closed if they have no active citations and they have no un-approved reports.

## How to test

- Review the code.
- Make sure all tests pass.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3757

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/tools/createMonitoringGoals.js
GarrettEHill, line 1: To support the following changes additional changes are needed to the following:
src/goalServices/changeGoalStatus.ts: ```
import * as Sequelize from 'sequelize';
import db from '../models';

interface GoalStatusChangeParams {
  goalId: number;
  userId: number; // Always required
  newStatus: string;
  reason: string;
  context: string;
  transaction?: Sequelize.Transaction;
}

export default async function changeGoalStatus({
  goalId,
  userId,
  newStatus,
  reason,
  context,
}: GoalStatusChangeParams) {
  const [user, goal] = await Promise.all([
    userId !== -1
      ? db.User.findOne({
          where: { id: userId },
          attributes: ['id', 'name'],
          include: [
            {
              model: db.Role,
              as: 'roles',
              attributes: ['name'],
              through: {
                attributes: [],
              },
            },
          ],
        })
      : null, // Skip user lookup if userId is -1
    db.Goal.findByPk(goalId),
  ]);

  if (!goal) {
    throw new Error('Goal not found');
  }

  if (!user && userId !== -1) {
    throw new Error('User not found');
  }

  const oldStatus = goal.status;

  await db.GoalStatusChange.create({
    goalId,
    userId: userId === -1 ? null : user.id, // Use null for -1, otherwise the user's ID
    userName: userId === -1 ? 'system' : user.name, // Use "system" for -1, otherwise the user's name
    userRoles: userId === -1 ? null : user.roles.map((role) => role.name), // Use null for -1, otherwise the user's roles
    oldStatus,
    newStatus,
    reason,
    context,
  });

  await goal.reload();

  return goal;
}
```

src/goalServices/changeGoalStatus.test.js: ```import faker from '@faker-js/faker';
import changeGoalStatus from './changeGoalStatus';
import db from '../models';
import { backDate } from '../lib/maintenance/common';

const fakeName = faker.name.firstName() + faker.name.lastName();
const mockUser = {
  id: faker.datatype.number(),
  homeRegionId: 1,
  name: fakeName,
  hsesUsername: fakeName,
  hsesUserId: fakeName,
  lastLogin: new Date(),
};

describe('changeGoalStatus service', () => {
  let user;
  let role;
  let goal;
  let goal2;
  let grant;
  let recipient;
  const newStatus = 'In Progress';
  const reason = 'All objectives achieved';
  const context = 'Tree planted successfully';

  beforeAll(async () => {
    user = await db.User.create(mockUser);
    recipient = await db.Recipient.create({
      id: faker.datatype.number(),
      name: faker.name.firstName(),
    });
    grant = await db.Grant.create({
      id: faker.datatype.number(),
      number: faker.datatype.string(),
      recipientId: recipient.id,
      regionId: 1,
      startDate: new Date(),
      endDate: new Date(),
    });
    goal = await db.Goal.create({
      name: 'Plant a tree',
      status: 'Draft',
      grantId: grant.id,
    });
    goal2 = await db.Goal.create({
      name: 'Plant a tree',
      status: 'Draft',
      grantId: grant.id,
    });
    role = await db.Role.create({
      id: faker.datatype.number(),
      name: 'Astronaut',
      isSpecialist: true,
    });
    await db.UserRole.create({
      userId: user.id,
      roleId: role.id,
    });
  });

  afterAll(async () => {
    await db.GoalStatusChange.destroy({ where: { goalId: [goal.id, goal2.id] }, force: true });
    await db.Goal.destroy({ where: { id: [goal.id, goal2.id] }, force: true });
    await db.GrantNumberLink.destroy({ where: { grantId: grant.id }, force: true });
    await db.Grant.destroy({ where: { id: grant.id }, individualHooks: true });
    await db.Recipient.destroy({ where: { id: recipient.id } });
    await db.UserRole.destroy({ where: { userId: user.id } });
    await db.Role.destroy({ where: { id: role.id } });
    await db.User.destroy({ where: { id: mockUser.id } });
  });

  it('should change the status of a goal and create a status change log for a valid user', async () => {
    await changeGoalStatus({
      goalId: goal.id,
      userId: mockUser.id,
      newStatus,
      reason,
      context,
    });

    const updatedGoal = await db.Goal.findByPk(goal.id);
    const statusChangeLog = await db.GoalStatusChange.findOne({
      where: { goalId: goal.id, newStatus },
      order: [['id', 'DESC']], // Order by `id` in descending order to get the newest value.
    });

    expect(updatedGoal.status).toBe(newStatus);
    expect(statusChangeLog).toBeTruthy();
    expect(statusChangeLog.oldStatus).toBe('Draft');
    expect(statusChangeLog.newStatus).toBe(newStatus);
    expect(statusChangeLog.reason).toBe(reason);
    expect(statusChangeLog.context).toBe(context);
    expect(statusChangeLog.userId).toBe(mockUser.id);
    expect(statusChangeLog.userName).toBe(user.name);
    expect(statusChangeLog.userRoles).toStrictEqual(['Astronaut']);
  });

  it('should change the status of a goal and create a status change log for system user', async () => {
    await changeGoalStatus({
      goalId: goal2.id,
      userId: -1, // System user
      newStatus,
      reason,
      context,
    });

    const updatedGoal = await db.Goal.findByPk(goal2.id);
    const statusChangeLog = await db.GoalStatusChange.findOne({
      where: { goalId: goal2.id, newStatus },
    });

    expect(updatedGoal.status).toBe(newStatus);
    expect(statusChangeLog).toBeTruthy();
    expect(statusChangeLog.oldStatus).toBe('Draft');
    expect(statusChangeLog.newStatus).toBe(newStatus);
    expect(statusChangeLog.reason).toBe(reason);
    expect(statusChangeLog.context).toBe(context);
    expect(statusChangeLog.userId).toBeNull(); // userId should be null for system user
    expect(statusChangeLog.userName).toBe('system'); // userName should be "system"
    expect(statusChangeLog.userRoles).toBeNull(); // userRoles should be null for system user
  });

  it('should throw an error if the goal does not exist', async () => {
    await expect(
      changeGoalStatus({
        goalId: 9999, // non-existent goalId
        userId: mockUser.id,
        newStatus,
        reason,
        context,
      })
    ).rejects.toThrow('Goal not found');
  });

  it('should throw an error if the user does not exist', async () => {
    await expect(
      changeGoalStatus({
        goalId: goal.id,
        userId: 9999, // non-existent userId
        newStatus,
        reason,
        context,
      })
    ).rejects.toThrow('User not found');
  });

  it('should throw an error if userId is null', async () => {
    await expect(
      changeGoalStatus({
        goalId: goal.id,
        userId: null, // Invalid userId
        newStatus,
        reason,
        context,
      })
    ).rejects.toThrow('User not found');
  });
});
```

GarrettEHill (changes requested):
I was asked not to merge in these changes, but post them into the PR as a review. The tests will need to be updated on src/tools/createMonitoringGoals.test.js

src/tools/createMonitoringGoals.js
```diff
@@ -141,12 +142,115 @@ const createMonitoringGoals = async () => {
...
+      const goalsToCloseIds = goalsToClose[0].map((goal) => goal.goalId);
+      await Goal.update({
+        status: 'Closed',
+      }, {
+        where: {
+          id: goalsToCloseIds,
+        },
+        individualHooks: true,
+        transaction,
+      });
```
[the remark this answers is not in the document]
AdamAdHocTeam: Code looks good except for one bug introduced here: goalsToOpen should be goalsToClose but this is fixed in the now pushed code.

src/goalServices/changeGoalStatus.ts
GarrettEHill, line 1: revert that last set of changes, it was fully intentional to still require a user. use -1 for system directed status changes

GarrettEHill (changes requested):

src/goalServices/changeGoalStatus.ts
```diff
@@ -3,49 +3,51 @@ import db from '../models';
...
   reason: string;
   context: string;
   transaction?: Sequelize.Transaction;
+  oldStatusToUse?: string | null;
 }
 
 export default async function changeGoalStatus({
   goalId,
-  userId = 1,
+  userId = null,
```
thewatermethod, line 16: if we are counting on a user here, do we just want to error at the top of this function if there is no user ID? I agree that I don't think 1 makes sense as a default value but the rest of these changes seem to allow a goal status change without a user, unless I'm misunderstanding them.

src/goalServices/changeGoalStatus.ts
GarrettEHill: Where is any of that being done? I don't see any code setting the user name to system other then the code you just removed.

src/goalServices/changeGoalStatus.ts
GarrettEHill: I had rewritten it to require the userId, and expected -1 to be passed when we are having the system instead of the user changing the status. But it is being reverted without providing any alternate method of setting the value to system.

src/goalServices/changeGoalStatus.ts
GarrettEHill: I think the correct answer would be to put it back, and update the other tests.

GarrettEHill: I don't this this method is as good. Please check that the other tests could not be updated to either pass a valid user or -1 and put back the other code. That way we have control of the flow and not potentially marking changes as system when it maybe an error looking up the current user.

src/goalServices/changeGoalStatus.ts
```diff
@@ -10,6 +10,39 @@ interface GoalStatusChangeParams {
   transaction?: Sequelize.Transaction;
 }
 
+export async function changeGoalStatusWithSystemUser({
```
thewatermethod, line 13: do you want to DRY this up and combine the bottom half of the two functions? 

I.E. both call basically this exact thing, maybe they could call an "updateGoalStatus" function or something:

```js
 // Change goal status.
  await db.GoalStatusChange.create({
    goalId: goal.id,
    userId: null, // For now we will use null to prevent FK constraint violation.
    userName: 'system',
    userRoles: null,
    oldStatus: goal.status,
    newStatus,
    reason,
    context,
  });

  // Reload goal.
  await goal.reload();

  //  Return goal.
  return goal;
```

I'll approve

AdamAdHocTeam: I thought about this but just decided to keep the separate.

PR #2591, opened 2025-01-10
[TTAHUB-3541] Add code coverage for src/services/recipient.js

kryswisnaskas:
## Description of change
Adds test coverage for src/services/recipient.js

image

## How to test
Verify the test coverage for src/services/recipient.js is above 90%

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3541

### Every PR

- [x] Meets issue criteria
- [n/a] UI review complete
- [n/a] QA review complete

### Before merge to main

- [n/a] OHS demo complete

### Production Deploy

### After merge/deploy


src/services/recipient.js
```diff
@@ -349,25 +349,33 @@ export async function recipientsByName(query, scopes, sortBy, direction, offset,
...
 
-  newTopics.sort();
+  newTopics.sort((a, b) => {
+    if (typeof a === 'string' && typeof b === 'string') {
+      return a.localeCompare(b);
+    }
+    if (typeof a === 'string') return -1;
+    if (typeof b === 'string') return 1;
+    return 0;
+  });
```
GarrettEHill, line 373: This wont reduce it as expected, the unique or similar functionality needs to be last or twice. topcis = ['apples', { name: "apples"}] => ["apples", "apples"]

src/services/recipient.js
```diff
@@ -381,6 +389,10 @@ function combineObjectiveIds(existing, objective) {
   return ids;
 }
 
+export function ensureArray(input) {
+  return Array.isArray(input) ? input : [];
+}
```
thewatermethod, line 394: just a thought, but maybe we could move this function to constants or somewhere else

thewatermethod (approved):

GarrettEHill (dismissed):

src/services/recipient.js
```diff
@@ -481,30 +504,41 @@ export function reduceObjectivesForRecipientRecord(
...
   return [...objectives].map((obj) => {
     // eslint-disable-next-line no-param-reassign
     obj.topics = reduceTopicsOfDifferingType(obj.topics);
     return obj;
-  }).sort((a, b) => ((
-    a.endDate === b.endDate ? a.id < b.id
-      : new Date(a.endDate) < new Date(b.endDate)) ? 1 : -1));
+  }).sort((a, b) => {
+    const dateA = a?.endDate?.trim() ? new Date(a.endDate) : new Date('1970-01-01');
+    const dateB = b?.endDate?.trim() ? new Date(b.endDate) : new Date('1970-01-01');
```
GarrettEHill, line 524: this will not be good when a.endDate = 'A'

src/services/recipient.js
kryswisnaskas: We are good here as is for the case `topics = ['apples', { name: "apples"}] => ["apples", "apples"]`. The second 'apples' topic is already handled by this logic:

```
      if (typeof topic === 'object' && topic.name) {
        return topic.name.trim();
      }
```
and deduplicated via `uniq`

Additionally, the existing test case on line 1024,` it('should handle mixed strings and objects with "name"', () => {` already tests and confirms it.

src/services/recipient.js
kryswisnaskas: Added handling of invalid dates.

src/services/recipient.js
kryswisnaskas: Moved.

src/services/recipient.js
GarrettEHill: I see the issue with how I was reading it now. the map is within the uniq.

GarrettEHill: The values from the string path do still need to be trimmed

AdamAdHocTeam (approved):

src/services/recipient.js
kryswisnaskas: > Its interesting that you are using multiple methods to unique the values, lodash.unique and [...new Set()].

Good catch! I prefer using native JavaScript functions like Set to avoid relying on external libraries, but since the code works fine as-is, I don’t see a big reason to refactor it right now. If we end up revisiting this file later, I’d be up for cleaning it up to use native methods.

kryswisnaskas: > The values from the string path do still need to be trimmed

`trim()` was added.

PR #2655, opened 2025-02-14
[TTAHUB-3860] Prevent unnecessary throwing of error from grant update - CDI replacements

AdamAdHocTeam:
## Description of change

The purpose of this ticket was to investigate an error that was being thrown during the AM grant replacement job.

"updateGrantsRecipients: Error updating grants: Expected one region and recipient for grant 9612, got 0 valid grants"

This error was being thrown when updating CDI grants to replace existing grants. The problem here is that the code doesn't take into account that there might not be any valid grant replacements for that CDI grant. So in this case the solution is to prevent an error from being thrown when there is nothing valid to process.

Upon looking at the code it's apparent that we may need to update some of the logic to handle cases when a single CDI grant replaces multiple normal grants. However I will create another ticket to check with OHS on this use case.

## How to test

- Review the code.
- Run the job and make sure the error no longer shows.

Note: I have attached the zip files to process to the ticket. You will need to create a CLI to run this in the package.json.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3860

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/lib/updateGrantsRecipients.js
+++ b/src/lib/updateGrantsRecipients.js
@@ -185,6 +185,13 @@ export const updateCDIGrantsWithOldGrantData = async (grantsToUpdate) => {
     const updates = grantsToUpdate.map(async (grant) => {
       // eslint-disable-next-line max-len
       const replacedGrants = await GrantReplacements.findAll({ where: { replacingGrantId: grant.id } });
+
+      // If we don't have any replaced grants replacements we have nothing to do for this grant.
+      // Prevent confusion of throwing exception below.
+      if (!replacedGrants.length) {
+        return;
+      }
+
       // eslint-disable-next-line max-len
       const validOldGrants = (await Promise.all(replacedGrants.map((rg) => Grant.findByPk(rg.replacedGrantId)))).filter(Boolean);
 
@@ -195,8 +202,9 @@ export const updateCDIGrantsWithOldGrantData = async (grantsToUpdate) => {
         throw new Error(`Expected one region and recipient for grant ${grant.id}, got ${validOldGrants.length} valid grants`);
       }
 
+      // eslint-disable-next-line consistent-return
       return grant.update({ recipientId, regionId });
-    });
+    }).filter(Boolean);
 
     await Promise.all(updates);
   } catch (error) {


src/lib/updateGrantsRecipients.js
```diff
@@ -195,8 +202,9 @@ export const updateCDIGrantsWithOldGrantData = async (grantsToUpdate) => {
         throw new Error(`Expected one region and recipient for grant ${grant.id}, got ${validOldGrants.length} valid grants`);
       }
 
+      // eslint-disable-next-line consistent-return
       return grant.update({ recipientId, regionId });
-    });
+    }).filter(Boolean);
```
thewatermethod, line 207:
```suggestion
    });
```

src/lib/updateGrantsRecipients.js
```diff
@@ -185,6 +185,13 @@ export const updateCDIGrantsWithOldGrantData = async (grantsToUpdate) => {
     const updates = grantsToUpdate.map(async (grant) => {
       // eslint-disable-next-line max-len
       const replacedGrants = await GrantReplacements.findAll({ where: { replacingGrantId: grant.id } });
+
+      // If we don't have any replaced grants replacements we have nothing to do for this grant.
+      // Prevent confusion of throwing exception below.
+      if (!replacedGrants.length) {
+        return;
```
thewatermethod, line 192:
```suggestion
        return Promise.resolve();
```

src/lib/updateGrantsRecipients.js
thewatermethod: no need to iterate over the list again

thewatermethod:
Should we log a message saying something like "Grant X has no replacement grants"?

AdamAdHocTeam:
> Should we log a message saying something like "Grant X has no replacement grants"?

I went back and fourth on this with Jon. I think if there is nothing to process we 

> Should we log a message saying something like "Grant X has no replacement grants"?

I went back and fourth on this, but I added something in case we need it.

src/lib/updateGrantsRecipients.js
```diff
@@ -195,8 +202,9 @@ export const updateCDIGrantsWithOldGrantData = async (grantsToUpdate) => {
         throw new Error(`Expected one region and recipient for grant ${grant.id}, got ${validOldGrants.length} valid grants`);
       }
 
+      // eslint-disable-next-line consistent-return
```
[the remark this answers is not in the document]
AdamAdHocTeam: applied

src/lib/updateGrantsRecipients.js
AdamAdHocTeam: updated.

thewatermethod (approved):

PR #2659, opened 2025-02-14
[TTAHUB-3912] Support multiple grant replacement and ensure all grants being replaced have the same region and recipient

AdamAdHocTeam:
## Description of change

So when looking into the CDI grant replacement code in updateGrantRecipients job. We currently were not supporting the updating of multiple grants becoming the single CDI replacement (many to one).

This change allows us to replace multiple grants with a CDI as long as they all have the same recipient and region id.

## How to test

- Review the code.
- Test a case or if not possible run the test.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3912

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


src/lib/updateGrantsRecipients.js
```diff
@@ -185,16 +185,30 @@ export const updateCDIGrantsWithOldGrantData = async (grantsToUpdate) => {
...
       const [recipientId] = uniq(validOldGrants.map((g) => g.recipientId));
 
-      if (!regionId || !recipientId || validOldGrants.length !== replacedGrants.length) {
+      if (!regionId || !recipientId) {
         throw new Error(`Expected one region and recipient for grant ${grant.id}, got ${validOldGrants.length} valid grants`);
       }
 
+      // Ensure allValidOldGrants have the same recipient and region.
+      if (!validOldGrants.every((g) => g.regionId === regionId && g.recipientId === recipientId)) {
+        throw new Error(`Expected all valid replaced grants to have the same recipient and region for CDI grant ${grant.id}`);
```
nvms, line 208: do we want to throw here instead of just logging and moving on to the next grant in the map? here's why I ask:

if `grantsToUpdate` contains 10 grants, and we throw on the 7th grant, that leaves 3 remaining grants that never get processed, right?

this comment applies also to the thrown error on 203

AdamAdHocTeam: So your suggesting adding another try catch inside the actual processing of each grant. This way we continue processing the remaining grants, right?

nvms: yep or just don't use promise.all - however you'd like to approach it

AdamAdHocTeam: Updated to log message and skip.

nvms (approved):

PR #2662, opened 2025-02-18
[TTAHUB-3751] remove dupe grant replacements

hardwarehuman:
## Description of change

This deletes GrantReplacements records with null grantReplacementTypeId values that are followed by another GrantReplacements record that is the same except the grantReplacementTypeId values are set.

here are also many other GrantReplacements entries where they have the same replacedGrantId and replacingGrantId but different non-null grantReplacementTypeId values. This leaves those intact because they are of potential future interest. Having said that, it's possible that those should also be deduped and we should pick only the most correct record, but that's beyond the scope of this PR.

## How to test

There's a little diagnostic query at the end that should show ~1824 deleted records now that it doesn't limit to only records on the same day

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3751

### Every PR

- [ ] Meets issue criteria
- [ ] QA review complete

### Before merge to main

### Production Deploy

### After merge/deploy


AdamAdHocTeam (commented):
Do we want to also clean up cases where the grant ids are the same. But the replacement dates are different with a null recplacement type?

image

nvms (approved):

hardwarehuman:
> Do we want to also clean up cases where the grant ids are the same. But the replacement dates are different with a null recplacement type?
> 

I think this is a pretty good idea. @nvms can you think of a reason not to do that?

nvms:
> > Do we want to also clean up cases where the grant ids are the same. But the replacement dates are different with a null recplacement type?
> 
> I think this is a pretty good idea. @nvms can you think of a reason not to do that?

I can't, but will defer to @hardwarehuman's judgement here to decide if this is potentially useful data we could use for something in the future

PR #2680, opened 2025-02-25
[TTAHUB-3615] Similarity API response failure - unexpected token

nvms:
## Description of change

If the API key environment variable that we use to authenticate to the similarity API is not defined correctly, then authentication to the API will fail. If authentication fails, then the request would return non-valid JSON, and the `return await response.json()` will throw the error mentioned in TTAHUB-3615.

If we assume that we have correctly defined the API key in cloud.gov, then we just need to check the response status before attempting to parse it as JSON. That's what this PR does.

## How to test

Make sure the change makes sense.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3615

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/services/similarity.js
+++ b/src/services/similarity.js
@@ -23,15 +23,28 @@ async function postToSimilarity(body) {
         body: JSON.stringify(body),
       },
     );
-    auditLogger.info(`${namespace} Similarity API response status: ${response.status}, body: ${JSON.stringify(response.body)}`);
+
+    // Check if response is OK before parsing JSON
+    if (!response.ok) {
+      const errorText = await response.text();
+      auditLogger.error(`${namespace} Similarity API error response: ${errorText.substring(0, 500)}`);
+      throw new Error(`API returned status ${response.status}: ${errorText.substring(0, 100)}`);
+    }
 
     return await response.json();
   } catch (error) {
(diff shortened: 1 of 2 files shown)


src/services/similarity.js
```diff
@@ -23,15 +23,28 @@ async function postToSimilarity(body) {
...
+        `${namespace} Failed to parse JSON response from Similarity API: ${error.message}`,
+        { error },
+      );
+    } else {
+      auditLogger.error(
+        `${namespace} Similarity API response failure: ${error.message}`,
+        { error },
+      );
+    }
+    throw error;
```
thewatermethod, line 47: do we need to both check response.ok and check specifically for syntax errors? What I mean is: we'd get valid JSON as long as the response is ok, wouldn't we?

thewatermethod: to be more clear, I am confused only about the purpose of the if statement starting on line 36

nvms: right - probably not necessary. I will remove this check

thewatermethod (approved):

PR #2687, opened 2025-02-26
[TTAHUB-3838] Buffer the min and max range on the horizontal bar chart

AdamAdHocTeam:
## Description of change

The smallest value on the QA FEI root cause bar chart was being cutoff. The issue is that if we set the min value for this range to be the smallest value in the set it will never show that value. I have added a small buffer for the start and end of the range.

## How to test

- Review the code
- Ensure the smallest value in the graph shows a bar
- Check other widgets that use this component make sure we didn't break them

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-3838

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/frontend/src/widgets/BarGraph.js
+++ b/frontend/src/widgets/BarGraph.js
@@ -54,7 +54,7 @@ function BarGraph({
 
   // We need to buffer the min range/scale otherwise we will cut off the smallest value.
   const minRange = Math.min(...counts);
-  const range = [minRange === 0 || minRange - 2 === 0 ? 0 : minRange - 2, Math.max(...counts) + 2];
+  const range = [minRange === 0 || minRange - 2 <= 0 ? 0 : minRange - 2, Math.max(...counts) + 2];
 
   const trace = {
     type: 'bar',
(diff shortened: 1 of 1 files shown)


frontend/src/widgets/BarGraph.js
```diff
@@ -52,7 +52,9 @@ function BarGraph({
     counts.push(dataPoint.count);
   });
 
-  const range = [Math.min(...counts), Math.max(...counts)];
+  // We need to buffer the min range/scale otherwise we will cut off the smallest value.
+  const minRange = Math.min(...counts);
+  const range = [minRange === 0 || minRange - 2 <= 0 ? 0 : minRange - 2, Math.max(...counts) + 2];
```
thewatermethod, line 57: is there a case where we wouldn't just want

```js
const range = [0, Math.max(...counts)+2];

```

thewatermethod: I mean, as the person who wrote this, I can tell you that we did not have that in mind at all

thewatermethod (approved):

PR #2756, opened 2025-04-11
TTAHUB-797 Implementing some code splitting

Andrew565:
## Description of change
Uses Dynamic Import to split out Plotly and react-admin into separate "chunks" so that the initial bundle size is greatly reduced. Takes initial download from about 5.34 MB down to 3.81 MB.

## How to test
Fire up the app and notice everything still works.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-797

## Screenshots
(Generated with source-map-explorer)
Before:
Screenshot 2025-04-11 at 5 09 11 PM
After:
Screenshot 2025-04-11 at 5 09 26 PM

### Every PR

- [X] Meets issue criteria
- [X] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


nvms (commented):
Screenshot 2025-04-15 at 3 37 08 PM

in the screenshot above I'm at the / route which does not require react-admin or plotly by any of the components rendered at that route, but these chunks are still being requested. is this the expected behavior? I thought the point was to defer loading of large libs like react-admin or plotly until they're needed.

perhaps putting an import statement at the top of the module is not the way to go. in the [examples from the create-react-app docs](https://create-react-app.dev/docs/code-splitting/) it looks like they handle importing the modules once the component that needs them is instantiated (i.e. these chunks are only requested by the client when they're actually needed). since we don't use class-based components, maybe it makes more sense to do these imports in a useEffect call?

something like this maybe?

```js
function Diag() {
  const [RA, setRA] = useState(null);
  useEffect(() => {
    import('react-admin').then(setRA);
  }, []);

  if (!RA) return null;

  // RA should be available here, so render Diag
}
```

Andrew565:
@nvms Great catch and good idea, I'll get right on that. :) *slaps forehead for not double-checking the network resources tab*

frontend/src/pages/Admin/__tests__/diag.js
```diff
@@ -38,6 +38,6 @@ describe('Diag', () => {
   it('renders the main view', async () => {
     render(<RenderDiag />);
     const requestErrorsView = await screen.findAllByText('Requesterrors');
-    expect(requestErrorsView.length).toBe(3);
+    expect(requestErrorsView.length).toBe(2);
```
Andrew565, line 41: I don't know why this dropped from 3 views to 2?

frontend/src/widgets/__tests__/TRHoursOfTrainingByNationalCenter.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
Andrew565, line 1: Adding this test to satisfy code coverage.

frontend/src/widgets/__tests__/TotalHrsAndRecipientGraph.js
```diff
@@ -50,8 +50,8 @@ describe('Total Hrs And Recipient Graph Widget', () => {
       });
     }).not.toThrow();
 
-    const svgGraph = document.querySelector('.plot-container.plotly svg');
-    expect(svgGraph).toBeInTheDocument();
+    const xAxisTitle = await screen.findByText('Date range');
```
Andrew565, line 53: Redid these tests to account for dynamic imports.

AdamAdHocTeam (commented):
Did a css selector get changed here?

Before:
image

After:
image

nvms:
> Did a css selector get changed here?
> 
> Before: image
> 
> After: image

I see this as well. I'm not sure what the fix here is, but it will need to be addressed before we merge

nvms (approved):

PR #2762, opened 2025-04-21
[TTAHUB-4057] Additional sequelizeConnection error handling

tommaroh:
## Description of change
* Based on [incident](https://ohstta.atlassian.net/wiki/spaces/TTAHUBTEAM/pages/931102721/Incident+2025.03.26.01), identified a number of places where `SequelizeConnectionAcquireTimeoutError` could still occur and was not being caught.
* This can be seen when searching [for the error](https://logs.fr.cloud.gov/goto/05e29d2eb9c53f173a17af6ff389b67a?security_tenant=hhs-acf-ohs-tta) and looking at the context in which it occurred, for instance:
`User 117 denied access due to missing SITE_ACCESS`
`Access error`
`Error finding or creating user in database -`
* Based on these log patterns, I was able to identify the callers and add additional handling.  The errors are being logged further down (ie `src/services/accessValidation.js`, `src/services/accessValidation.js`, but I have opted to make the changes at middleware layer as it has the context to call handleErrors() and handle the request fully.  Tried to add some comments and make the flow as clear as possible as this code is central to user authentication and access, please let me know if you have further suggestions.

## How to test
* This issue is difficult to reproduce, but given that this involves changes to code that handles core request/authentication flow I think it would be worthwhile to test login and permission access to resources in the Hub to verify this has not caused any regressions.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-4057

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy


Andrew565 (approved):

src/middleware/authMiddleware.js
```diff
@@ -50,11 +54,20 @@ export default async function authMiddleware(req, res, next) {
...
-    next();
-  } else {
+  } else if (!await validateUserAuthForAccess(Number(userId))
+    .catch((error) => {
+      // unrecoverable, ie database error
+      res.send(500);
+      handleErrors(req, res, error, namespace);
+      process.exit(1);
+    })) {
+    // user lacks permission to access this resource
```
nvms, line 66: a couple of things here:

- using `.catch` on an `await`ed promise is redundant. since this is already an async function, we can clean this up a bit and just use a try/catch around the call to `validateUserAuthForAccess`:

```ts
try {
  // call validateUserAuthForAccess
} catch (error) {
  // kill the process
}

instead of:

```js
if (<maybe kill the process>) {
  // we're only here because the process didn't die?
}
```

I agree we should kill the process and allow for it to restart itself, but how about something like:

```js
export default async function authMiddleware(req, res, next) {
  const userId = await currentUserId(req, res);

  if (res.headersSent) return;

  if (!userId) {
    return res.sendStatus(401);
  }

  try {
    const hasAccess = await validateUserAuthForAccess(Number(userId));
    if (!hasAccess) {
      auditLogger.warn(`User ${userId} denied access due to missing SITE_ACCESS`);
      return res.sendStatus(403);
    }
  } catch (error) {
    handleErrors(req, res, error, namespace);
    res.sendStatus(500);
    console.error('Unrecoverable error in authMiddleware. Exiting process.');
    process.exit(1);
  }

  next();
}

```

src/middleware/tokenMiddleware.js
```diff
@@ -30,7 +33,10 @@ const tokenMiddleware = async (req, res, next) => {
       title: 'Unauthenticated User',
       detail: 'User token is missing or did not map to a known user',
     });
-  } else if (await validateUserAuthForAccess(userId)) {
+  } else if (await validateUserAuthForAccess(userId)
+    .catch((error) => {
+      handleErrors(req, res, error, namespace);
+    })) {
```
nvms, line 39: exact same notes about this if statement as the other

nvms:
```js
const tokenMiddleware = async (req, res, next) => {
  const userId = await currentUserId(req, res) || await retrieveUserFromHSES(req);

  if (!userId) {
    return res.status(401).json({
      status: '401',
      title: 'Unauthenticated User',
      detail: 'User token is missing or did not map to a known user',
    });
  }

  try {
    const hasAccess = await validateUserAuthForAccess(userId);
    if (!hasAccess) {
      auditLogger.warn(`User ${userId} denied access due to missing SITE_ACCESS`);
      return unauthorized(res, 'User does not have appropriate permissions to view this resource');
    }

    auditLogger.info(`User ${userId} making API request`);
    res.locals.userId = userId;
    next();
  } catch (error) {
    // handleErrors returns a promise, and sends a 500 status to the client
    await handleErrors(req, res, error, namespace);
  }
};
```

src/middleware/authMiddleware.js
```diff
@@ -3,6 +3,19 @@ import ClientOAuth2 from 'client-oauth2';
...
+ * Authentication Middleware
+ *
+ * This middleware handles user authentication using the request session
+ * uid that gets set after successful login via HSES. Non-authenticated
+ * users who attempt to access a page that requires authentication will be
+ * served a 401
+ * @param {*} req - request
+ * @param {*} res - response
+ * @param {*} next - next middleware
+ */
```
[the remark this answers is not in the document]
tommaroh: fixed

src/middleware/authMiddleware.js
tommaroh: Given that handleErrors will already throw if it encounters `ConnectionAcquireTimeoutError`, I'm thinking let it handle and throw as anything else in-flight may get handled more cleanly instead of using process.exit()?  Otherwise took your approach, much appreciated.  Take another look and let me know what you think.

nvms: > Given that handleErrors will already throw if it encounters `ConnectionAcquireTimeoutError`, I'm thinking let it handle and throw as anything else in-flight may get handled more cleanly instead of using process.exit()? Otherwise took your approach, much appreciated. Take another look and let me know what you think.

yeah, that sounds okay to me

src/middleware/authMiddleware.js
```diff
@@ -50,11 +53,21 @@ export default async function authMiddleware(req, res, next) {
   }
 
   if (!userId) {
+    // user not found / not authenticated
     res.sendStatus(401);
```
nvms, line 57: we should return here so we don't proceed with the result of the handler - we want to prevent accidentally sending status headers twice in the same response cycle

```suggestion
    return res.sendStatus(401);
```

nvms (commented):
looks much better, but there's one thing that needs to be addressed before merge

src/middleware/authMiddleware.js
tommaroh: done

src/middleware/authMiddleware.js
tommaroh: done

nvms (approved):

AdamAdHocTeam (approved):

src/middleware/tokenMiddleware.js
tommaroh: done

tommaroh:
Deployed this to dev environment, was able to validate user authentication and permissions checking is still working as expected. Asked @fbonds to poke at it also and see if he noticed any issues.

PR #2763, opened 2025-04-21
[TTAHUB-4077] Delete unwanted objective

hardwarehuman:
## Description of change

A user who wanted to correct the title of an Objective on a Goal ended up creating a (near) duplicate instead, but they don't currently have a way to remove the extra Objective. This PR is for removing the extra.

## How to test

Before this should get back one Objective, and after zero:
`SELECT COUNT(*) FROM "Objectives" WHERE "deletedAt" IS NULL AND title = 'TTA Specialists will provide foundational information on the Head Start Director role and responsibilities that include leadership development; systems to support program planning and staffing supports; program governance; and the five-year grant application process.';`

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-4077

### Every PR

- [ ] Meets issue criteria

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

### After merge/deploy

### DIFF
--- a/src/migrations/20250421000000-delete-dupe-objective.js
+++ b/src/migrations/20250421000000-delete-dupe-objective.js
@@ -1,0 +1,26 @@
+const { prepMigration } = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      await queryInterface.sequelize.query(/* sql */`
+        -- A user created a new objective as a correction of this one
+        -- It wasn't used but they don't have a way to delete it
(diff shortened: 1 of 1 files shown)


Andrew565:
I may not understand what's going on here exactly, but it doesn't feel to me like a PR and a permanent migration file are the right way to solve what should be a one-off database fix? Is there a "business reason" why we don't just go into the database directly, run this query once, and call it good?

hardwarehuman:
> I may not understand what's going on here exactly, but it doesn't feel to me like a PR and a permanent migration file are the right way to solve what should be a one-off database fix? Is there a "business reason" why we don't just go into the database directly, run this query once, and call it good?

I think this is a good question, but I think the motivations are:
1) To avoid anyone going in and doing anything manually where they might fatfinger and cause something terrible (this is the big one) e.g. if someone accidentally had a semicolon before the WHERE in this and therefore marked *all* records as deleted. This way someone else has a chance to take a look and make sure there's no obvious mistakes before anything runs on prod.
2) To have a record of all non-user interventions.

PR #2767, opened 2025-04-22
[TTAHUB-4018] Prevent null topics while saving goals

kryswisnaskas:
## Description of change
Current Behavior
When editing goals, the frontend first retrieves the goal and objectives data from the backend, followed by a second API request to fetch topic IDs. While I wasn’t able to reproduce the issue directly, it’s possible that if the second request fails or encounters a timing issue, the topics may be sent back to the backend without IDs. This causes an error when attempting to save to the database with null topics.

This PR:

- Retrieves topic IDs as part of the initial API request to avoid dependency on a second call.

- Adds handling for scenarios where topics may still be sent to the API without IDs.

## How to test
Look at the code. Unfortunately, I wasn't able to trigger the error by using the form.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-4018

### Every PR

- [x] Meets issue criteria
- [n/a] UI review complete
- [n/a] QA review complete

### Before merge to main

- [n/a] OHS demo complete

### Production Deploy

### After merge/deploy


tommaroh (approved):
LGTM

Andrew565 (approved):

src/services/reportCache.js
```diff
@@ -108,27 +109,50 @@ export const cacheCourses = async (objectiveId, activityReportObjectiveId, cours
...
+  if (topicsNeedingLookup.length > 0) {
+    const topicNames = topicsNeedingLookup.map((t) => t.name);
+    const foundTopics = await Topic.findAll({
+      where: { name: topicNames },
+    });
+
+    // Log any that weren't found
+    const foundNames = new Set(foundTopics.map((t) => t.name));
+    const missing = topicNames.filter((n) => !foundNames.has(n));
+    if (missing.length) {
```
AdamAdHocTeam, line 129: I'm wondering how possible it would be for this case to happen.

nvms: I think the only way we could end up here is if the UI sends a topic name that doesn't exist in the database, but I'm not sure under what circumstance that would happen

src/services/reportCache.js
```diff
@@ -108,27 +109,50 @@ export const cacheCourses = async (objectiveId, activityReportObjectiveId, cours
 };
 
 const cacheTopics = async (objectiveId, activityReportObjectiveId, topics = []) => {
-  const topicIds = topics.map((topic) => {
-    if (!topic.id) {
-      auditLogger.error(`Error saving ARO topics: ${JSON.stringify(topics)} for objectiveId: ${objectiveId} and activityReportObjectiveId: ${activityReportObjectiveId}`);
+  // Find all topics with missing ids
+  const topicsNeedingLookup = topics.filter((t) => !t.id && t.name);
```
nvms, line 113: do we know why we might receive topics without an id in the first place? do we only *most of the time* receive topics with an id?

src/services/reportCache.js
kryswisnaskas: We’ve seen that our frontend does occasionally send topics without IDs, and my current theory is that it’s due to the initial data loaded from the API not including them (i.e., the backend only returned name).

PR #2843, opened 2025-06-10
[NO-TICKET] Add more local dev info to README

tommaroh:
## Description of change

* Make it clear that installing `cross-env` is a prerequisite when running locally
* Update links to dev environments
* Update info for adding a new env var cfg/secret for deployed environments
* Add steps for deploying a remote branch to dev env
* Reorganize some infra / CI information
* Move advice for creating a deploy key further down, this is rarely necessary

## How to test

View file in preview mode


tommaroh:
> This looks good @tommaroh -- one thing I may suggest, there is still a comment around setting this value in [.env.example](https://github.com/HHS/Head-Start-TTADP/blob/a738ff8af326f91cf6cf7a4001157d86cda3002c/.env.example#L4) that suggests to set the user & group values ( this is actually what led me to do this during my setup). So I'd recommend maybe removing this comment, or making it clear its optional.

Thanks for catching this.  I removed it from the `.env.example` and added a bit to the troubleshooting section in case it's still relevant for someone in the future.

PR #2874, opened 2025-07-07
[TTAHUB-4228, TTAHUB-4229] Process geographic regions and add geo groups creation

kryswisnaskas:
## Description of change
This PR implements the backend functionality to:

- Process and store geographic data sets for grants, including State IDs and geo tags.
- Create "My Groups" automatically for each geographic region based on these geo tags.
- Add the new geo groups with the following names:
  - Northeast
  - Midwest
  - West
  - AIAN
  - Southeast
  - Southwest

- Adds 2 users as creator and a collaborator to all 6 groups.
- Includes logic to update these groups as grants become inactive and replaced.

## How to test
Verify the tests pass

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-4229

### Every PR

- [x] Meets issue criteria
- [n/a] UI review complete
- [n/a] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


src/lib/updateGrantsRecipients.js
```diff
@@ -224,6 +225,103 @@ export const updateCDIGrantsWithOldGrantData = async (grantsToUpdate) => {
...
+      transaction,
+    });
+    const regionMap = new Map(regionGroups.map((g) => [g.name, g.id]));
+    // Get all replacing grants to protect them
+    const replacingGrantIds = (await GrantReplacements.findAll({
+      attributes: ['replacingGrantId'],
+      transaction,
+    })).map((r) => r.replacingGrantId);
+
+    const safeReplacingIds = replacingGrantIds.length > 0 ? replacingGrantIds : [-1];
```
AdamAdHocTeam, line 247: Why have an array with -1 when we have an array with no length? Just seems a bit redundant but maybe I haven't found the usage yet.

mhaley37: Had same thought, as I would expect this to be an empty array ( or undefined) rather than `[-1]`. Is a special value when seen somewhere ( `safeReplacingIds` is referenced one other time as part of the replacements of the `sequalize` query.
```

      // Remove stale GroupGrants (excluding replacing grants)
      const result = await sequelize.query(

File: src/lib/updateGrantsRecipients.js [hunk 2]
```diff
@@ -347,6 +445,8 @@ export async function processFiles(hashSumHex) {
           grantSpecialistEmail: valueFromXML(g.grants_specialist_email),
           annualFundingMonth: valueFromXML(g.annual_funding_month),
           inactivationReason: valueFromXML(g.inactivation_reason),
+          geographicRegion: valueFromXML(g.geographic_region),
+          geographicRegionId: parseInt(g.geographic_region_id, 10) || null,
```
mhaley37, line 449: `??` feels like a better fit here than `||` given you likely don't want a `0` to be treated as falsey.

AdamAdHocTeam (approved):
Overall looks good left some minor comments.

mhaley37 (approved):
Forgot to approve, but this all looks good just left a comment.

src/lib/updateGrantsRecipients.js
kryswisnaskas: 0 isn’t valid right now, but I refactored that line. Used `condition ? parseInt(...) : null` instead of `parseInt() ?? null` since `parseInt `can return `NaN `for empty/missing values, and we want null in the db.

src/lib/updateGrantsRecipients.js
kryswisnaskas: Good question! The `[-1]` is a safety hack to avoid SQL errors when using `WHERE ... NOT IN (:ids)`. If we pass an empty array to sequelize replacements, Postgres will throw a syntax error (`NOT IN ()`). Using `[-1]` ensures the query is always valid and will never filter out any real grantId.

src/lib/updateGrantsRecipients.js
kryswisnaskas: Reverting to previous `parseInt(...) || nul`l to avoid inserting `'NaN'` values and preserve correct `null `mapping for invalid ids.

PR #2907, opened 2025-07-24
[TTAHUB-4348] Fix Google Analytics flag variable

tommaroh:
## Description of change

Typo during a refactor of the build process and env vars resulted in `REACT_APP_GTM_ENABLED` not being populated to the application in dev and prod envs.  This resulted in analytics not being captured by Google Tag Manager (GTM). 
* Fix the typo and include all the `REACT_APP_` vars in app manifest used for cf push
* Other manifest changes are due to sorting the list of env vars.
* Add axios to yarn audit list

## How to test

I am now seeing the correct values show up in cloud.gov console after a deploy to dev.  I still do not see analytics showing up in the development environment, however this may be due to our switch to the colored dev environments.  May need to deploy this out to production to fully validate the fix.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-4348

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete


tommaroh:
> Quick heads-up: since these are `REACT_APP_*` variables, they need to be available at build time to get picked up by the frontend. It looks like we used to push them during the React build step, but that doesn’t seem to be happening anymore.
> 
> We’ll need to make sure these are passed in during the build process (e.g., through CI), otherwise GTM won’t actually get injected.

They already are (have been) available during the frontend build.  The site is actually rebuilt during the deploy step, the env vars are set by the `set_env_vars: target_env: << parameters.target_env >>` step, that happens before the frontend build.  The problem is that there was an extra $ in the dev and prod deploy_vars file.  Staging was fine, and we see analytics from staging, so the build process doesn't seem to be an issue.  If we don't want the REACT_APP variables exposed at runtime, I can take them back out of the manifest.

thewatermethod (approved):

PR #2936, opened 2025-08-12
[TTAHUB-4371] Add script for connecting to replica DB

tommaroh:
## Description of change

* This is a workaround script for connecting to a replica DB in cloud.gov.  We currently have a replica running in prod and in dev-pink
* The usual `connect-to-service` plugin does not work for replica databases unfortunately, so I have created something equivalent.  This script gets the db info via VCAP_SERVICES on the app host, sets up an SSH tunnel through host in a background process, then starts a psql connection to this given address.  
* It will catch exit signals to the main process and attempt to close the SSH tunnel.  I have validated this works in a variety of cases myself, but just in case it doesn't it will a) print the PID of the tunnel process when created b) cleanup any previous zombie processes it finds on startup
* We still need more for TTAHUB-4371, but this is an important first step.

## How to test

```
> ./bin/replica-connect.sh tta-smarthub-dev-pink
Establishing SSH tunnel with PID: 38115
Connecting to db replica for tta-smarthub-dev-pink on port 5432...
psql (14.18 (Homebrew), server 15.12)
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
Type "help" for help.

cgawsbrokerprodbt584djy6n6cnuz=>
\q
Cleaning up...
SSH tunnel closed.
```

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-4371

### Every PR

- [x] Meets issue criteria
- N/A Meets accessibility standards (WCAG 2.1 Levels A, AA)
- N/A API Documentation updated
- N/A Boundary diagram updated
- N/A Logical Data Model updated
- N/A [Architectural Decision Records](https://adr.github.io/) written for major infrastructure decisions
- N/A UI review complete
- N/A QA review complete

### Before merge to main

- [ ] OHS demo complete

### DIFF
--- a/bin/replica-connect.sh
+++ b/bin/replica-connect.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ $# -eq 0 ]; then
+    echo "Usage: $0 app_name"
+    echo "Example: $0 tta-smarthub-prod"
+    exit 1
+fi
+
+APP_NAME="$1"
+
+RAW_DATA=$(cf ssh $APP_NAME -c 'echo $VCAP_SERVICES')
+
+# parse
+REPLICA_HOST=$(echo $RAW_DATA | jq --raw-output '.["aws-rds"][0] | .credentials | .replica_host')
+
+PORT=$(echo $RAW_DATA | jq --raw-output '.["aws-rds"][0] | .credentials | .port')
+USER=$(echo $RAW_DATA | jq --raw-output '.["aws-rds"][0] | .credentials | .username')
+PASSWD=$(echo $RAW_DATA | jq --raw-output '.["aws-rds"][0] | .credentials | .password')
+DB_NAME=$(echo $RAW_DATA | jq --raw-output '.["aws-rds"][0] | .credentials | .db_name')
+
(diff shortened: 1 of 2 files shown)


bin/replica-connect.sh
```diff
@@ -0,0 +1,45 @@
...
+    exit 0
+}
+
+trap 'cleanup' EXIT
+
+tunnel_cmd="cf ssh -N -L ${PORT}:${REPLICA_HOST}:${PORT} $APP_NAME"
+$tunnel_cmd &
+SSH_PID=$!
+echo "Establishing SSH tunnel with PID: $SSH_PID"
+sleep 3  # Wait for the SSH tunnel to establish
```
kryswisnaskas, line 41: Is the 3 second wait sufficient, or would it be more reliable to replace it with a loop that checks when the tunnel is ready?

thewatermethod (approved):
works!

bin/replica-connect.sh
tommaroh: It's going to add some complication to try that, how do you feel about starting with the sleep and moving to a loop check later if it seems necessary?  So far 3 seconds seems sufficient in my testing

PR #2973, opened 2025-08-25
[NO-TICKET] Auto-retry CircleCI from failed step 3x

tommaroh:
## Description of change

Noticed that CircleCI released a new feature: auto-retries.  Due to the fact that our backend tests are still somewhat flaky, I have enabled 3 automatic retries on this job and the db restore job.  This will function the same as "retry from failed" behavior, which is what we would usually click when the normal build_test_deploy job.

Reference:
https://circleci.com/changelog/automatic-retry-of-workflows-released/

## How to test

Validated job runs as normal, and that jobs will retry as expected.

## Issue(s)

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,7 @@ workflows:
     # It will build the app, run all tests, then deploy if the branch matches
     # autodeploy: main->staging, production->prod
     when: pipeline.parameters.action == "build_test_deploy"
+    max_auto_reruns: 3
     jobs:
     - build_and_lint
     - build_similarity_api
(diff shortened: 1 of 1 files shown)


AdamAdHocTeam (commented):
What would happen if its a valid failure or a mixed bag of flakey and valid failures. Im not sure that we want to always retry but I could be wrong.

tommaroh:
> What would happen if its a valid failure or a mixed bag of flakey and valid failures. Im not sure that we want to always retry but I could be wrong.

If it's a consistent failure then the job will still fail.  My sense is that right now the backend tests are so intermittently flaky that the first thing people do is retry the run anyway.  Until we can resolve the intermittent failures to a more deterministic state, this seems like a reasonable option.

PR #3140, opened 2025-11-24
[TTAHUB-4587] Upgrade AWS SDK to v3

tommaroh:
## Description of change

* Upgrade AWS SDK node module to v3.  This is a fairly significant change to AWS SDK, of particular note is that each service is now a separate module under the main aws-sdk package
* Existing tests needed to be rewritten
* Change tsconfig to include all type definitions (the default) instead of only jest.  Source: https://www.typescriptlang.org/tsconfig/#types
> types: This property, also within compilerOptions, is an array of strings representing the names of specific type definition packages that should be included. When this field is present, TypeScript will only include the types from the packages listed, overriding the default behavior of including all visible @types packages.

## How to test

* Automated tests pass
* Validate s3-dependant behavior on the site works as expected
* Tested file upload, successful scan, file delete
Screenshot 2025-12-05 at 10 31 42 AM

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-4587

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy

### DIFF
--- a/src/models/file.js
+++ b/src/models/file.js
@@ -1,6 +1,6 @@
 const { Model } = require('sequelize');
 const { getSignedDownloadUrl } = require('../lib/s3');
-const { afterCreate, afterFind, afterDestroy } = require('./hooks/file');
+const { afterDestroy } = require('./hooks/file');
 
 export default (sequelize, DataTypes) => {
   class File extends Model {
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -4,6 +4,7 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "outDir": "../build/server",
+    "skipLibCheck": true,
     "incremental": true,
     "resolveJsonModule": true, 
     "noErrorTruncation": true,
@@ -11,7 +12,6 @@
     "declaration": false,
     "allowJs": true,
     "removeComments": true,
-    "skipLibCheck": true
   },
   "include": [
     "./**/*.js",


tommaroh:
> @tommaroh I assume this is still in flight and not yet ready for another look?

It's ready for another look now.  Uploads appear to be working, although the ClamAV integration is dev seems to be broken again so they aren't scanning.
Screenshot 2025-12-02 at 2 40 12 PM
Screenshot 2025-12-02 at 2 40 24 PM
Screenshot 2025-12-02 at 2 43 36 PM

src/lib/s3.js
```diff
@@ -28,60 +37,61 @@ const generateS3Config = () => {
...
       },
     };
   }
 
   // Return null if S3 is not configured
   return {
-    bucketName: null,
-    s3Config: null,
+    s3Bucket: null,
+    s3Config: { region: 'us-gov-west-1' },
```
thewatermethod: Should this be stored in an env? (I would think so)

src/models/file.js
```diff
@@ -77,15 +77,13 @@ export default (sequelize, DataTypes) => {
     },
     url: {
       type: DataTypes.VIRTUAL,
```
thewatermethod: It looks like we are trying to move this to a DB column, I would think we'd need a migration to 1) set this to a text column and 2) move this work to the hooks so it is persistent. 

Is the presigned url static? can it be stored like this?

thewatermethod: I just wouldn't think that would work, that the presigned url would be ephemeral to that session

src/lib/s3.js
thewatermethod: Oh I missed that context, my bad, fine to leave IMO

src/models/file.js
tommaroh: Found another approach, using a support library which can sign requests synchronously so we can keep the existing behavior.  Other approaches were going to involve widespread code changes I would prefer to avoid.  Tested successfully

PR #3491, opened 2026-04-08
[TTAHUB-5144] Fix goal categories on training reports list

thewatermethod:
## Description of change

Missed `raw: true` on a sequelize query, which meant that the results were not properly aggregated. The table was also counting events and not sessions. Revised both of these things and reworked the test to validate more strictly

Also fixed the `<SimpleSortableTable>` to accurately sort number values via configuration; previously, it was sorting lexicographically.

## How to test

Checkout prod data. Confirm that the "goal categories on training reports" widget accurately sorts numeric values and represents 

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5144

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


tommaroh (approved):

AdamAdHocTeam (approved):

frontend/src/widgets/StandardGoalList.js
```diff
@@ -16,7 +16,7 @@ export function StandardGoalsListTable({
   // Define the columns for the SimpleSortableTable
   const columns = [
     { key: 'name', name: 'Goal category' },
-    { key: 'count', name: '# of reports' },
+    { key: 'count', name: '# of reports', sortType: 'number' },
```
kryswisnaskas, line 19: Claude said: The PR changes the backend metric from `COUNT(DISTINCT sessionReports.eventId)` (one count per event) to `COUNT(DISTINCT sessionReports.id)` (one count per session), but this column header still says `'# of reports'`. An event with three complete sessions will now display as `3` instead of `1`. If counting sessions is the intended fix, this label should be updated — e.g., `'# of sessions'` — to avoid misleading users who compare this widget against other report-level metrics elsewhere in the dashboard.

src/widgets/trStandardGoalList.ts
```diff
@@ -50,15 +50,12 @@ export default async function trStandardGoalList(scopes: IScopes): Promise<{ nam
...
     ],
     order: [
-      [sequelize.fn('COUNT', sequelize.fn('DISTINCT', sequelize.col('sessionReports.eventId'))), 'DESC'],
+      [sequelize.fn('COUNT', sequelize.fn('DISTINCT', sequelize.col('sessionReports.id'))), 'DESC'],
       ['standard', 'ASC'],
     ],
-  })).map((gt: {
-    toJSON: () => ({ name: string, count: number })
-  }) => gt.toJSON());
+    raw: true,
```
kryswisnaskas, line 59: Claude said: Adding `raw: true` means Sequelize no longer wraps results in model instances, so the pg driver returns PostgreSQL's `COUNT()` BIGINT as a **string**, not a number. The TS return type on line 8 (`Promise<{ name: string; count: number }[]>`) is now a type lie at runtime, and `StandardGoalList.js` line 53 also declares `count: PropTypes.number` — same problem.\n\nThe tests confirm the actual runtime type: every assertion wraps the value in `Number()` (e.g. `Number(teachingPracticesResult.count)`) — that wrapper would be unnecessary if `count` were already a number.\n\nThe earlier Copilot comment on this was correct; the author's dismissal ("perhaps Sequelize is handling the BIGINT → string → INT conversion") is not accurate — no such coercion happens with `raw: true`.\n\n**Fix — cast in the query:**\n```ts\nsequelize.cast(\n  sequelize.fn('COUNT', sequelize.fn('DISTINCT', sequelize.col('sessionReports.id'))),\n  'integer'\n)\n```\nOr map before returning:\n```ts\nreturn db.GoalTemplate.findAll({ ..., raw: true })\n  .then((rows: any[]) => rows.map(r => ({ name: r.name, count: Number(r.count) })));\n```\nThen update the return type on line 8 and the `PropTypes` in `StandardGoalList.js` to match.

frontend/src/widgets/StandardGoalList.js
thewatermethod: Session reports are reports

src/widgets/trStandardGoalList.ts
thewatermethod: be0368e

PR #3498, opened 2026-04-10
Fix bug with goals table print goals

thewatermethod:
## Description of change

Found while testing out the prior implementation of select checkboxes > print to PDF. Didn't want to pollute the parent PR by including this fix there

## How to test
Go to RTR > Goals and objectives
Select a handful of goals. 
Click preview and print selected
Confirm your selection is all that is displayed

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy

### DIFF
--- a/src/services/standardGoals.ts
+++ b/src/services/standardGoals.ts
@@ -843,11 +843,13 @@ export async function standardGoalsForRecipient(
   onlyApprovedObjectives = false,
 ) {
   const { goal: scopes } = await filtersToScopes(filters, {});
+  const validGoalIds = [goalIds].flat().filter((id) => !Number.isNaN(Number(id)));
 
   const goals = await Goal.findAll({
     attributes: ['id'],
     where: {
       [Op.and]: [
+        ...(validGoalIds.length ? [{ id: validGoalIds }] : []),
         ...scopes,
         sequelize.where(
           sequelize.col('Goal.id'),
--- a/src/services/standardGoals.ts
+++ b/src/services/standardGoals.ts
@@ -843,7 +843,25 @@ export async function standardGoalsForRecipient(
   onlyApprovedObjectives = false,
 ) {
   const { goal: scopes } = await filtersToScopes(filters, {});
-  const validGoalIds = [goalIds].flat().filter((id) => !Number.isNaN(Number(id)));
+
+  const rawGoalIds = [goalIds].flat();
+  const validGoalIds = rawGoalIds.filter((id) => Number.isInteger(Number(id)) && Number(id) > 0);
+
+  // If the caller explicitly passed IDs but none were valid, return empty rather than all goals
+  if (rawGoalIds.length && !validGoalIds.length) {
+    return {
+      count: 0,
+      goalRows: [],
+      statuses: {
+        total: 0,
+        [GOAL_STATUS.NOT_STARTED]: 0,
+        [GOAL_STATUS.IN_PROGRESS]: 0,
+        [GOAL_STATUS.CLOSED]: 0,
+        [GOAL_STATUS.SUSPENDED]: 0,
+      },
+      allGoalIds: [],
+    };
+  }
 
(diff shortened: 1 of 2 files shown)


src/services/standardGoals.ts
```diff
@@ -843,11 +843,13 @@ export async function standardGoalsForRecipient(
   onlyApprovedObjectives = false,
 ) {
   const { goal: scopes } = await filtersToScopes(filters, {});
+  const validGoalIds = [goalIds].flat().filter((id) => !Number.isNaN(Number(id)));
 
   const goals = await Goal.findAll({
     attributes: ['id'],
     where: {
       [Op.and]: [
+        ...(validGoalIds.length ? [{ id: validGoalIds }] : []),
```
kryswisnaskas, line 870: Claude said: **Finding 1 — Invalid `goalIds` silently returns all goals instead of none**

When every submitted ID fails the `Number.isNaN` check (e.g. `?goalIds=abc`), `validGoalIds` ends up empty and this condition skips the filter entirely — returning every goal for the recipient. That is indistinguishable from the "no filter requested" case (`goalIds` defaulting to `[]`), so a corrupted or non-numeric selection in the print view silently shows the full list instead of surfacing an error.

Additionally, `Number("") === 0` and `Number(null) === 0` both pass `!Number.isNaN(...)`, so an empty-string query param (`?goalIds=`) is included as ID `0`, which matches no real rows — producing an empty print view with no indication of why.

Suggested fix:
```ts
const rawGoalIds = [goalIds].flat();
const validGoalIds = rawGoalIds.filter(id => Number.isInteger(Number(id)) && Number(id) > 0);

// If the caller explicitly passed IDs but none were valid, return empty rather than all goals
if (rawGoalIds.length && !validGoalIds.length) {
  return { count: 0, goalRows: [], statuses: { total: 0, ... }, allGoalIds: [] };
}
```

src/services/standardGoals.ts
```diff
@@ -843,11 +843,13 @@ export async function standardGoalsForRecipient(
   onlyApprovedObjectives = false,
 ) {
   const { goal: scopes } = await filtersToScopes(filters, {});
+  const validGoalIds = [goalIds].flat().filter((id) => !Number.isNaN(Number(id)));
```
kryswisnaskas: Claude said: **Finding 3 — No test coverage for the new `goalIds` filtering behavior**

The existing integration test suite in `src/services/standardGoal.test.js` already creates goals with known IDs inside a `beforeAll` block, so tests for this new branch would be low-effort to add. The scenarios that need coverage:

- Passing a single valid `goalId` returns only that goal (not others for the same recipient)
- Passing a subset of valid IDs returns only that subset
- Passing all-invalid IDs (e.g. `'abc'`) does not silently return all goals
- Passing an empty array (the default) continues to return all goals unchanged

Without these, the silent-fallback bug in Finding 1 would not have been caught by tests even if it had been written as a regression.

kryswisnaskas:
Trying out an additional review pass using AI (Claude/Codex) to help catch edge cases and things that are easy to miss.

Goal here is to speed up reviews and catch things earlier. Open to feedback if this becomes noisy or not useful.

src/services/standardGoals.ts
thewatermethod: In most cases there are no valid IDs sent, since the print view and the RTTAPA both use this service. This is implemented as intended

thewatermethod: Well, I went ahead and added the suggestion: e89b19e

src/services/standardGoals.ts
thewatermethod: 6592b4d

kryswisnaskas (approved):
👍

tommaroh (approved):

PR #3508, opened 2026-04-15
[TTAHUB-5155] Finding category hotspot graph backend

thewatermethod:
## Description of change

Add backend for finding category hotspot graph 
Small refactor to add a utility function for continuous month strings

## How to test
It uses the defined calculation: An approved activity report is counted within a specific month based on the start date of the activity.

It otherwise appears valid and correct, and matches other implementation 

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5155

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**
- [ ] Reviewer added _(after transitioning to Open to ensure Slack notifications trigger; `elainaparish` is the authorized approver under normal circumstances)_

### After merge/deploy


src/widgets/monitoring/reportCountByFindingCategory.ts
```diff
@@ -0,0 +1,112 @@
...
+      {
+        model: Citation,
+        as: 'citationModel',
+        attributes: ['guidance_category'],
+        required: true,
+      },
+    ],
+    raw: true,
+    nest: true,
+  }) as CitationRow[];
```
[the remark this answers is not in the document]
thewatermethod: @copilot apply changes based on [this feedback](https://github.com/HHS/Head-Start-TTADP/pull/3508#discussion_r3087122974)

Andrew565 (approved):

src/widgets/monitoring/reportCountByFindingCategory.ts
```diff
@@ -0,0 +1,96 @@
...
+  const rows = await sequelize.query<AggregatedRow>(
+    `SELECT
+      COALESCE(c.guidance_category, :noCategory) AS guidance_category,
+      TO_CHAR(DATE_TRUNC('month', ar."startDate")::date, 'YYYY-MM-DD') AS month_start,
+      COUNT(DISTINCT ar.id)::int AS report_count
+    FROM "ActivityReports" ar
+    JOIN "ActivityReportObjectives" aro ON aro."activityReportId" = ar.id
+    JOIN "ActivityReportObjectiveCitations" aroc ON aroc."activityReportObjectiveId" = aro.id
+    JOIN "Citations" c ON c.id = aroc."citationId"
+    WHERE ar.id IN (:approvedReportIds)
```
kryswisnaskas, line 72: Codex said: This does the work in two round-trips: first it loads every approved report ID into application memory, then it sends that whole array back to Postgres through `IN (:approvedReportIds)`. That creates a scale risk for wide filters and also opens a consistency window where the approved-report set and the aggregation are no longer computed from the same snapshot. I would push the AR filtering into the aggregation query itself, either with a scoped subquery/CTE or by reusing the existing `scopeToWhere(ActivityReport, ...)` pattern so the database can do this in one pass.

src/widgets/monitoring/reportCountByFindingCategory.test.js
```diff
@@ -0,0 +1,406 @@
...
+      { guidance_category: 'Health', month_start: '2025-03-01', report_count: 1 },
+    ]);
+
+    const data = await reportCountByFindingCategory({ activityReport: [] });
+
+    const health = data.find((d) => d.name === 'Health');
+    expect(health).toEqual({ name: 'Health', months: ['Jan 2025', 'Feb 2025', 'Mar 2025'], counts: [1, 0, 1] });
+  });
+
+  it('groups citations with null guidance_category under "No finding category assigned"', async () => {
```
kryswisnaskas, line 313: Codex said: The test file covers the happy-path shaping pretty well, but it still misses a few behaviors that matter for this implementation: blank/whitespace `guidance_category` values (the SQL only normalizes `NULL` today), the large approved-report-id result set behind the `IN (:approvedReportIds)` path, and the failure path when `sequelize.query` rejects. I would add focused tests for those cases before merging so the current implementation and any follow-up refactor stay pinned down.

src/widgets/monitoring/reportCountByFindingCategory.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  if (!approvedReports.length) {
+    return [];
+  }
+
+  const approvedReportIds = approvedReports.map((r) => r.id);
+
+  const rows = await sequelize.query<AggregatedRow>(
+    `SELECT
+      COALESCE(c.guidance_category, :noCategory) AS guidance_category,
```
kryswisnaskas: Codex said: This only normalizes `NULL` guidance categories. The monitoring import path stores `ms.guidance` verbatim, so empty strings and whitespace-only values can still exist in `Citations.guidance_category` and will come back here as unlabeled buckets instead of falling under the fallback label. I would normalize by trimming first, converting an empty string to `NULL`, and then applying the fallback label; in SQL terms, that means `BTRIM(...)`, `NULLIF(..., empty string)`, then `COALESCE(..., :noCategory)`. The `GROUP BY` should use the same normalized expression. The test coverage should also include empty-string and whitespace-only categories, not just `NULL`.

src/widgets/monitoring/reportCountByFindingCategory.ts
thewatermethod: I don't think it's wise to use the scopeToWhere pattern; the way we did it here (two round trips) is how we commonly do it across all widgets.

src/widgets/monitoring/reportCountByFindingCategory.ts
thewatermethod: There are no whitespace or empty string guidance_categories, just null or populated. There's no reason to add work to this query that we don't need to perform, is there?

src/widgets/monitoring/reportCountByFindingCategory.ts
thewatermethod: If we desperately need to do it in one trip, I would revert to doing the query once and calculating in Javascript as I originally had it

thewatermethod: I added a constraint to the first "trip" to only return AR's with citations; this will help to address the concern about the scale risk. The consistency window is not an issue here, even in the unlikely event that the data changed in the millisecond between the two queries.

src/widgets/monitoring/reportCountByFindingCategory.test.js
thewatermethod: I added an additional pair of tests

src/widgets/monitoring/reportCountByFindingCategory.ts
thewatermethod: I'm going to resolve this without action given the actual state of the data. If you think it needs further work @kryswisnaskas just let me know

tommaroh (approved):

src/widgets/monitoring/reportCountByFindingCategory.test.js
kryswisnaskas: Thanks, the `EXISTS` filter definitely helps and makes this better. I still read the original concern as mostly about the large id handoff rather than uncited ARs. I know we use this two query pattern in other widgets, but it would be good to move away from it as this area grows. If that feels too large for this PR, would you be open to at least adding a warn log when the approved report id count gets large, and then tracking the single query refactor in a follow-up?

src/widgets/monitoring/reportCountByFindingCategory.ts
kryswisnaskas: Fair enough. If we are confident `guidance_category` cannot be blank or whitespace only in practice, I’m fine not pushing on it here. I was looking at it more as defensive coding, since the import path stores the source value verbatim.

src/widgets/monitoring/reportCountByFindingCategory.test.js
thewatermethod: @kryswisnaskas Sure. At what threshold would you like the warn log to fire? 1000 reports?

src/widgets/monitoring/reportCountByFindingCategory.ts
thewatermethod: edbc042

src/widgets/monitoring/reportCountByFindingCategory.test.js
kryswisnaskas: 1000 reports sounds good for a warning. Thanks.

thewatermethod: @kryswisnaskas I added the threshold warning, but testing it via the frontend UI in my other PR the first attempt was 1002 reports which returned a result in 32ms (very fast), so I think a larger number is warranted

kryswisnaskas: That seems fair. Based on that, 3000 seems like a better threshold.

kryswisnaskas (approved):
👍

PR #3517, opened 2026-04-16
[NO JIRA] Don't return Findings where the citations link has been deleted in the source data

hardwarehuman:
## Description of change

Different services are behaving differently, and the citations service wasn't written to notice when the link to Standards (which contain the citation text & etc) has been deleted at the source. However, other backend services (and the fact tables) _do_ notice this and exclude those findings. That may mean that the citation service is allowing users to select citations that won't "exist" for other services.

## How to test

We'd like to see how this is working with the ARs that have been generating 500 errors.

## Issue(s)

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


thewatermethod (approved):

kryswisnaskas (approved):
Verified against one of the ARs that was generating the 500 error: the UI no longer shows "Network error" and simply omits the source deleted citations from the list.

kryswisnaskas:
Claude said:

This fix correctly stops the 500s, but it converts a noisy error into silence — after this merges we lose visibility into which users are affected.

Worth adding a `warn` log in `getCitationsByGrantIds` in `citations.ts` before this ships so we keep the signal:

```ts
import { auditLogger } from '../logger';

// at the end of getCitationsByGrantIds, replace the return:
const results = grantsByCitations[0] as CitationsByGrantId[];

if (results.length === 0) {
  auditLogger.warn(
    `citations.getCitationsByGrantIds - zero active citations returned for grantIds: [${grantIds.join(', ')}]. `
    + 'MonitoringFindingStandards or MonitoringStandards rows may all be source-deleted.',
  );
}

return addCitationNames(results);
```

This uses the existing `auditLogger` pattern from other services. Without it, the only way to know who's blocked after this merges is to query `ActivityReportObjectiveCitations` directly.

tommaroh (approved):

kryswisnaskas (approved):
👍

PR #3525, opened 2026-04-21
[TTAHUB-5158] Add Categories guidance categories list table

hardwarehuman:
## Description of change

Adds the `Categories` table primarily so the frontend can have an auto-updating list of potentially-applicable guidance category values.

## How to test

I'm not sure there's anything to really test yet, but it should at least create the table and it should have 57 rows for now.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5158

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


thewatermethod:
Do we want to name this "Categories"? I'm worried it's too generic

hardwarehuman:
@thewatermethod I'm definitely not committed to Categories. Three other options that occur to me are GuidanceCategories, CitationCategories, and FindingCategories. Any preferences?

thewatermethod:
> @thewatermethod I'm definitely not committed to Categories. Three other options that occur to me are GuidanceCategories, CitationCategories, and FindingCategories. Any preferences?

Nope @hardwarehuman, I think any/all of those make sense

docs/monitoring-fact-tables.md
```diff
@@ -152,7 +162,7 @@ A `DeliveredReview` is considered `complete` when none of its linked Findings ar
 
 ## Source Code
 
-- **Models**: `src/models/deliveredReview.js`, `src/models/citation.js`, `src/models/deliveredReviewCitation.js`, `src/models/grantDeliveredReview.js`, `src/models/grantCitation.js`
+- **Models**: `src/models/deliveredReview.js`, `src/models/citation.js`, `src/models/findingCategory.js`, `src/models/deliveredReviewCitation.js`, `src/models/grantDeliveredReview.js`, `src/models/grantCitation.js`
 - **Update script**: `src/tools/updateMonitoringFactTables.ts`
 - **CLI wrapper**: `src/tools/updateMonitoringFactTablesCLI.ts`
-- **Migration**: `src/migrations/20260219034204-create-monitoring-fact-tables.js`
+- **Migrations**: `src/migrations/20260219034204-create-monitoring-fact-tables.js`, `src/migrations/20260421000000-create_categories_table.js`
```
kryswisnaskas:
```suggestion
- **Migrations**: `src/migrations/20260219034204-create-monitoring-fact-tables.js`, `src/migrations/20260421000000-create_finding_categories_table.js`
```

kryswisnaskas:
Codex said:
---------------------------------------------------------
Medium: blank or whitespace-only guidance values will become real categories and can be linked to citations. `MonitoringStandard.guidance` is nullable free text in [monitoringStandard.js](/tmp/pr3525-review/src/models/monitoringStandard.js:56), and the upsert only filters `guidance_category IS NOT NULL` in [updateMonitoringFactTables.ts](/tmp/pr3525-review/src/tools/updateMonitoringFactTables.ts:416). That means `''` or `'   '` produces a bogus `FindingCategories.name`, which is exactly the kind of empty-input edge case the frontend will render badly. Fix: normalize at the SQL boundary with something like `NULLIF(BTRIM(ms.guidance), '')`, and use the normalized value consistently for both insert and join. Add tests for `null`, empty string, and whitespace-only guidance.

kryswisnaskas:
Trying out an additional review pass using AI (Codex/Claude) to help catch edge cases and things that are easy to miss.

Goal here is to speed up reviews and catch things earlier. Open to feedback if this becomes noisy or not useful.

src/tools/updateMonitoringFactTables.ts
```diff
@@ -408,6 +408,29 @@ const updateMonitoringFactTables = async () => {
...
+    WHERE guidance_category IS NOT NULL
+    ON CONFLICT (name)
+    DO UPDATE SET
+      "updatedAt" = NOW(),
+      "deletedAt" = NULL
+    WHERE
+      "FindingCategories"."deletedAt" IS NOT NULL
+    ;
+
+    -- Categories deleted record marking
```
thewatermethod, line 425: Is there a case for not deleting them on our end? If we'll expect that data to be present on past reports, etc

thewatermethod (approved):

src/tools/updateMonitoringFactTables.ts
hardwarehuman: There probably is at least a case for it, but I want to be clear that the logic would only delete categories that IT-AMS had removed even from all historical Monitoring reviews. We'd want it to show on the historic AR to maintain FOIA accuracy, but when it came to filter values or charts (which I would expect to be the use of this table) I don't think we'd want to see the guidance categories that IT-AMS had expunged from all past and present reviews. It's probably worth asking our OHS counterparts to be sure, though.

kryswisnaskas (approved):

PR #3526, opened 2026-04-21
[TTAHUB-5217]  Handle various date formats for session start and end date sort

AdamAdHocTeam:
## Description of change

The issue here is there are some different date formats in the db. When we attempt to sort by start or end date we get the error. I will create a new ticket to investigate why we have different date formats but for now this fixes their issue.

https://jira.acf.gov/browse/TTAHUB-5220 - addresses the larger issue of how these dates are getting through.

## How to test

- Review the code
- Make sure you can sort start and end date on the TR dashboard.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5217

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy

### DIFF
--- a/src/services/sessionReports.ts
+++ b/src/services/sessionReports.ts
@@ -360,28 +360,30 @@ export async function getPossibleSessionParticipants(
 }
 
 /**
- * Get training reports (sessions) with pagination, sorting, and filtering
- * @param params Query parameters including pagination, sorting, filtering, and format
- * @returns JSON object with count and rows
- */
-/**
  * Builds a SQL CASE expression that casts a JSONB date string field to a proper date,
  * handling the inconsistent formats found in production data:
- *   YYYY-MM-DD  (ISO — direct cast)
- *   M/D/YY      (short year — TO_DATE MM/DD/YY)
- *   MM/DD/YYYY  (US standard — TO_DATE MM/DD/YYYY, the dominant format and ELSE default)
+ *   YYYY-MM-DD  (ISO — direct cast, strict match)
+ *   M/D/YY or MM/DD/YY  (short year — TO_DATE MM/DD/YY)
+ *   M/D/YYYY or MM/DD/YYYY  (US standard — TO_DATE MM/DD/YYYY)
  *   null / ''   (guarded by NULLIF → NULL)
+ *   any other value (unrecognized format → NULL to prevent sort errors)
  */
 function sessionReportDateSort(field: string): string {
   const col = `"SessionReportPilot".data->>'${field}'`;
   return `CASE
     WHEN NULLIF(${col}, '') IS NULL THEN NULL
-    WHEN ${col} ~ '^\\d{4}-' THEN (${col})::date
-    WHEN ${col} ~ '/\\d{2}$' THEN TO_DATE(${col}, 'MM/DD/YY')
-    ELSE TO_DATE(${col}, 'MM/DD/YYYY')
+    WHEN ${col} ~ '^\\d{4}-\\d{2}-\\d{2}$' THEN (${col})::date
+    WHEN ${col} ~ '^\\d{1,2}/\\d{1,2}/\\d{2}$' THEN TO_DATE(${col}, 'MM/DD/YY')
+    WHEN ${col} ~ '^\\d{1,2}/\\d{1,2}/\\d{4}$' THEN TO_DATE(${col}, 'MM/DD/YYYY')
+    ELSE NULL
   END`;
 }
 
+/**
+ * Get training reports (sessions) with pagination, sorting, and filtering
+ * @param params Query parameters including pagination, sorting, filtering, and format
+ * @returns JSON object with count and rows
+ */
 export async function getSessionReports(
   params: GetSessionReportsParams,
 ): Promise<GetSessionReportsResponse> {
--- a/src/services/sessionReports.ts
+++ b/src/services/sessionReports.ts
@@ -364,6 +364,24 @@ export async function getPossibleSessionParticipants(
  * @param params Query parameters including pagination, sorting, filtering, and format
  * @returns JSON object with count and rows
  */
+/**
+ * Builds a SQL CASE expression that casts a JSONB date string field to a proper date,
+ * handling the inconsistent formats found in production data:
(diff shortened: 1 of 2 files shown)


thewatermethod (approved):

src/services/sessionReports.ts
```diff
@@ -364,6 +364,24 @@ export async function getPossibleSessionParticipants(
...
  */
+/**
+ * Builds a SQL CASE expression that casts a JSONB date string field to a proper date,
+ * handling the inconsistent formats found in production data:
+ *   YYYY-MM-DD  (ISO — direct cast)
+ *   M/D/YY      (short year — TO_DATE MM/DD/YY)
+ *   MM/DD/YYYY  (US standard — TO_DATE MM/DD/YYYY, the dominant format and ELSE default)
+ *   null / ''   (guarded by NULLIF → NULL)
+ */
+function sessionReportDateSort(field: string): string {
```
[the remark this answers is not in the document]
thewatermethod: Valid to fix this

tommaroh (approved):

src/services/sessionReports.ts
```diff
@@ -394,8 +412,8 @@ export async function getSessionReports(
   const sortMap: SessionReportSortSortMap = {
     id: ['id'],
     sessionName: [sequelize.literal('("SessionReportPilot".data->>\'sessionName\')::text')],
-    startDate: [sequelize.literal('TO_DATE(NULLIF("SessionReportPilot".data->>\'startDate\', \'\'), \'MM/DD/YYYY\')')],
-    endDate: [sequelize.literal('TO_DATE(NULLIF("SessionReportPilot".data->>\'endDate\', \'\'), \'MM/DD/YYYY\')')],
+    startDate: [sequelize.literal(sessionReportDateSort('startDate'))],
+    endDate: [sequelize.literal(sessionReportDateSort('endDate'))],
```
kryswisnaskas, line 418: Codex said:
----------------------------------------------------
This change applies the new parsing logic to `endDate` too, but the added regression coverage only exercises `Session_start_date`. That means `endDate` is still unproven even though it shares the same risk surface.

Can we add at least one mixed-format `Session_end_date` regression test, ideally alongside a malformed-value case so both branches stay aligned?

src/services/sessionReports.ts
```diff
@@ -364,6 +364,24 @@ export async function getPossibleSessionParticipants(
...
+ *   M/D/YY      (short year — TO_DATE MM/DD/YY)
+ *   MM/DD/YYYY  (US standard — TO_DATE MM/DD/YYYY, the dominant format and ELSE default)
+ *   null / ''   (guarded by NULLIF → NULL)
+ */
+function sessionReportDateSort(field: string): string {
+  const col = `"SessionReportPilot".data->>'${field}'`;
+  return `CASE
+    WHEN NULLIF(${col}, '') IS NULL THEN NULL
+    WHEN ${col} ~ '^\\d{4}-' THEN (${col})::date
+    WHEN ${col} ~ '/\\d{2}$' THEN TO_DATE(${col}, 'MM/DD/YY')
```
kryswisnaskas: Codex said:
----------------------------------------------------
I am still worried about the `TO_DATE` branches here. PostgreSQL `to_date` is permissive and can normalize bad inputs instead of rejecting them, so this can return a successfully sorted response with the wrong chronological order for malformed values rather than failing loudly.

Relying on `TO_DATE(..., MM/DD/YY)` and `TO_DATE(..., MM/DD/YYYY)` means bad source data can silently become a different date. I would prefer a stricter parse path or a round-trip check before using the parsed value for sorting.

src/services/sessionReports.ts
```diff
@@ -364,6 +364,24 @@ export async function getPossibleSessionParticipants(
...
+ *   YYYY-MM-DD  (ISO — direct cast)
+ *   M/D/YY      (short year — TO_DATE MM/DD/YY)
+ *   MM/DD/YYYY  (US standard — TO_DATE MM/DD/YYYY, the dominant format and ELSE default)
+ *   null / ''   (guarded by NULLIF → NULL)
+ */
+function sessionReportDateSort(field: string): string {
+  const col = `"SessionReportPilot".data->>'${field}'`;
+  return `CASE
+    WHEN NULLIF(${col}, '') IS NULL THEN NULL
+    WHEN ${col} ~ '^\\d{4}-' THEN (${col})::date
```
kryswisnaskas: Codex said:
----------------------------------------------------
This fixes the specific TO_DATE(..., MM/DD/YYYY) crash for ISO-formatted values, but this branch is still broader than the actual safe ISO shape. `^\d{4}-` will also match values like `2026-13-01`, `2026-02-31`, or `2026-03-13foo`, and the raw `::date` cast can still fail and take down the whole sort query.

I think this needs a stricter match such as `^\d{4}-\d{2}-\d{2}$` plus validation, or a fallback to `NULL` for anything that does not round-trip cleanly.

thewatermethod:
This issue had been identified in testing an unrelated widget last week and I created a PR to address it: https://github.com/HHS/Head-Start-TTADP/pull/3529, I just didn't get it out fast enough. We may or may not need both

src/services/sessionReports.ts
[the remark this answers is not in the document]
AdamAdHocTeam: @copilot apply changes based on [this feedback](https://github.com/HHS/Head-Start-TTADP/pull/3526#discussion_r3120216757)

kryswisnaskas (approved):
Looks better. 👍 Codex still flagged one more minor issue. I think it's good to be aware of it and potentially address it in a follow-up PR, so that we are not blocking this one.

Codex said:
--------------------------------
src/services/sessionReports.ts in the PR still does WHEN ${col} ~ '^\\d{4}-\\d{2}-\\d{2}$' THEN (${col})::date. That is stricter, but it still throws on exact-shape invalid ISO values like 2026-13-01 or 2026-02-31. So the “ISO branch can still 500 the query” finding is reduced, not eliminated.

PR #3527, opened 2026-04-21
[TTAHUB-5221] Fix for users seeing TR's from other regions

AdamAdHocTeam:
## Description of change

Users with one region would see TR reports from another region on the TR regional dashboard. The issue here is that the generic component we use to load the selected dashboard expects the prop name to be filtersToApply but we passed filters.

## How to test

- Review the code
- Make sure users only see TR's from their region.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5221

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


thewatermethod (approved):
I have a follow on backend PR for this

frontend/src/pages/RegionalDashboard/components/TrainingReportDashboard.js
```diff
@@ -92,9 +92,9 @@ export default function TrainingReportDashboard({ filters }) {
 }
 
 TrainingReportDashboard.defaultProps = {
-  filters: [],
+  filtersToApply: [],
```
kryswisnaskas, line 95: Codex said:
------------------------------------------------------------
This still fails open if the filter prop is ever omitted again. `TrainingReportDashboard` now reads `filtersToApply`, but the default here is still `[]`, and the new test explicitly locks that in. That is the same failure mode that caused the bug: a prop mismatch silently turns into an unscoped request and cross-region data exposure.

Concrete fix: make `filtersToApply` required, remove the empty default, and refuse to fetch if the prop is missing. If an empty filter set is ever legitimate, make the parent pass `[]` explicitly so a wiring bug does not look like a valid state.

frontend/src/pages/RegionalDashboard/components/__tests__/TrainingReportDashboard.js
```diff
@@ -53,4 +60,31 @@ describe('Training report Dashboard page', () => {
     expect(screen.getByText('Hours of training by National Center')).toBeInTheDocument();
     expect(screen.getByText('Number of TR sessions by topic')).toBeInTheDocument();
   });
+
+  it('passes filtersToApply to the session reports table fetch', async () => {
```
kryswisnaskas, line 64: Codex said:
------------------------------------------------------------
The added regression tests still do not exercise the broken runtime path. The new unit test renders the child directly and only checks that one matching `getSessionReportsTable` call happened. Meanwhile the existing route-level test for training reports only checks the heading and even mocks activity-report endpoints instead of training-report ones. That means this suite would still pass if `Dashboard` stopped wiring `filtersToApply`, if `expandFilters` broke, or if the page made an initial unfiltered request before a filtered one.

Concrete fix: add an integration test at `RegionalDashboard`/`Dashboard` for a one-region user and assert every training-report request URL includes `region.in[]=1`, with no unfiltered request.

thewatermethod:
https://github.com/HHS/Head-Start-TTADP/pull/3528 in conjunction with this PR should fully address this issue, including the _unscoped_ case that codex mentions above

PR #3542, opened 2026-04-27
[TTAHUB-5184] Use fact tables & new live value views for Monitoring Goal creation

hardwarehuman:
## Description of change

This refactors monitoring goal creation to use the Monitoring fact tables, plus new live value views that contain the logic for figuring out the last time anyone delivered TTA or closed a goal for `Citations` or `DeliveredReviews records`. The views and the new scopes they enable seem likely to be very useful, and the documentation should explain how to use them. There's also a migration in there that references the view definition that should cover most but not all future cases where the views need to be created to support testing or local dev.

## How to test

If running correctly on prod data, it will actually create three new Goals due to fixing a couple of old edge cases. Probably the bigger question, though, is whether the 
```
const citations = await Citation.scope('withLiveValues').findAll({ where: { active: true } });
citations[0].liveValues.last_closed_goal;
citations[0].liveValues.last_tta;
```
 usage approach will work for folks, and if it's all sufficiently self-explanatory, including the comments in the code describing the approach and especially how to think about when a new migration might be needed.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5184

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


thewatermethod (approved):
Looks great. Two comments, one from me and one from the robot.

1) The duplicated migration/app code could be moved to a shared function (?) if we don't want to rely on obeisance to the comment, although it's probably not code that will change super often.

2) **Problem**: The `withLiveValues` scope uses four correlated scalar subqueries — one per virtual field — each hitting the view independently for every row in the result set. If the view's CTEs are not materialized by the planner, each row triggers the full CTE chain four times. **Impact**: For typical usage (small result sets with `WHERE` clauses), this is unlikely to be noticeable. But a `findAll` on all active Citations could perform poorly. This is worth monitoring but not blocking. **Suggested Fix**: Consider using a single `LEFT JOIN` to the view instead of four subqueries. This would require a raw attributes approach or a Sequelize literal join, but collapses four view hits into one per query.

hardwarehuman:
@thewatermethod I think those were both good points, and I made changes. Now references aren't _quite_ as flat and clean, but the join should scale well to any size dataset. Also the migration now pulls its implementation out of the method in  `updateMonitoringFactTables.ts`. There's still circumstances which could need a new migration file, but fewer now. I'm not 100% sure it's an improvement because maybe the explanation of when it's necessary is more complicated than just telling people to just always include a new migration.

thewatermethod (approved):

AdamAdHocTeam (approved):

PR #3569, opened 2026-05-01
[TTAHUB-5146, TTAHUB-5149, TTAHUB-5148] Goals table

kryswisnaskas:
## Description of change
Implements the Goals table

## How to test
Navigate to the "Goals Dashboard"
Verify the UI matches the Design
Verify the goals can be paginated and sorted

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5146
* https://jira.acf.gov/browse/TTAHUB-5149
* https://jira.acf.gov/browse/TTAHUB-5148

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


frontend/src/hooks/useCheckboxSelection.js
```diff
@@ -32,6 +39,39 @@ export default function useCheckboxSelection({ items, allItemIds = [], getItemId
...
+
+          return selections;
+        },
+        {},
+      );
+
+      return Object.keys(nextSelections).length === Object.keys(previousSelections).length
+        ? previousSelections
+        : nextSelections;
+    });
```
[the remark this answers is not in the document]
kryswisnaskas: Addressed.

frontend/src/pages/GoalDashboard/GoalDashboardGoalCards.js
```diff
@@ -0,0 +1,163 @@
...
+              goal={goal}
+              recipientId={goalRecipientId}
+              regionId={goalRegionId}
+              recipientName={goal.grant?.recipient?.name || ''}
+              showRecipientColumn
+              deletableStatuses={[GOAL_STATUS.NOT_STARTED]}
+              onGoalDeleted={onGoalDeleted}
+              handleGoalCheckboxSelect={handleCheckboxSelect}
+              isChecked={isChecked(goal.id)}
+              allSelectedGoalIds={selectedGoalCheckboxes}
```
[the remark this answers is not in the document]
kryswisnaskas: Addressed.

frontend/src/hooks/useCheckboxSelection.js
```diff
@@ -73,6 +113,14 @@ export default function useCheckboxSelection({ items, allItemIds = [], getItemId
     setSelectedCheckboxes(allIdCheckboxes);
   };
 
+  const selectIds = (ids) => {
+    const selectedIdsToSet = ids.reduce(
+      (obj, id) => ({ ...obj, [String(id)]: true }),
+      {},
+    );
+    setSelectedCheckboxes(selectedIdsToSet);
+  };
```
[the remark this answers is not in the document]
kryswisnaskas: Addressed.

src/services/dashboards/goal.js
```diff
@@ -18,6 +29,16 @@ const DISPLAY_STATUS = {
...
 const MIN_STANDARD_GOAL_CREATED_AT = '2025-09-09';
+const DEFAULT_GOAL_DASHBOARD_PER_PAGE = 10;
+const MAX_GOAL_DASHBOARD_PER_PAGE = 50;
+const GOAL_DASHBOARD_SORT_FIELDS = ['createdOn', 'goalStatus', 'goalCategory'];
+
+const isoDateToDisplayDate = (date) => {
+  const [year, month, day] = date.split('-');
+  return `${month}/${day}/${year}`;
+};
+
```
thewatermethod, line 40: why not use moment as in the rest of the codebase

frontend/src/hooks/useFetch.js
thewatermethod, line 1: Can you explain what we're doing here? It looks like we could maybe replace some of this logic with an AbortController but without explanation it's difficult to say. It also could be the case that there is something in the dependency array causing us to fetch multiple times, if that is the case this is trying to solve

frontend/src/pages/GoalDashboard/GoalDashboardGoalCards.js
thewatermethod, line 1: I would think we'd want the loading spinner when changing page number, per page, etc

frontend/src/pages/GoalDashboard/GoalDashboardGoalsSection.js
```diff
@@ -0,0 +1,237 @@
...
+    const nextAllGoalIds = nextDashboardGoals?.allGoalIds || [];
+    setDashboardGoals((previousDashboardGoals) => ({
+      ...previousDashboardGoals,
+      allGoalIds: nextAllGoalIds,
+    }));
+    return nextAllGoalIds;
+  }, [goalsQuery, setDashboardGoals]);
+
+  return (
+    <Container
```
thewatermethod: Did we check to see if we could use the existing "WidgetContainer"

frontend/src/pages/GoalDashboard/GoalDashboardGoalsSection.js
```diff
@@ -0,0 +1,237 @@
...
+    };
+  }
+
+  return {
+    sortBy: value.slice(0, directionSeparatorIndex),
+    direction: value.slice(directionSeparatorIndex + 1),
+  };
+};
+
+function GoalDashboardGoalsSection({ dataStartDateDisplay }) {
```
thewatermethod, line 95: we have hooks that handle sorting and pagination, is it possible to use those?

thewatermethod:
Found via AI:

**HIGH: Deleting the last item on the final page leaves the dashboard on an invalid page**

**File**: `frontend/src/pages/GoalDashboard/GoalDashboardGoalsSection.js:113`

**Problem**: `handleGoalDeleted` only mutates the cached `count`, `goalRows`, and `allGoalIds`. It does **not** clamp `sortConfig.activePage` / `sortConfig.offset` when the delete removes the only row on the current last page.

**Impact**: If a user is on the last page and deletes the final remaining goal there, the UI can stay on a now-nonexistent page. `PaginationCard` will render an invalid range like `0-10 of 10`, and the dashboard can show an empty state even though earlier pages still have goals.

**Suggested Fix**:
```js
useEffect(() => {
  if (!dashboardGoals) {
    return;
  }

  const maxPage = Math.max(Math.ceil((dashboardGoals.count || 0) / perPage), 1);

  if (sortConfig.activePage > maxPage) {
    setSortConfig((previousConfig) => ({
      ...previousConfig,
      activePage: maxPage,
      offset: (maxPage - 1) * perPage,
    }));
  }
}, [dashboardGoals, perPage, sortConfig.activePage]);
```

This keeps pagination state in sync after local deletions and naturally triggers a refetch for the new valid page when needed.

**Additional Context**: The current tests only cover deleting a goal from the first page (`frontend/src/pages/GoalDashboard/__tests__/index.js:156`), so this regression is not currently caught.

## Testing Recommendations

1. Add a UI test that navigates to page 2 of a 2-page result set, deletes the only goal on that page, and verifies the component returns to page 1 with a valid count header.
2. Verify the count header never renders an invalid `0-x of y` range after local dashboard deletions.

src/services/dashboards/goal.js
kryswisnaskas: `moment` would work here too, but this is just reformatting a fixed ISO date constant. A small string transform felt simpler, and it avoids adding another moment usage when the goal is to move away from it over time.

frontend/src/hooks/useFetch.js
kryswisnaskas: Added a comment.

A local `AbortController` would have worked here too. `useFetch` was kept because this screen was already using it. Rather than adding a separate fetch pattern just for this screen, the stale response problem was fixed in the shared hook.

frontend/src/pages/GoalDashboard/GoalDashboardGoalCards.js
kryswisnaskas: This should now be covered through `WidgetContainer`.

frontend/src/pages/GoalDashboard/GoalDashboardGoalsSection.js
kryswisnaskas: Refactored to use `WidgetContainer`

frontend/src/pages/GoalDashboard/GoalDashboardGoalsSection.js
kryswisnaskas: Took a look at the existing hooks, and they seem aimed at client side sorting/paging over already loaded data. This screen is server driven, so it needs to track query state and refetch when sort or page changes.

frontend/src/hooks/useFetch.js
thewatermethod: Could we not update useFetch to use an AbortController rather than our own implementation? If that's not possible, we should definitely add some comments.

kryswisnaskas: `AbortController` would be a broader shared hook change. `useFetch` and its callers are not currently standardized around passing a `signal`, so this PR kept the existing contract and added comments to explain the logic.

kryswisnaskas:
> Found via AI:
> 
> **HIGH: Deleting the last item on the final page leaves the dashboard on an invalid page**
> 
> **File**: `frontend/src/pages/GoalDashboard/GoalDashboardGoalsSection.js:113`
> 
> **Problem**: `handleGoalDeleted` only mutates the cached `count`, `goalRows`, and `allGoalIds`. It does **not** clamp `sortConfig.activePage` / `sortConfig.offset` when the delete removes the only row on the current last page.
> 
> **Impact**: If a user is on the last page and deletes the final remaining goal there, the UI can stay on a now-nonexistent page. `PaginationCard` will render an invalid range like `0-10 of 10`, and the dashboard can show an empty state even though earlier pages still have goals.
> 
> **Suggested Fix**:
> 
> ```js
> useEffect(() => {
>   if (!dashboardGoals) {
>     return;
>   }
> 
>   const maxPage = Math.max(Math.ceil((dashboardGoals.count || 0) / perPage), 1);
> 
>   if (sortConfig.activePage > maxPage) {
>     setSortConfig((previousConfig) => ({
>       ...previousConfig,
>       activePage: maxPage,
>       offset: (maxPage - 1) * perPage,
>     }));
>   }
> }, [dashboardGoals, perPage, sortConfig.activePage]);
> ```
> 
> This keeps pagination state in sync after local deletions and naturally triggers a refetch for the new valid page when needed.
> 
> **Additional Context**: The current tests only cover deleting a goal from the first page (`frontend/src/pages/GoalDashboard/__tests__/index.js:156`), so this regression is not currently caught.
> 
> ## Testing Recommendations
> 1. Add a UI test that navigates to page 2 of a 2-page result set, deletes the only goal on that page, and verifies the component returns to page 1 with a valid count header.
> 2. Verify the count header never renders an invalid `0-x of y` range after local dashboard deletions.

Addressed.

AdamAdHocTeam (commented):
Overall what you have looks good and seems to work fine.

Im not sure about the migrations in #1 as it seems outside the scope of this ticket.

See some below AI findings: 

 1. Migration calls updateMonitoringFactTables() outside its transaction — HIGH

 File: src/migrations/20260429220319-expand_monitoring_fact_table_columns.js:83

     });  // transaction closes here — all schema changes committed

     await updateMonitoringFactTables();  // runs outside the transaction

 The seven addColumn calls finish inside a transaction. If updateMonitoringFactTables() then fails (network blip, query
 error, etc.), the schema changes are permanently committed but the fact tables are stale/unpopulated. There is no rollback
  path. The only way to recover is a manual down migration followed by a re-run, which loses whatever data was previously
 in those columns.

 Fix: Move await updateMonitoringFactTables() inside the transaction body so that any failure rolls back both the schema
 changes and the data population atomically.

 ---
 2. Deleting a goal from a non-last page leaves the page underfilled — MEDIUM

 File: frontend/src/pages/GoalDashboard/GoalDashboardGoalsSection.js:110–125

 handleGoalDeleted removes goals from local state (setDashboardGoals) but never triggers a refetch. When a user deletes a
 goal from page 1 of N pages, the page now shows perPage - 1 items instead of perPage. Goals from the next page never
 backfill because goalsQuery hasn't changed (sort/offset/perPage are all unchanged), so useFetch has no reason to refetch.

 The auto-reset useEffect (line 141–151) only fires when activePage > maxPage, which won't happen if the deletion doesn't
 collapse the page count.

 Fix: After a deletion that doesn't collapse the page count, either trigger a refetch (e.g., bump a refreshKey in the
 useFetch deps) or accept that deleting one item will leave a gap.

 ---
 3. useFetch cleanup emits a spurious setIsAppLoading(false) before a new fetch starts — LOW

 File: frontend/src/hooks/useFetch.js:61–65

 return () => {
   cancelled = true;
   if (useAppLoading && setIsAppLoading && requestId === requestIdRef.current) {
     setIsAppLoading(false);   // fires here…
   }
 };
 // …then the new effect immediately sets setIsAppLoading(true) again

 At cleanup time requestIdRef.current still equals the OLD requestId (the new effect hasn't run yet), so the condition is
 true and setIsAppLoading(false) fires. The new effect then immediately calls setIsAppLoading(true). This causes a
 one-render flicker on the global loading indicator.

 None of the new code in this PR passes useAppLoading=true, so this won't be visible right now, but any future caller that
 does will get the flicker.

 Fix: In the cleanup, skip the setIsAppLoading(false) call when the cleanup is triggered by a deps change (i.e., a new
 fetch is about to start). The simplest guard is to check cancelled && requestId === requestIdRef.current only on unmount,
 not on re-run.

 ---
 Missing Test Cases

 A. Goal deletion from a non-last page (ties to bug #2)

 No test covers: delete 1 goal from page 1 when there are 2+ pages. Verify whether the page refetches or shows perPage - 1
 items. Either outcome should be explicitly asserted so the behavior is documented and doesn't silently regress.

 B. goalCategory sort path in the backend service

 File: src/services/dashboards/goal.js:228–238 / src/services/dashboards/goal.test.js

 orderDashboardGoalsBy has a dedicated branch for sortBy === 'goalCategory' that uses
 sequelize.col('goalTemplate.standard'), but neither the unit tests nor the integration test (goal.integration.test.js)
 exercise this sort option. If the goalTemplate join alias changes or the column is missing, this branch would fail at
 runtime with no test catching it.

 C. Pagination reset when goal deletion collapses page count

 No test covers: user is on page 2 of 2 (with exactly 1 goal on page 2), deletes that goal, and expects the UI to reset to
 page 1 and refetch. The useEffect at line 141–151 is intended to handle this, but there is no test that triggers
 activePage > maxPage via deletion.

 D. fetchAllGoalIds API failure in GoalDashboardGoalCards

 File: frontend/src/pages/GoalDashboard/GoalDashboardGoalCards.js:41–58

 The handleSelectAllResults function wraps the onSelectAllGoals() call in a try/catch and sets selectAllError=true on
 failure. No test verifies that the error Alert renders when onSelectAllGoals rejects, or that selectAllLoading resets to
 false in the finally block after an error.

kryswisnaskas:
@AdamAdHocTeam addressed the findings in `76a6751` with the exception of the migration which wasn't introduced in this PR.

AdamAdHocTeam (approved):

PR #3654, opened 2026-05-27
[TTAHUB-5332] Approved ARs and TR sessions by goal category graph BE

AdamAdHocTeam:
## Description of change

This is the BE for the new AR vs TR Session goal category graph on the 'TTA History' tab. 

## How to test

- Review the code make sure we haven't overlooked any spec.
- Make sure all BE tests pass.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5332

### Every PR

- [x] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


src/widgets/approvedARAndTRByGoalCategory.ts
```diff
@@ -0,0 +1,174 @@
...
+        model: db.SessionReportPilot,
+        as: 'sessionReports',
+        attributes: [],
+        required: true,
+        where: {
+          data: { status: TRAINING_REPORT_STATUSES.COMPLETE },
+          eventId: events.map(({ id }) => id),
+        },
+      },
+    ],
```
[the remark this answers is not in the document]
AdamAdHocTeam: We are showing goals and filtering on goal created date.

thewatermethod:
This duplicates existing logic in `/src/widgets/standardGoalsList.js` and `src/widgets/trStandardGoalList.ts` - is it possible we could we just reuse those functions instead of adding a new implementation?

src/widgets/approvedARAndTRByGoalCategory.ts
```diff
@@ -0,0 +1,174 @@
...
+        attributes: [],
+        required: true,
+        where: {
+          createdAt: { [Op.gte]: GOAL_CUTOFF_DATE },
+          prestandard: false,
+        },
+      },
+      {
+        through: { attributes: [] },
+        model: db.SessionReportPilot,
```
thewatermethod, line 107: I think you could just include events and make them required here. (inner join) and then you wouldn't need two queries

thewatermethod: Actually probably messes up the group by

src/widgets/approvedARAndTRByGoalCategory.ts
```diff
@@ -0,0 +1,174 @@
...
+ * calculatedStatus = APPROVED.
+ * Scoped by scopes.activityReport.
+ */
+async function getApprovedARCountsByCategory(
+  scopes: IScopes,
+): Promise<ICategoryCount[]> {
+  return db.ActivityReport.findAll({
+    attributes: [
+      [sequelize.col('activityReportGoals.goal.goalTemplate.standard'), 'standard'],
+      [sequelize.fn('COUNT', sequelize.literal('DISTINCT "ActivityReport"."id"')), 'count'],
```
thewatermethod: Minor: this is returned as a string since it's a BIGINT. It doesn't functionally cause any issues because at the end of the day you are parsing it as an int in JavaScript, but it makes the type wrong

src/widgets/approvedARAndTRByGoalCategory.ts
```diff
@@ -0,0 +1,174 @@
...
+ * Goal.createdAt >= 2025-09-01, and the session data.status = COMPLETE.
+ * This query does not filter on SessionReportPilotGoalTemplate.createdAt.
+ * Scoped by scopes.trainingReport.
+ */
+async function getApprovedTRCountsByCategory(
+  scopes: IScopes,
+): Promise<ICategoryCount[]> {
+  const events = await db.EventReportPilot.findAll({
+    attributes: ['id'],
+    where: { [Op.and]: [scopes.trainingReport] },
```
kryswisnaskas: Codex said:
The TR half of this widget is not recipient-scoped for the TTA History page. `TTAHistory` always adds `recipientId` to the page filters, but `trainingReportsFiltersToScopes` has no `recipientId` handler, and this query only applies `scopes.trainingReport` before counting sessions. On a recipient record, that means the graph can pull TR sessions from other recipients in the same region. Fix by adding a training-report recipient/grant scope and joining `SessionReportPilot -> grants -> recipient` (or equivalent) in the counted query instead of relying on event-level scopes alone.

src/widgets/approvedARAndTRByGoalCategory.ts
```diff
@@ -0,0 +1,174 @@
...
+    where: {
+      creationMethod: CREATION_METHOD.CURATED,
+      standard: { [Op.not]: MONITORING_STANDARD, [Op.ne]: null },
+    },
+    include: [
+      {
+        // A GoalTemplate only qualifies when at least one non-prestandard Goal
+        // using it was created on or after the cutoff date — consistent with
+        // the AR side.
+        model: db.Goal,
```
kryswisnaskas: Codex said:
The TR cutoff logic can leak categories from out-of-scope data. This `GoalTemplate -> goals` join treats a template as eligible if it has any non-prestandard post-cutoff `Goal`, but that join is not tied back to the same recipient/region/grant population as the sessions being counted. A post-2025 goal for some other recipient can therefore make an in-scope session category appear even when the current filtered population only has pre-cutoff goals. Fix by deriving eligible template IDs from the same scoped grant/recipient set, or by joining the goal side through the same scoped grants before grouping.

src/widgets/approvedARAndTRByGoalCategory.test.js
```diff
@@ -0,0 +1,707 @@
...
+    // Shared data
+    await Grant.destroy({ where: { id: [grant.id, grant2.id] }, individualHooks: true });
+    await Recipient.destroy({ where: { id: recipient.id } });
+    await User.destroy({ where: { id: user.id } });
+  });
+
+  // ─── Structural tests ──────────────────────────────────────────────────────
+
+  it('returns activityReportCount and sessionReportCount for each category', async () => {
+    const scopes = await filtersToScopes({ 'region.in': [String(grant.regionId)] });
```
kryswisnaskas: Codex said:
The new tests do not cover either of those failure paths. These cases only exercise `region.in` and a no-match region, so they would not catch recipient leakage or cutoff leakage from another recipient/region. Add regressions with a second recipient in the same region and with an out-of-scope post-cutoff goal sharing the same template.

src/widgets/approvedARAndTRByGoalCategory.ts
kryswisnaskas: To reproduce this use:
- load the obfuscated prod data
- issue a request to `http://localhost:3000/api/widgets/approvedARAndTRByGoalCategory?region.in[]=1&recipientId.ctn[]=530`
- observe that there are categories returned with "`sessionReportCount`" > 0, while that recipient doesn't have any completed sessions.

AdamAdHocTeam:
> This duplicates existing logic in `/src/widgets/standardGoalsList.js` and `src/widgets/trStandardGoalList.ts` - is it possible we could we just reuse those functions instead of adding a new implementation?

I think the cutoff date makes them dif.

src/widgets/approvedARAndTRByGoalCategory.ts
kryswisnaskas: To repro this finding:

- use the obfuscated prod DB
- issue a call to `http://localhost:3000/api/widgets/approvedARAndTRByGoalCategory?region.in[]=10&standard.in[]=Family%20Support`
- in the response, note that Family Support has `sessionReportCount = 3`
- verify separately that Region 10 has 0 non-prestandard curated Family Support goals

src/widgets/approvedARAndTRByGoalCategory.ts
AdamAdHocTeam: updated.

src/widgets/approvedARAndTRByGoalCategory.ts
AdamAdHocTeam: updated.

src/widgets/approvedARAndTRByGoalCategory.test.js
AdamAdHocTeam: updated.

src/widgets/approvedARAndTRByGoalCategory.ts
AdamAdHocTeam: updated.

thewatermethod (approved):

src/widgets/approvedARAndTRByGoalCategory.ts
```diff
@@ -0,0 +1,203 @@
+import { REPORT_STATUSES, TRAINING_REPORT_STATUSES } from '@ttahub/common';
+import { Op } from 'sequelize';
+import { CREATION_METHOD } from '../constants';
+import db, { sequelize } from '../models';
+import type { IScopes } from './types';
+
+const GOAL_CUTOFF_DATE = new Date('2025-09-01');
```
kryswisnaskas, line 7: Codex said: This still defines the cutoff using Goal.createdAt >= 2025-09-01. That does not match the existing TTA History widget logic, which uses report/event dates in src/widgets/standardGoalsList.js:16 and src/widgets/trStandardGoalList.ts:17. That means this new graph can disagree with the rest of the tab for the same filters. For example, an October 2025 approved AR tied to a goal created in August is counted by standardGoalsList but excluded here. Suggested fix: use the same cutoff rule as the existing widgets, or explicitly document and isolate this widget if product intentionally wants different semantics.

AdamAdHocTeam: Updated to be pre standard (as before) and activity start date GTE than 09/01.

kryswisnaskas (approved):
Looks good!

PR #3661, opened 2026-05-29
[TTAHUB-5539] Specification for actionable notification backend

thewatermethod:
## Description of change

Design backend for actionable notifications. Frontend tickets are briefly listed but work will be broken down separately.

## How to test

Review implementation proposal. Identify additional work that may need to be completed. Ask questions where specification is unclear.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5339

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


specs/actionable-notifications/index.md
```diff
@@ -0,0 +1,163 @@
...
+- Avoid the use of hooks to create notifications. Doing it inline (for example, right when changes are requested to an activity report rather than a hook that requests to a status change) makes it traceable and reduces the need to query the database again for associated metadata in a new context
+- Use Typescript throughout, strictly
+- Validate requests with Joi
+
+## Steps
+
+### Database Tables/Sequelize Models
+
+**Ticket #1: Create Notifications schema**
+Points: 5
```
AdamAdHocTeam, line 29: do we need number of pts in a read me?

specs/actionable-notifications/index.md
```diff
@@ -0,0 +1,163 @@
...
+
+- userId: FK to users, nullable
+- entityId: ID of the source entity, nullable (polymorphic; database FK enforcement is not possible unless the schema uses separate typed columns/tables)
+  Potential links are: group, communicationLog, activityReport, collabReport, trainingReport, sessionReport
+  https://sequelize.org/docs/v6/advanced-association-concepts/polymorphic-associations/
+- type: NOTIFICATION_TYPE[enum]
+- link: computed link
+- label: label for link
+- text: computed message (see notification configuration, next section)
+- isArchived: Boolean, default false
```
AdamAdHocTeam: Any need to track the day viewed or archived?

specs/actionable-notifications/index.md
thewatermethod: This is partially for breaking down the work for Corinne; as I noted in the PR description, I would delete this spec file and turn it into a readme as part of the work itself

specs/actionable-notifications/index.md
thewatermethod: Can't see why. Do you have a reason?

specs/actionable-notifications/index.md
```diff
@@ -0,0 +1,163 @@
...
+        where: {
+            id: notificationId,
+        }
+    }
+);
+
+```
+
+```deleteNotification(notificationId)```
+Deletes a notification with the given ID
```
AdamAdHocTeam: Not sure if we need permission check.

AdamAdHocTeam (approved):
I think this looks good. Should we mention something about cleaning up no longer valid notifications. For example two approvers added to report and submitted. Then the report is moved back to needs action?

Overall good start (minor comments)

specs/actionable-notifications/index.md
AdamAdHocTeam: Someone says they never saw the notification?

specs/actionable-notifications/index.md
thewatermethod: Permissions are usually checked in handlers, these are services; as written, the delete path is only called via the scheduled job

specs/actionable-notifications/index.md
thewatermethod: Cool, I added viewedAt and archivedAt in place of the booleans

thewatermethod:
> I think this looks good. Should we mention something about cleaning up no longer valid notifications. For example two approvers added to report and submitted. Then the report is moved back to needs action? 

@AdamAdHocTeam good call, I updated to add a new service type

specs/actionable-notifications/index.md
```diff
@@ -0,0 +1,195 @@
...
+**Ticket #1: [Create Notifications schema](https://jira.acf.gov/browse/TTAHUB-5383)**
+Points: 5
+
+- userId: FK to users, nullable
+- entityId: ID of the source entity, nullable (polymorphic; database FK enforcement is not possible unless the schema uses separate typed columns/tables)
+  Potential links are: group, communicationLog, activityReport, collabReport, trainingReport, sessionReport
+  https://sequelize.org/docs/v6/advanced-association-concepts/polymorphic-associations/
+- type: NOTIFICATION_TYPE[enum]
+- link: computed link
+- label: label for link
```
kryswisnaskas, line 39: Codex said: High: Global notifications still do not have a workable per-user read/archive model. The schema still says `userId` is nullable for global notifications while `archivedAt` and `viewedAt` live on the notification row itself. That means a single global notification row cannot track whether different users have viewed or archived it independently. This is the same core ownership problem from the previous version, just with timestamps instead of booleans. The fix is still to separate notification content from per-user state, or fan out one row per recipient.

specs/actionable-notifications/index.md
```diff
@@ -0,0 +1,195 @@
...
+- text: computed message (see notification configuration, next section)
```
kryswisnaskas, line 41: Codex said: Medium: The source entity model is missing a discriminator. This defines only a polymorphic `entityId` even though the source can be an activity report, collaboration report, training report, session report, and so on. Numeric ids will overlap across tables, so `entityId` alone is ambiguous for dedupe, cleanup, analytics, or backfills. Relying on notification type to imply the source entity will get brittle as the system adds more types. The fix is to add `entityType` or move to explicit typed associations.

specs/actionable-notifications/index.md
```diff
@@ -0,0 +1,195 @@
...
+
+```createGlobalNotification```
+
+```updateNotification(notificationId, updatedNotification)```
+Updates notifications, atomically (only _archivedAt_, _triggeredAt_, and _viewedAt_ will be updated, should be enforced via code, in both the service, the joi validation, and the model configuration if possible)
+
+```js
+// just an example
+await Notification.update(
+    updatedNotification,
```
kryswisnaskas: Codex said: Medium: `updateNotification` is specified as an id-only mutation even though ownership is only described at the handler layer. The service shape here updates by `notificationId`, while later the spec says users should only be able to update and view their own notifications. That makes the service easy to misuse later from another callsite and turns an authorization rule into a convention. The fix is to make the service actor-aware, for example `updateNotificationForUser(userId, notificationId, changes, transaction)`, and enforce ownership in the `WHERE` clause. If global notifications stay in scope, their per-user state needs the separate model from the earlier schema concern.

specs/actionable-notifications/index.md
```diff
@@ -0,0 +1,195 @@
...
+Deletes a notification with the given ID
+should not be called via handlers, only programmatically by the scheduled job (#6)
+
+```deleteNotificationsByEntityAndType(entityId, notificationType)```
+Deletes all notifications for a given entity and type. Used to invalidate stale notifications when a state change makes them no longer actionable (see [Notification Lifecycle](#notification-lifecycle--stale-notification-cleanup), below).
+Should not be called via handlers — call it inline in the same service function that performs the state change.
+
+```getNotifications(scopes)```
+Retrieve all notifications for given scopes. Includes pagination and sorting. offset based, consistent with the rest of the site
+
```
kryswisnaskas, line 208: Codex said: Medium: `deleteNotificationsByEntityAndType(entityId, notificationType)` is too coarse for the cleanup scenarios the spec now documents. This deletes by entity and type only, but one of the listed scenarios is removing an approver from an Activity Report, which should remove that approver’s pending notification, not every approver notification for that report. As written, the helper will force callers to over-delete. The fix is to make the cleanup API recipient-aware, for example `deleteNotifications({ entityId, notificationType, userId })`, or to include a dedicated recipient-scoped delete method.

kryswisnaskas (commented):
Adding four inline review findings from Codex.

specs/actionable-notifications/index.md
kryswisnaskas: A good repro would be:

- Create an Activity Report with two approvers, Approver A and Approver B.
- Submit the report for approval so the system creates two approval request notifications for the same report and notification type, one for each approver.
- Change the approver assignments after those notifications exist so only Approver A should lose their pending notification, while Approver B should still be able to act.
- Run the cleanup path using `deleteNotificationsByEntityAndType(reportId, ACTIVITY_REPORT_APPROVAL_REQUESTED)`.
- Check notifications for both approvers.

Expected result: only Approver A’s stale notification is deleted.

Actual result with the current API shape: both notifications are deleted, because the delete key only includes `entityId` and `notificationType`, not the approver.

specs/actionable-notifications/index.md
thewatermethod: Permissions are checked in the handler, as is our convention everywhere in the codebase. This low-level service function may be leveraged by another callsite appropriately.

specs/actionable-notifications/index.md
thewatermethod: This is resolved in the implementation PR

specs/actionable-notifications/index.md
thewatermethod: As I add the actual associations, I will leverage [existing sequelize polymorphic affordances](https://sequelize.org/docs/v6/advanced-association-concepts/polymorphic-associations/#configuring-a-many-to-many-polymorphic-association) through a junction table

specs/actionable-notifications/index.md
thewatermethod: This is a good point. I will address in a separate branch and then bring my updates to the spec.

PR #3665, opened 2026-06-01
[TTAHUB-5383] Add notification model

thewatermethod:
## Description of change

Add a model, migration, and seeder for the "Notifications" table

## How to test

Confirm that the schema makes sense and meets requirements, that the spec is accurate, and that the enum captures all the types of notification specified in the CSV

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5383

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


src/constants.js
```diff
@@ -121,6 +121,10 @@ const USER_SETTINGS = {
   },
 };
 
+const NOTIFICATION_TYPES = {
+  ACTIVITY_REPORT_CHANGES_REQUESTED: USER_SETTINGS.EMAIL.KEYS.CHANGE_REQUESTED,
+};
```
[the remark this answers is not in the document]
thewatermethod: Will fix this in the spec PR

AdamAdHocTeam (approved):

src/migrations/20260601000000-create-notifications-table.js
```diff
@@ -0,0 +1,123 @@
...
+              model: 'Users',
+              key: 'id',
+            },
+          },
+          entityId: {
+            type: Sequelize.INTEGER,
+            allowNull: true,
+          },
+          type: {
+            type: Sequelize.ENUM(NOTIFICATION_TYPES),
```
hardwarehuman, line 70: I feel like we'll save ourselves some headaches in the future if we make a `NotificationTypes` table instead of the enums. It just seems likely that we'll end up wanting different notification types handled differently in a way we don't want to have to hardcode.

specs/actionable-notifications/notifications.csv
```diff
@@ -0,0 +1,176 @@
...
+Activity Report,System,Near report submission deadline (5th working day of the month),Email,Submit draft Activity Report R01-AR-17967: Submission deadline 07/05/2026,,Creator,Out of scope,,Actionable Notifications email content,Version 2 per Heather - https://adhoc.slack.com/archives/C022R5301L4/p1774474899507799
+Activity Report,System,Near report submission deadline (5th working day of the month),In-app,The submission deadline for Activity Report R01-AR-17967 is near. ,Take action,Collaborator,Out of scope,,,
+Activity Report,System,Near report submission deadline (5th working day of the month),Email,Submit draft Activity Report R01-AR-17967: Submission deadline 07/05/2026,,Collaborator,Out of scope,,Actionable Notifications email content,
+Activity Report,System,"Activity Report pending approval for ""X' days",In-app,"An Activity Report for [Recipient name] has been pending approval for ""X"" days.",View AR,TTAC/Managers,Out of scope,,,NEW - Wishlist item 03/24/26 Digest?
```
hardwarehuman, line 24: Not important, but you have imbalanced quotes on this and the next line with initial double quotes and following single-quote: `"X'`

src/migrations/20260601000001-add-actionable-notifications-feature-flag.js
```diff
@@ -0,0 +1,21 @@
+const { prepMigration, updateUsersFlagsEnum } = require('../lib/migration');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      const sessionSig = __filename;
+      await prepMigration(queryInterface, transaction, sessionSig);
+      return updateUsersFlagsEnum(
```
hardwarehuman, line 9: it seems like we could do this in the previous migration.

hardwarehuman (commented):
As I look at the use of `entityId` I feel like it would probably be easier not to try to track everything about notification types with enums

src/migrations/20260601000000-create-notifications-table.js
thewatermethod: @hardwarehuman 
Notifications as designed require custom metadata (links, labels, titles, variations) that would be, to my mind, difficult to put into a database without some kludgy JSON columns

Screenshot 2026-06-05 at 3 46 02 PM

I am handling this with linked enums

```js
// NOTIFICATION_TYPES.ACTIVITY_REPORT_NEEDS_ACTION is a valid enum value

const NOTIFICATION_CONFIGURATION = {
  [NOTIFICATION_TYPES.ACTIVITY_REPORT_NEEDS_ACTION]: {
    textFn: ({ userName, recipientName }) =>
      `${userName} has requested changes to your Activity Report for ${recipientName}.`,
    // whether or not we display primary button style or outline button style ("view" vs "take action")
    actionable: true,
    linkFn: ({ id }) => `/activity-reports/${id}`,
    linkText: (metadata) => 'View AR',
    displayId: (metadata) => `${metadata.displayId}`,
    }, 
   // and so forth for each other type. these values are all computed when the notification is inserted
};
```

My feeling is that as we add these bespoke types, it becomes easier to make sure we have everything we need if the JSON and values are colocated and unit-testable. 

Is there another way to solve this with a NotificationTypes table?

specs/actionable-notifications/notifications.csv
thewatermethod: This is just an export from the Google Sheets spec file, I intend to remove this file before actual release of this feature

src/migrations/20260601000000-create-notifications-table.js
hardwarehuman: So to check my understanding: Each notification type has variable internal logic that doesn't really map well to a relational database table structure unless we made that table structure much more complex (or JSONBlobby) than just having a block of code for each type and no need to reference anywhere else.

thewatermethod: @hardwarehuman Yeah that's just about the long and short of how I'm looking at it... example:

The collab report text is
{author's name}  submitted the {activityName} for review

The activity report's corresponding one is
{author's name} has submitted an Activity Report for review

BUT the needs action one is
{approver's name} has requested changes to your Activity Report for {recipient's name}

I _think_ to get this stuff into a database we'd need to contrive some kind of system to replace template strings in a JSON blob? OR we could store arbitrary metadata, keep the strings in code, and do that calculation at display time. It's definitely possible there's an easier path I'm not seeing, but as that implies, I'm not seeing it. I remain open to suggestions though.

hardwarehuman: I had a comment and then navigated away secure in the knowledge that Github saves partial comments, but that knowledge proved false in this case.

The short version is that a table makes sense if for each notification type there's a limited and well-defined set of texts/toggles/flags that define its behavior. I was thinking in terms of where the `entityId` joins, which can only be one of a certain set of tables. You can just have a single method that looks at a column value and knows where to fetch the rest. If this was true of `entityId` and across the board where a few table options can fully define behavior without reference to anything else, then putting it in a table simplifies management.

But this only works if it's straightforward. If every option links to methods that then have to do a bunch more reasoning about the notification and this would need to be updated every time we add or update a notification, then we're just adding complexity and spreading logic around. It sounds like this more complicated scenario is more nearly the case than the straightforward one.

PR #3674, opened 2026-06-03
[TTAHUB-5387] Add notification handlers

thewatermethod:
## Description of change

Add handlers for actionable notifications. Note that only a few operations actually require API endpoints, most of the notification CRUD actions will be handled programmatically by triggers

## How to test

Confirm that the spec is accurate to the UI expectations. Review the code and the tests.

## Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5387

### Every PR

- [ ] Meets issue criteria
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


src/policies/notifications.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    id: 2,
+    permissions: [{ regionId: 1, scopeId: SCOPES.READ_WRITE_REPORTS }],
+  };
+
+  const ownedNotification = { userId: 2 };
+  const otherNotification = { userId: 99 };
+  const globalNotification = { userId: undefined };
+
+  describe('isAdmin', () => {
+    it('returns true when user has the ADMIN scope', () => {
```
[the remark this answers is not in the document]
thewatermethod: [4cb11ef](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/4cb11ef497ab26097a5511c54deb0b954c9221fd)

src/routes/notifications/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  '/admin',
+  userAdminAccessMiddleware,
+  transactionWrapper(createGlobalNotificationHandler)
+);
+router.put(
+  '/:notificationId',
+  checkNotificationIdParam,
+  transactionWrapper(updateNotificationHandler)
+);
+router.get('/', transactionWrapper(getNotificationsHandler));
```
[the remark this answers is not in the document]
thewatermethod: [4cb11ef](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/4cb11ef497ab26097a5511c54deb0b954c9221fd)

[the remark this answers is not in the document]
thewatermethod: [4cb11ef](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/4cb11ef497ab26097a5511c54deb0b954c9221fd)

src/services/notifications.ts
```diff
@@ -154,13 +159,21 @@ async function getNotifications(
   scopes: NotificationScope[],
   { limit = NOTIFICATION_PER_PAGE, offset = 0, sortBy = 'triggeredAt', sortDirection = 'DESC' } = {}
 ) {
+  const sort = ALLOWED_SORT_FIELDS.includes(sortBy as AllowedSortField) ? sortBy : 'triggeredAt';
+  const direction = ALLOWED_SORT_DIRECTIONS.includes(sortDirection as AllowedSortDirection)
+    ? sortDirection
+    : 'DESC';
```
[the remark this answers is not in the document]
thewatermethod: [4cb11ef](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/4cb11ef497ab26097a5511c54deb0b954c9221fd)

src/services/notifications.ts
```diff
@@ -154,13 +159,21 @@ async function getNotifications(
...
+
+  const limitValue = Math.min(Number(limit) || NOTIFICATION_PER_PAGE, 100);
+  const offsetValue = Number(offset || 0);
+
```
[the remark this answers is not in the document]
thewatermethod: [4cb11ef](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/4cb11ef497ab26097a5511c54deb0b954c9221fd)

src/routes/notifications/handlers.test.ts
```diff
@@ -0,0 +1,223 @@
...
+        notificationData.type,
+        {
+          metadata: {
+            id: 123,
+            recipientName: 'System Update',
+            userName: 'System Update',
+            date: notificationData.triggeredAt.toISOString(),
+            displayId: 'SYS-001',
+          },
+        }
```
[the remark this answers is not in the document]
thewatermethod: [4cb11ef](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/4cb11ef497ab26097a5511c54deb0b954c9221fd)

src/routes/notifications/handlers.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+  describe('createGlobalNotificationHandler', () => {
+    it('returns 201 with the created notification', async () => {
+      const notificationData = {
+        type: 'SYSTEM_PLANNED_OUTAGE',
+        id: 123,
+        text: 'System Update',
+        triggeredAt: new Date('2026-06-01'),
+        displayId: 'SYS-001',
+      };
```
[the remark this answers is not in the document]
thewatermethod: [4cb11ef](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/4cb11ef497ab26097a5511c54deb0b954c9221fd)

src/routes/notifications/handlers.ts
```diff
@@ -0,0 +1,95 @@
...
+    const notificationData = req.body;
+    const notification = await createGlobalNotification(notificationData.type, {
+      metadata: {
+        id: notificationData.id ?? null,
+        recipientName: notificationData.text ?? null,
+        userName: notificationData.text ?? null,
+        date: notificationData.triggeredAt?.toISOString() ?? null,
+        displayId: notificationData.displayId ?? null,
+      },
+    });
```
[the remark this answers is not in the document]
thewatermethod: [4cb11ef](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/4cb11ef497ab26097a5511c54deb0b954c9221fd)

src/routes/notifications/handlers.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+export async function createGlobalNotificationHandler(req: Request, res: Response) {
+  // admin access is checked in the middleware
+  try {
+    const notificationData = req.body;
+    const notification = await createGlobalNotification(notificationData.type, {
+      metadata: {
+        id: notificationData.id ?? undefined,
+        recipientName: notificationData.text ?? undefined,
```
AdamAdHocTeam, line 101: Should both map to notification data.

AdamAdHocTeam (approved):
Overall looks good might need minor tweaks as we implement it on the FE but overall good.

AI says (but might just be a result of what we are starting out with implementing):

High — Undefined keys in NOTIFICATION_TYPE_MAP
[notificationType.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

The trainingReport and other buckets reference keys that do not exist in NOTIFICATION_TYPES (e.g., TRAINING_REPORT_POC_ADDED_DIGEST, COMMUNICATION_LOG_TTA_STAFF_ADDED_DIGEST, etc.). Since constants.js defines no digest types, these all evaluate to undefined at runtime. That means Op.in queries will contain undefined values, which can generate invalid SQL or silently corrupt filtering.

The coverage test only checks in one direction (every NOTIFICATION_TYPES value is in a bucket), so it won't catch these extra undefined entries.

Fix: Either add the missing digest enum keys to NOTIFICATION_TYPES / the migration's enum array, or remove the undefined references from the map until those types are defined.

src/routes/notifications/handlers.ts
thewatermethod: can you explain what you mean?

thewatermethod: oh I see: I'll clean up the params when we actually go to make the complete picture, there's way more here than we need

src/services/notifications.ts
```diff
@@ -164,26 +178,66 @@ async function deleteNotificationsByEntityAndType(
...
+  });
+
+  return notifications.map((notification) => {
+    const notificationWithStates = notification as NotificationWithState & {
+      userStates?: NotificationUserStateModel[];
+    };
+    const userState = notificationWithStates.userStates?.[0] ?? null;
+
+    notificationWithStates.userState = userState;
+    notificationWithStates.viewedAt = userState?.viewedAt ?? null;
```
kryswisnaskas: Codex said: GET /notifications is very likely to drop the flattened state fields from the JSON response. In this service the code assigns userState, viewedAt, and archivedAt directly onto Sequelize model instances, and the handler returns those instances with res.json(notifications). Sequelize serializes models via get({ plain: true }), so plain instance properties are not included in toJSON(). Since this PR also removes viewedAt and archivedAt from the Notification model, that means the API contract regresses unless you convert each row to a plain object before merging state, or define/set those values as actual model data or virtual fields. I would fix this by returning plain objects from getNotifications() and adding a handler-level test that asserts the serialized JSON body, not just direct property access on the model instance.

src/routes/notifications/handlers.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    const policy = new NotificationsPolicy(user, notification);
+
+    if (!policy.canUpdateNotification()) {
+      return res
+        .status(StatusCodes.FORBIDDEN)
+        .json({ message: 'User does not have permission to update this notification' });
+    }
+
+    const updated = await updateNotificationState(notification.id, userId, updatedNotification);
```
kryswisnaskas, line 70: Codex said: The admin update path authorizes writes that the data model cannot represent correctly. The policy allows an admin to update any notification, but this handler always writes state for the current caller's userId. For a user-scoped notification owned by someone else, that creates a NotificationUserState row for the admin, even though getNotifications() only returns non-global notifications where Notification.userId matches the caller. So the admin can create state rows they can never read back, and their update does not affect the actual recipient's notification state. Either remove the admin override for non-global notifications or change the update contract so an admin explicitly targets the notification owner.

src/routes/notifications/handlers.ts
```diff
@@ -0,0 +1,89 @@
...
+  } catch (error) {
+    await handleErrors(req, res, error, logContext);
+  }
+}
+
+export async function createGlobalNotificationHandler(req: Request, res: Response) {
+  // admin access is checked in the middleware
+  try {
+    const notificationData = req.body;
+    const notification = await createGlobalNotification(notificationData.type, {
```
kryswisnaskas: Codex said: POST /notifications/admin lacks input validation and will convert bad requests into 500s. Here notificationData.type is passed straight through and notificationData.triggeredAt is fed to new Date(...).toISOString() without checking whether it is valid. An unknown type or invalid date string throws a generic error, and the shared API error handler maps generic errors to 500. This route also accepts any notification type, including templates that are not safe to broadcast globally. I would add explicit payload validation here, return 400 for bad type or triggeredAt, and whitelist the notification types that are valid for this admin endpoint.

src/services/notifications.ts
```diff
@@ -94,33 +101,40 @@ async function createGlobalNotification(
...
   for (const field of allowedFields) {
-    if (field in updatedNotification) {
-      fieldsToUpdate[field] = updatedNotification[field];
+    if (field in updates) {
+      fieldsToUpdate[field] = updates[field];
     }
   }
 
-  return notification.update(fieldsToUpdate);
+  let state = await NotificationUserState.findOne({
```
kryswisnaskas, line 124: Could we match the spec and make this a real `upsert` instead of `findOne` followed by `create/update`? My concern is that, even inside the request transaction, the current read then write flow can still race under concurrent requests and fail on the unique constraint rather than behaving like an atomic upsert.

thewatermethod: @kryswisnaskas I can if you want, but I was told that Sequelize + Postgres runs the upsert (really an insert... on conflict do nothing) outside of the transaction wrapper. That's why we converted all our upserts to this pattern years ago

src/routes/notifications/handlers.ts
thewatermethod: [6defd31](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/6defd31332746b1db3ef14fb5fac592432b1a3dc)

src/routes/notifications/handlers.ts
thewatermethod: [6defd31](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/6defd31332746b1db3ef14fb5fac592432b1a3dc)

src/services/notifications.ts
thewatermethod: [6defd31](https://github.com/HHS/Head-Start-TTADP/pull/3674/commits/6defd31332746b1db3ef14fb5fac592432b1a3dc)

PR #3681, opened 2026-06-05
TTAHUB-5344 and TTAHUB-5348 - Compliant Follow-up Reviews with TTA Support widget Table View

Andrew565:
## Description of change

Table view of the Compliant Follow-up Reviews with TTA Support widget for the Regional Dashboard Monitoring page.

## How to test

1. Load it up and see if it works.

## Jira Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5344
* https://jira.acf.gov/browse/TTAHUB-5348

### Every PR

- [X] Linked Jira issue
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


frontend/src/widgets/CompliantFollowUpReviewsWithTtaSupport.js
```diff
@@ -0,0 +1,156 @@
...
+  // Placeholder menu items until export is implemented
+  // const { exportRows } = useWidgetExport(
+  //     tableData,
+  //     [...months, 'Total'],
+  //     {},
+  //     'Finding category',
+  //     EXPORT_NAME
+  //   );
+
+  const menuItems = useWidgetMenuItems(
```
kryswisnaskas, line 73: Codex said: This still exposes an Export table action even though exportRows is never passed. useWidgetMenuItems calls exportRows() whenever the widget is in table mode, so the export action will fail the first time a user clicks it. Either wire useWidgetExport here now, or remove the export menu path until the table export is implemented.

frontend/src/widgets/CompliantFollowUpReviewsWithTtaSupport.js
```diff
@@ -0,0 +1,156 @@
...
+    <div className="margin-bottom-3">
+      <WidgetContainerSubtitle>
+        Compliant follow-up reviews, broken out by those with and without citations addressed by
+        approved activity reports during the correction period.
+      </WidgetContainerSubtitle>
+      <div className="margin-top-1">
+        <DrawerTriggerButton drawerTriggerRef={drawerTriggerRef}>
+          About this data
+        </DrawerTriggerButton>
+      </div>
```
kryswisnaskas, line 91: Codex said: This empty-state logic is checking data.length, but this widget does not consume an array response. The backend returns an object with months and reviews, so data.length is undefined for the real success shape. That means an empty response can fall through and render a blank widget instead of NoResultsFound, and loading currently also routes through the empty-state branch. This should be based on the actual shape, for example !loading && data?.months?.length === 0 or an equivalent check on reviews.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
```diff
@@ -0,0 +1,231 @@
...
+      id: TRACE_IDS.COMPLIANT_FOLLOW_UP_REVIEWS_WITH_TTA_SUPPORT,
+    };
+  }
+
+  // Main query to get the count of follow-up reviews with and without TTA support for each month in the range
+  const rows = await sequelize.query<IMonthlyCounts>(
+    `WITH months AS (
+      SELECT unnest(ARRAY[:monthStarts]::date[]) AS month_start
+    ),
+    review_set AS (
```
kryswisnaskas, line 115: Codex said: Grouping by review_name and then counting COUNT(review_name) is a data-integrity bug. If two different reviews share the same name in the same month, they collapse into one row here, and if review_name is NULL the row is excluded from the counts entirely because COUNT(column) ignores nulls. This should group on a stable review identifier such as dr.id or gdr.id and count COUNT(*) or COUNT(DISTINCT dr.id) instead.

frontend/src/widgets/CompliantFollowUpReviewsWithTtaSupport.js
```diff
@@ -0,0 +1,156 @@
...
+            showTotalColumn
+            stickyFirstColumn
+            stickyLastColumn
+            enableCheckboxes={false}
+            selectAllIdPrefix="compliant-follow-up-reviews"
+            hideFirstColumnBorder
+            footerData={false}
+          />
+        ) : (
+          <CompliantReviewsGrid data={data} />
```
kryswisnaskas: Codex said: Toggling away from the default table view will throw here because CompliantReviewsGrid is neither imported nor defined in this file. Since the menu already exposes Display graph, this is a user-triggered runtime failure. If this PR is intentionally table-only, the graph toggle should be removed until the alternate view exists.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
```diff
@@ -0,0 +1,231 @@
...
+      reviews: [
+        { name: 'Follow-up reviews with TTA', values: x.map(() => 0) },
+        { name: 'Follow-up reviews without  TTA', values: x.map(() => 0) },
+        { name: 'Total', values: x.map(() => 0) },
+      ],
+      id: TRACE_IDS.COMPLIANT_FOLLOW_UP_REVIEWS_WITH_TTA_SUPPORT,
+    };
+  }
+
+  // Main query to get the count of follow-up reviews with and without TTA support for each month in the range
```
kryswisnaskas: Codex said: This aggregation is not scoped to the same data set the widget title describes. The month range above is derived from filtered grant citations and approved reports, but this raw SQL never constrains gdr to the filtered grants, never applies the delivered-review filters, and never limits the TTA join to the scoped citations or approved reports. With dashboard filters active, this can over-count out-of-scope reviews and mark a review as with TTA because of unrelated data. A safer fix is to mirror the monitoringOverview scoping here: derive the scoped grant/review/report IDs first, then apply them explicitly inside this query.

frontend/src/widgets/CompliantFollowUpReviewsWithTtaSupport.js
```diff
@@ -0,0 +1,156 @@
...
+import withWidgetData from './withWidgetData';
+import './CompliantFollowUpReviewsWithTtaSupport.css';
+import NoResultsFound from '../components/NoResultsFound';
+
+const EXPORT_NAME = 'Compliant follow-up reviews with TTA support';
+
+export function CompliantFollowUpReviewsWithTtaSupport({ loading, data }) {
+  const { setIsAppLoading } = useContext(AppLoadingContext);
+  const drawerTriggerRef = useRef(null);
+  const capture = useMediaCapture(EXPORT_NAME);
```
kryswisnaskas: Codex said: The menu actions are wired as if screenshot/export support exists, but this hook call does not match the useMediaCapture contract. The hook expects a ref plus a title; here it only receives the title string, so the screenshot path will fail when clicked. If this widget is not ready to support screenshots and exports yet, the safer option is to suppress those menu items until the ref and export wiring are implemented.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
```diff
@@ -0,0 +1,179 @@
...
+    JOIN "DeliveredReviewCitations" drc
+      ON drid = drc."deliveredReviewId"
+    JOIN "GrantCitations" gc_scoped
+      ON gc_scoped."citationId" = drc."citationId"
+      AND gc_scoped.id IN (:grantCitationIds)
+    JOIN "Citations" c
+      ON c.id = gc_scoped."citationId"
+      AND c."deletedAt" IS NULL
+    LEFT JOIN "ActivityReportObjectiveCitations" aroc
+      ON aroc."citationId" = c.id
```
kryswisnaskas, line 132: Codex said: This still does not apply scopes.activityReport when deciding whether a review has TTA support. The join only checks that the activity report is approved and that its startDate falls in the correction period, so any approved AR on the citation can count even if it falls outside the dashboards activity-report filters. That means the numerator can still drift from the filtered dashboard state. The fix is to precompute scoped approved report IDs with the ORM, like monitoringOverview.ts does, and join or filter on those IDs here.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  if (!grantCitations.length) {
+    return EMPTY_RESULT;
+  }
+
+  const grantIds = uniq(grantCitations.map((gc: { grantId: number }) => gc.grantId));
+  const grantCitationIds = grantCitations.map((gc: { id: number }) => gc.id);
+
+  // Apply the deliveredReview scope (e.g. review_type) and grant filter via the ORM, since
+  // WhereOptions can't be injected into raw SQL. The resulting ids are passed to the SQL where
```
kryswisnaskas, line 65: Codex said: This still never enforces the widgets own business definition of compliant follow-up reviews. DeliveredReview.findAll only applies scopes.deliveredReview, complete: true, and the date bound; it does not add outcome: Compliant or review_type: Follow-up. Unless every caller always injects those constraints, this widget can count complete reviews that are not compliant follow-ups. The fix is to hardcode those predicates in this query, the same way monitoringOverview.ts does.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
Andrew565: @copilot Apply the suggested fix in https://github.com/HHS/Head-Start-TTADP/pull/3681#discussion_r3423035217

thewatermethod:
Bottom row stickiness seems a bit funky

https://github.com/user-attachments/assets/f3a3dd6d-0c8c-4474-b986-7b3c09e6e937

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
```diff
@@ -0,0 +1,178 @@
...
+  // scoped_reviews re-fetches the necessary columns and drives the rest of the query.
+  const deliveredReviews = await DeliveredReview.findAll({
+    attributes: ['id'],
+    where: {
+      [Op.and]: [
+        ...scopes.deliveredReview,
+        { complete: true },
+        { complete_date: { [Op.gte]: MIN_MONITORING_DATE } },
```
thewatermethod, line 73: @Andrew565 @hardwarehuman the design says the following: "Date filtering: The date range you select at the top of the dashboard filters the data based on the Follow-up review received date." 

The scopes applied are to "DeliveredReview.review_delivery_date" but it looks like this sequelize function will only return the initial, now-completed review, which will make the date filter function in unexpected ways. Am I misunderstanding what this query is returning?

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
Andrew565: Tried this but the bot hallucinated that scopes.activityReport was a thing. It doesn't appear to actually be passed in here. Reverted to the previous iteration.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
```diff
@@ -0,0 +1,178 @@
...
+    return EMPTY_RESULT;
+  }
+
+  const grantIds = uniq(grantCitations.map((gc: { grantId: number }) => gc.grantId));
+  const grantCitationIds = grantCitations.map((gc: { id: number }) => gc.id);
+
+  // Apply the deliveredReview scope (e.g. review_type) and grant filter via the ORM, since
+  // WhereOptions can't be injected into raw SQL. The resulting ids are passed to the SQL where
+  // scoped_reviews re-fetches the necessary columns and drives the rest of the query.
+  const deliveredReviews = await DeliveredReview.findAll({
```
kryswisnaskas, line 67: Codex said: This still does not enforce the widgets own business definition of compliant follow-up reviews. DeliveredReview.findAll only applies scopes.deliveredReview, complete: true, and the date bound; it does not add outcome: Compliant or review_type: Follow-up. Unless every caller always injects those constraints, this widget can count complete reviews that are not compliant follow-ups. The fix is to hardcode those predicates here, the same way monitoringOverview.ts does.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
```diff
@@ -0,0 +1,178 @@
...
+      ON gc_scoped."citationId" = drc."citationId"
+      AND gc_scoped.id IN (:grantCitationIds)
+    JOIN "Citations" c
+      ON c.id = gc_scoped."citationId"
+      AND c."deletedAt" IS NULL
+    LEFT JOIN "ActivityReportObjectiveCitations" aroc
+      ON aroc."citationId" = c.id
+    LEFT JOIN "ActivityReportObjectives" aro
+      ON aroc."activityReportObjectiveId" = aro.id
+    LEFT JOIN "ActivityReports" ar
```
kryswisnaskas, line 135: Codex said: This still does not apply scopes.activityReport when deciding whether a review has TTA support. The join only checks that the activity report is approved and that its startDate falls in the correction period, so any approved AR on the citation can count even if it falls outside the dashboards activity-report filters. That means the numerator can still drift from the filtered dashboard state. The fix is to precompute scoped approved report IDs with the ORM, like monitoringOverview.ts does, and join or filter on those IDs here.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
Andrew565: This is a misunderstanding of the schema: Adding the predicates "outcome: compliant" etc would be incorrect and exclude reviews that have been made compliant by all of their citations being addressed. The proper way to determine compliant follow up reviews is to find reviews that are complete.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
Andrew565: Tried adding in a check for scopes.activityReport here but the linter said it wasn't accessible and failed, so I think this is just another hallucination. The SQL does enough here to filter using the DeliveredReviews scope.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
hardwarehuman: Reviews where `DeliveredReviews.corrected` is true, which incorporates both `DeliveredReviews.corrected` and that all findings were actually corrected in the end.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
hardwarehuman: The real issue is probably that we need to update the existing widget to match, as we discussed in standup. I don't think that's been done yet in this PR.

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
kryswisnaskas: The issue is not whether s`copes.activityReport `can be dropped directly into the raw SQL string; it usually cannot, which is why lint complained. The issue is that the current SQL only filters ActivityReports by approved status and correction period date, not by the dashboard activityReport scope. In src/widgets/monitoring/monitoringOverview.ts , the `ActivityReport` include explicitly applies both constraints:
```
where: {
  [Op.and]: [
    { calculatedStatus: REPORT_STATUSES.APPROVED },
    ...scopes.activityReport,
  ],
},
```
That is the behavior this widget is missing. scopes.deliveredReview only constrains the review set; it does not constrain which approved ActivityReports are allowed to satisfy has_tta. The correct fix is to precompute the scoped approved ActivityReport IDs with the ORM and then join or filter on those IDs in the SQL, following the same pattern.

frontend/src/pages/RegionalDashboard/components/MonitoringReportDashboard.js
```diff
@@ -24,7 +25,9 @@ export default function MonitoringReportDashboard({ filtersToApply }) {
         <FindingCategoryHotspot filters={filtersToApply} />
       </Grid>
       <Grid row>
-        <CompliantFollowUpReviewsWithTtaSupport filters={filtersToApply} />
+        <FeatureFlag flag="compliant_follow_up_reviews_tta_support">
```
thewatermethod: 'compliant_follow_up_reviews_tta_support' => 'monitoring-regional-dashboard'

src/migrations/20260623154010-add-compliant-follow-up-reviews-tta-support-flag.js
thewatermethod, line 1: oh I see you added a flag. There is already one hanging around for monitoring regional dash, I should have mentioned

Andrew565: Hmm, gotcha. Should I change to the existing one?

thewatermethod: I think so, that's why I left it in there (was assuming I would do this widget), that way we have less to cleanup in the future

src/widgets/monitoring/compliantFollowUpReviewsWithTtaSupport.ts
kryswisnaskas: This looks correct now. The only thing I want to note is that this widget handles the Date filter in a different way from `monitoringOverview`. Instead of applying Activity Report scope in the backend, it depends on the Monitoring page turning Date into `reportDeliveryDate` first. I do not think that is a bug, but it is different from the usual pattern and harder to spot when reading the code. A small test for that Date filter behavior would help make the intent clear.

kryswisnaskas (approved):

src/migrations/20260623154010-add-compliant-follow-up-reviews-tta-support-flag.js
Andrew565: Further convo with Matt, he said it's not really a big deal, we just have to remember to remove this feature flag once it's ready.

thewatermethod:
> Bottom row stickiness seems a bit funky
Screen.Recording.2026-06-18.at.10.44.15.AM.mov

@Andrew565 I'm still seeing the bottom row is not sticky, otherwise looks good

frontend/src/widgets/CompliantFollowUpReviewsWithTtaSupport.js
```diff
@@ -0,0 +1,351 @@
+import PropTypes from 'prop-types';
```
AdamAdHocTeam, line 1: This is a large file can anything be broken out into a helper file?

PR #3749, opened 2026-07-06
[TTAHUB-5517/5518] Remove FEI and Specialist role filters from TTA History

AdamAdHocTeam:
## Description of change

This PR removes two filters from the TTA History tab (TTAHUB-5517/TTAHUB-5518). The FEI response and the Specialist role filters.

## How to test

- Review the code
- Make sure neither filter appears on the TTA History tab
- Make sure the existing filters till work

## Jira Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5517
* https://jira.acf.gov/browse/TTAHUB-5518

### Every PR

- [x] Linked Jira issue
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy

### DIFF
--- a/frontend/src/pages/RecipientRecord/pages/TTAHistory.js
+++ b/frontend/src/pages/RecipientRecord/pages/TTAHistory.js
@@ -1,6 +1,6 @@
 import { Grid } from '@trussworks/react-uswds';
 import PropTypes from 'prop-types';
-import React, { useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react';
+import React, { useCallback, useContext, useMemo, useState } from 'react';
 import { Helmet } from 'react-helmet';
 import { v4 as uuidv4 } from 'uuid';
 import ActivityReportsTable from '../../../components/ActivityReportsTable';
@@ -25,7 +25,10 @@ const VALID_TOPICS = new Set(TTAHISTORY_FILTER_CONFIG.map((f) => f.id));
 
 export default function TTAHistory({ recipientName, recipientId, regionId }) {
   const [resetPagination, setResetPagination] = useState(false);
-  const filterKey = `ttahistory-filters-${recipientId}`;
+  // Bump the version suffix whenever filters are added or removed from the
+  // config so that stale filters saved in a browser tab's session storage are
+  // ignored rather than sent to the API.
+  const filterKey = `ttahistory-filters-v2-${recipientId}`;
   const { user } = useContext(UserContext);
   const regions = useMemo(() => getUserRegions(user), [user]);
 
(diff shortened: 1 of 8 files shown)


Andrew565 (approved):

frontend/src/pages/RecipientRecord/pages/constants.js
```diff
@@ -30,9 +28,7 @@ export const getGoalsAndObjectivesFilterConfig = (grantNumberParams) =>
 const TTAHISTORY_FILTER_CONFIG = [
   { ...startDateFilter, display: 'Date started' },
   { ...endDateFilter, display: 'Date ended' },
-  activityReportGoalResponseFilter,
   myReportsFilter,
```
kryswisnaskas, line 31: Codex said: stale `role` and `activityReportGoalResponse` filters can still silently affect TTA History results. The PR removes those options from `TTAHISTORY_FILTER_CONFIG`, but TTA History still loads filters from URL/session storage and applies all of them to the table/widgets. A user with a bookmarked `?role.in[]=...` URL or an existing `ttahistory-filters-<recipientId>` session value will keep seeing filtered data from controls that no longer exist. The pill UI also loses the filter name because `FilterPills.js` only resolves labels from the current config.

Concrete fix: sanitize TTA History filters against `TTAHISTORY_FILTER_CONFIG` before rendering and before `expandFilters(filters)`, and persist the cleaned list back to the hook so session storage and URL are corrected. Add a regression test that seeds either `?role.in[]=Grantee%20Specialist` or session storage and asserts no `role.in[]` / `activityReportGoalResponse.in[]` request is made.

frontend/src/pages/RecipientRecord/pages/TTAHistory.js
```diff
@@ -35,6 +38,17 @@ export default function TTAHistory({ recipientName, recipientId, regionId }) {
     },
   ]);
 
+  const sanitizedFilters = useMemo(
```
kryswisnaskas: Codex said: stale bookmark cleanup drops the default year-to-date filter and broadens TTA History to all dates. In `TTAHistory.js`, `useSessionFiltersAndReflectInUrl` still gives URL filters precedence over the default start-date filter. The new sanitization removes invalid topics, but if a bookmark contains only a removed filter like `?role.in[]=Grantee%20Specialist`, `sanitizedFilters` becomes `[]` and the page fetches without `startDate.win`. That changes the page from its normal year-to-date default to an unbounded history query, which is both a behavior regression and a potential performance hit on widgets/table requests. The new URL test only asserts that `role.in[]` is absent; it does not verify that `startDate.win` is still present, so this regression can slip through.

Concrete fix: when sanitization removes all URL/session-derived filters, fall back to the default start-date filter instead of persisting `[]`. Practically, extract the default TTA History filter array to a constant and have the cleanup path replace an empty sanitized set with that default. Add a regression test that starts from `?role.in[]=Grantee%20Specialist` and asserts the resulting requests still include `startDate.win=<year-to-date-range>`.

kryswisnaskas (approved):
If loading all history after stripping stale bookmark filters is intentional, then I agree this isn’t a bug.

frontend/src/pages/RecipientRecord/pages/TTAHistory.js
```diff
@@ -35,6 +38,32 @@ export default function TTAHistory({ recipientName, recipientId, regionId }) {
...
+    return next;
+  }, [filters]);
+
+  useEffect(() => {
+    if (sanitizedFilters.length !== filters.length) {
+      setFiltersInHook(sanitizedFilters);
+    }
+  }, [sanitizedFilters, filters.length, setFiltersInHook]);
+
   const setFilters = useCallback(
```
thewatermethod, line 48: 1) why not sanitize the filters here in "setFiltersInHook" or in the "filtersToApply" useMemo instead of all that stuff you're doing above? 

2) I don't think in general we should start adding code to deal with possible bookmarks just because the AI reviewers are worried about it, that's been explicitly a "no" every time I ask the stakeholders about it. 

3) As far as session storage: ditto, but also we could simply add a version number to increment the cache key on line 28 and then old session storage is ignored. Since session storage is literally confined to a browser tab, I don't really see any downside to doing that anyways.

I'm concerned because we added a substantial amount of complexity (multiple hook calls, three lists of filters) to this file to deal with what I think are probably be edge cases.

AdamAdHocTeam: Thanks I alway for get we have a version on the cache.

AdamAdHocTeam: this has been fixed.

thewatermethod (approved):

PR #3825, opened 2026-08-06
TTAHUB-5633 - RTR TTA History TR Table Backend

Andrew565:
## Description of change

Backend service for the upcoming RTR TTA History TR table.

## How to test

1. Run the tests locally, ensure they all pass.

## Jira Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5633

### Every PR

- [X] Linked Jira issue
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


src/routes/sessionReports/handlers.ts
```diff
@@ -330,12 +329,16 @@ export const getSessionReportsHandler = async (req: Request, res: Response) => {
...
+
     // Previously, we returned a FORBIDDEN (403) error after checking to see if
     // a user had regions, however, missing region URL params would cause
     // overly permissive access to session reports.
     // Using setTrainingReportReadRegions switches to the pattern used throughout the rest
     // of our application/handlers (the user experience will be the same: an empty sessions table)
-    const filteredFilterParams = await setTrainingReportReadRegions(normalizedFilterParams, userId);
+    const filteredFilterParams = recipientId
+      ? normalizedFilterParams
+      : await setTrainingReportReadRegions(normalizedFilterParams, userId);
```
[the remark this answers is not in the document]
Andrew565: @thewatermethod @AdamAdHocTeam @kryswisnaskas What do you all think about this edge case? Would it make sense to add the admin check?

src/routes/sessionReports/handlers.ts
```diff
@@ -330,12 +329,16 @@ export const getSessionReportsHandler = async (req: Request, res: Response) => {
...
 
+    const recipientId = [req.query.recipientId].flat()[0];
+
     // Previously, we returned a FORBIDDEN (403) error after checking to see if
     // a user had regions, however, missing region URL params would cause
     // overly permissive access to session reports.
     // Using setTrainingReportReadRegions switches to the pattern used throughout the rest
     // of our application/handlers (the user experience will be the same: an empty sessions table)
-    const filteredFilterParams = await setTrainingReportReadRegions(normalizedFilterParams, userId);
+    const filteredFilterParams = recipientId
```
kryswisnaskas, line 339: Claude said: Supplying recipientId skips region-based authorization entirely, letting any authenticated user pull training session data for any recipient in any region.

The `/api/session-reports` route (src/routes/sessionReports/index.js:17) has no other authorization middleware — `setTrainingReportReadRegions` inside the handler is the only access-control gate, restricting results to the regions the caller is permitted to read. When `recipientId` is present, the handler skips that call entirely (`filteredFilterParams = recipientId ? normalizedFilterParams : await setTrainingReportReadRegions(...)`) and routes to `getSessionReportsByRecipient`, which explicitly strips any region filters (`removeRegionFilters`) and queries across all regions. A logged-in user with access to only Region 1 can call `GET /api/session-reports?recipientId=<any-id>` and receive that recipient's session/training report data (session names, dates, topics, participant/recipient details) for grants in regions they have no assigned access to. There is no check that the caller's read regions intersect the recipient's grant regions.

Andrew565: This is intentional, see my previous comment/question on whether or not we should add an admin check.

src/routes/sessionReports/handlers.ts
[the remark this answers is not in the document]
thewatermethod: I don't follow. Wouldn't adding an admin check make it so that the desired behavior (users see sessions outside their regions) wouldn't work? Or are we saying something else?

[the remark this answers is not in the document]
Andrew565: Hmm, yeah, I think you're right.

src/routes/sessionReports/handlers.ts
kryswisnaskas: Codex said: If cross-region RTR visibility is intended for non-admin users, the fix does not need to be admin-only, but it still needs an explicit authorization gate for that specific capability. Right now any authenticated user can trigger the bypass with `recipientId`; we should restrict it to whatever role/scope is allowed to view recipient TR history across regions, or intersect results with readable regions otherwise.

thewatermethod: You're proposing adding a new permission? I think we should check with OHS before doing that

kryswisnaskas: This is not necessarily a new permission suggestion. The issue is that `/api/session-reports?recipientId=... `currently allows the intended cross-region behavior without checking that the caller can view that recipient’s RTR. So the fix should preserve “all sessions for the recipient’s grants regardless of TR region,” but add a check that proves the caller is allowed to view that recipient record.

Andrew565: Is there a suggestion as to how to implement this?

thewatermethod: check the user's permissions versus the recipient/region but don't filter the results by the region

thewatermethod:
@Andrew565 just lmk when this is ready for a re-look

src/routes/sessionReports/handlers.ts
thewatermethod: @Andrew565 your approach makes sense to me

src/services/sessionReports.ts
```diff
@@ -574,3 +601,211 @@ export async function getSessionReports(
     rows,
   };
 }
+
+/**
+ * Clone of getSessionReports that filters rows to sessions listing grants
+ * from a specific recipient, and ignores any region restrictions.
+ */
+export async function getSessionReportsByRecipient(
```
thewatermethod, line 611: This is close to a verbatim replica of the `getSessionReports` function right above it, which will be a maintainability problem if it persists. We could definitely dry this up with one main`fetchSessionReports` function and some helper functions to handle the sorting, and then `getSessionReportsByRecipient` could just handle the recipient ID/region ID stripping stuff

src/services/sessionReports.ts
```diff
@@ -394,36 +396,44 @@ function sessionReportDateSort(field: string): string {
...
+      ...acc,
+      [key]: filters[key],
+    };
+  }, {});
+}
+
+function recipientGrantFilter(grantIds: number[]) {
+  return sequelize.literal(`EXISTS (
+    SELECT 1
+    FROM jsonb_array_elements("SessionReportPilot"."data"->'recipients') elem
```
kryswisnaskas, line 415: Codex said: Non-blocking hardening suggestion: `jsonb_array_elements("SessionReportPilot"."data"->'recipients')` assumes `data.recipients` is always a JSON array. If a legacy or malformed completed session has missing/null/object/string `recipients`, Postgres can throw `cannot extract elements from a scalar` and fail the whole recipient TR history request.

Could make this resilient by treating non-array values as an empty array:

```ts
function recipientGrantFilter(grantIds: number[]) {
  return sequelize.literal(`EXISTS (
    SELECT 1
    FROM jsonb_array_elements(
      CASE
        WHEN jsonb_typeof("SessionReportPilot"."data"->'recipients') = 'array'
        THEN "SessionReportPilot"."data"->'recipients'
        ELSE '[]'::jsonb
      END
    ) elem
    WHERE (
      jsonb_typeof(elem->'value') = 'number'
      OR (jsonb_typeof(elem->'value') = 'string' AND elem->>'value' ~ '^[0-9]+$')
    )
    AND (elem->>'value')::integer IN (${grantIds.join(', ')})
  )`);
}
```

A small regression test with a completed session missing `data.recipients` would cover this.

kryswisnaskas (approved):
Thanks for adding the recipient policy gate. Approving now; I left one non-blocking hardening suggestion for graceful handling of malformed legacy recipients data.

PR #3855, opened 2026-08-20
[TTAHUB-5663] Add recipient timeline shell

kryswisnaskas:
## Description of change

Adds the TTA Timeline tab/page to the recipient record, gated behind a new `tta_timeline` feature flag (added via migration). The page renders filters (date, communication purpose, goal category, TTA event type), a sort direction dropdown, and a "hide multi-recipient communications" toggle, and fetches data from the `/recipient/:recipientId/region/:regionId/timeline` endpoint (added in [TTAHUB-5654](https://jira.acf.gov/browse/TTAHUB-5654), this branch's parent) via a new `getRecipientTimeline` fetcher. The backend service is currently a stub returning an empty `{ count: 0, events: [] }` contract, so the page only shows the empty state today — this PR is the frontend shell; real event data is follow-up work behind the same flag.

Builds on `kw/ttahub-5654` . This diff only includes the frontend Timeline page/fetcher/tab wiring and the feature flag migration.

## How to test

1. Run `yarn db:migrate` to add the `tta_timeline` feature flag enum value.
2. Enable the `tta_timeline` flag for your user and open a recipient record page; confirm a new "TTA Timeline" tab appears and links to the Timeline page.
3. On the Timeline page, confirm filters (date, purpose, goal category, event type), the sort dropdown, and the "Hide multi-recipient communications" checkbox render, and that the empty state ("No timeline events found") shows since the backend service is still a stub.
4. Confirm the tab/route are not visible/reachable when the `tta_timeline` flag is off (renders not found).
5. Verify the unit tests pass.

## Jira Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5663

### Every PR

- [x] Linked Jira issue
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


frontend/src/pages/RecipientRecord/pages/Timeline.tsx
```diff
@@ -0,0 +1,274 @@
...
+  ) => (
+    <FilterSelect
+      onApply={onApplyQuery}
+      inputId={`event-type-${condition}-${id}`}
+      labelText="Select TTA event types to filter by"
+      options={EVENT_TYPES}
+      selectedValues={query}
+    />
+  ),
+};
```
thewatermethod: can we decompose filters to their own file like in other places in the codebase?

frontend/src/pages/RecipientRecord/pages/Timeline.tsx
```diff
@@ -0,0 +1,274 @@
...
+  id: string;
+  topic: string;
+  condition: string;
+  query: string | string[];
+}
+
+interface TimelineResponse {
+  count: number;
+  events: unknown[];
+}
```
thewatermethod: should this type be shared across the backend?

AdamAdHocTeam (commented):
1. Region-scoping change to checkRecipientAccessAndExistence also affects monitoring routes — confirm intent — (severity: medium). In src/routes/utils.ts the existence check changed from recipientById(recipientId, []) to recipientById(validatedRecipientId, { where: { regionId: validatedRegionId } }). This function is shared with the four monitoring handlers in src/routes/monitoring/handlers.ts. Previously a recipient resolved regardless of region; now it must have a qualifying grant in the requested region or the call returns 404. For monitoring that may be more correct, but it's a behavioral change to existing endpoints outside this feature's scope — confirm it's intended and that no monitoring flow relies on cross-region recipient lookups.

2. Fetcher validates the parsed ID but sends the raw string in the URL — (severity: low). In frontend/src/fetchers/recipient.js, parseInt('1abc', 10) returns 1 and passes the Number.isNaN guard, but the request path is built from the original recipientId/regionId via join(recipientUrl, recipientId, 'region', regionId, 'timeline'). Harmless because the backend re-validates strictly, but the guard gives false confidence. Use the already-parsed id/idRegion in join(...).

3. Timeline filter topic/condition whitelists are duplicated with no shared source — (severity: low). src/routes/recipient/middleware.ts hardcodes TIMELINE_FILTER_TOPICS, DATE_FILTER_CONDITIONS, and SELECT_FILTER_CONDITIONS. These currently match the frontend (goalCategoryFilter → standard, purposeFilter → purpose, DATE_CONDITIONS/FILTER_CONDITIONS), so no bug today — but the lists must stay in lockstep manually, and a future frontend condition/topic addition will silently produce a 400. Consider sourcing these from @ttahub/common.

AdamAdHocTeam (commented):
Just double check we don't already have these:

1. parsePositiveInteger overlaps with the existing safeParseInt. src/lib/safeParse.ts already has safeParseInt(value): number | null doing a similar job, but the new src/lib/parsePositiveInteger.ts is intentionally stricter — it rejects scientific notation (1e1), trailing chars (1abc), zero, and negatives, whereas safeParseInt uses lenient parseInt and accepts all of those. Not a true duplicate (the strictness is the point of the security hardening), but the repo now has two similarly-named int parsers with different semantics. Worth either a short comment on parsePositiveInteger explaining why safeParseInt wasn't used, or a follow-up to consolidate. At minimum, future devs should know safeParseInt is the lenient one.

2. checkRecipientTimelineQuery re-implements existing validation-middleware boilerplate. The validate → auditLogger.error → res.status(400).send(msg) block in src/routes/recipient/middleware.ts is the same shape already used in src/routes/activityReports/middleware.ts (checkReviewReportBody/checkSubmitReportBody) and in checkIdParamMiddleware. Following the convention is fine, but the boilerplate is copy-pasted because there's no shared wrapper. Pre-existing pattern — extracting a small helper is optional, not required.

3. IDs are parsed twice on the timeline route. The chain is checkRecipientIdParam → checkRegionIdParam → checkRecipientTimelineQuery → handler. The two checkIdParam calls already parse and store res.locals.validatedParams. Then checkRecipientAccessAndExistence re-derives them via res.locals.validatedParams?.recipientId ?? parsePositiveInteger(...). Redundant for this route — but the ?? fallback is what keeps the monitoring routes (which don't run those middlewares) working, so it's harmless, just a double-parse worth being aware of.

Good reuse (no reinvention): FeatureFlag, FilterPanel / FilterSelect / FilterDateRange, purposeFilter, goalCategoryFilter, useFetch, formatDateRange, DECIMAL_BASE, and the migration helpers (prepMigration, updateUsersFlagsEnum) are all correctly reused rather than duplicated.

The only one worth acting on is #1 (clarify or consolidate the two int-parsers); #2 and #3 are minor/informational.

frontend/src/fetchers/recipient.js
```diff
@@ -86,3 +86,27 @@ export const getRecipientLeadership = async (recipientId, regionId) => {
...
+    throw new Error('Recipient ID must be a number');
+  }
+
+  const idRegion = parseInt(regionId, DECIMAL_BASE);
+  if (Number.isNaN(idRegion)) {
+    throw new Error('Region ID must be a number');
+  }
+
+  const query = new URLSearchParams({ direction });
+  filters.forEach((filter) => query.append('filters', filter));
```
thewatermethod, line 119: Why aren't we leveraging existing helpers/filter code?

frontend/src/pages/RecipientRecord/pages/Timeline.tsx
```diff
@@ -0,0 +1,274 @@
...
+    query: LAST_TWELVE_MONTHS,
+  },
+];
+
+const serializeFilter = ({ topic, condition, query }: TimelineFilter) =>
+  JSON.stringify({ topic, condition, query });
+
+export default function Timeline({ recipientId, regionId }: TimelineProps): React.ReactElement {
+  const aboutDrawerRef = useRef<HTMLButtonElement>(null);
+  const [filters, setFilters] = useState<TimelineFilter[]>(INITIAL_FILTERS);
```
thewatermethod: I'd look into the useFilters hook

frontend/src/pages/RecipientRecord/pages/Timeline.tsx
```diff
@@ -0,0 +1,274 @@
...
+const INITIAL_FILTERS: TimelineFilter[] = [
+  {
+    id: uuidv4(),
+    topic: 'date',
+    condition: 'is within',
+    query: LAST_TWELVE_MONTHS,
+  },
+];
+
+const serializeFilter = ({ topic, condition, query }: TimelineFilter) =>
```
thewatermethod: We may already have code to do this, as our filter implementation is robust and shared across many pages

frontend/src/pages/RecipientRecord/pages/Timeline.tsx
kryswisnaskas: Addressed.

frontend/src/pages/RecipientRecord/pages/Timeline.tsx
kryswisnaskas: This is just the response shape. I left it local for now since this PR only needs `count` and an empty `events` array. TTAHUB-5657 should define the real event shape, so that’s the better place to share it across the backend.

frontend/src/pages/RecipientRecord/pages/Timeline.tsx
kryswisnaskas: Addressed.

frontend/src/fetchers/recipient.js
kryswisnaskas: Addressed.

frontend/src/pages/RecipientRecord/pages/Timeline.tsx
kryswisnaskas: Addressed.

kryswisnaskas:
Thanks @AdamAdHocTeam . I addressed the actionable parts here: the fetcher now rejects malformed IDs and uses the parsed IDs in the URL, and I consolidated the stricter positive int parsing into the existing helper path.

The region scoping change is intentional since this endpoint should only resolve recipients for the requested region. I left the filter topic/condition contract local for now, but moved Timeline onto the shared `useFilters` pattern; TTAHUB-5657 should be the better place to share the fuller version if it grows.

AdamAdHocTeam (approved):

PR #3856, opened 2026-08-20
TTAHUB-5637 & TTAHUB-5642 - Training Reports Table Frontend and Export

Andrew565:
## Description of change

Frontend for the RTR -> TTA History -> Training Reports table.

## How to test

1. Load up the app, navigate to the RTR -> TTA History page
2. View the new table, make sure it looks right
3. Attempt to do various exports, make sure they work right

## Jira Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5637
* https://jira.acf.gov/browse/TTAHUB-5642

### Every PR

- [X] Linked Jira issue
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


frontend/src/pages/RecipientRecord/pages/TTAHistory.js
```diff
@@ -9,16 +9,65 @@ import Drawer from '../../../components/Drawer';
...
+      offset: 0,
+    },
+    'trainingReportsTable'
+  );
+
+  const requestTrainingReportsSort = useRequestSort(setTrainingReportsSortConfig);
+
+  const { data: trainingReportsData, error: trainingReportsError } = useFetch(
+    { rows: [], count: 0 },
+    () => getSessionReportsTable(trainingReportsSortConfig, [], recipientId),
```
kryswisnaskas: Is passing an empty filter array intentional for this iteration?

AdamAdHocTeam (commented):
Double check this: 

Main suggestion:
The Training Reports table on TTA History should use the same active page filters as the rest of the page. Right now it appears to fetch with an empty filters array, which can make Training Reports disagree with the overview/widgets/activity reports when users apply filters.

Why this matters:
Users can apply filters and see most page sections update, but Training Reports may still show unfiltered data. That creates a correctness/trust issue in the UI.

Recommended change:
Pass filtersToApply into the Training Reports fetch and table props in TTA History (same pattern already used in Regional Dashboard Training Reports), so all sections share one filtering contract.

Test gap to address:
Add or update a TTA History test that verifies session-reports requests include active page filters (not just recipientId/sort). Current added tests mostly cover sort scoping by recipient and won’t catch this filter mismatch.

frontend/src/pages/RecipientRecord/pages/TTAHistory.js
Andrew565: Updated to make better use of shared filter state per Adam's suggestion.

Andrew565:
@kryswisnaskas @AdamAdHocTeam Incorporated Adam's feedback, ready for re-review.

frontend/src/pages/RecipientRecord/pages/TTAHistory.js
```diff
@@ -9,16 +9,67 @@ import Drawer from '../../../components/Drawer';
...
+      offset: 0,
+    },
+    'trainingReportsTable'
+  );
+
+  const requestTrainingReportsSort = useRequestSort(setTrainingReportsSortConfig);
+
+  const { data: trainingReportsData, error: trainingReportsError } = useFetch(
+    { rows: [], count: 0 },
+    () => getSessionReportsTable(trainingReportsSortConfig, filters, recipientId),
```
kryswisnaskas, line 55: Codex said: **[P1] “My reports” TR roles will return no Training Reports.**

The shared filters now reach `/api/session-reports`, but that handler does not pass the authenticated `userId` into `filtersToScopes`. Therefore `TR event creator`, `TR event collaborator`, and `TR POC` filters receive an undefined user ID and resolve to `id = null`. This means the Training Reports table and CSV exports return no results when a user selects one of those TR roles.

frontend/src/pages/RecipientRecord/pages/TTAHistory.js
```diff
@@ -9,16 +9,67 @@ import Drawer from '../../../components/Drawer';
...
+    [trainingReportsSortConfig, filters, recipientId],
```
kryswisnaskas, line 56: Codex said: **[P2] Applying filters retains the previous page offset.**

If a user is on page 2 or later and changes filters, the request still uses the previous `activePage` and `offset`. A narrower result can therefore display an empty table even when page 1 contains matching rows. The Activity Reports table resets pagination when filters change, but the Training Reports table does not.

frontend/src/pages/RecipientRecord/pages/TTAHistory.js
Andrew565: Addressed.

frontend/src/pages/RecipientRecord/pages/TTAHistory.js
Andrew565: Addressed.

kryswisnaskas, line 56: Codex said: **[P2] TTA History pagination issue remains unresolved.**

Pagination reset was added to `TrainingReportDashboard`, which serves the Regional Dashboard. The reported problem is in `TrainingReportsSectionInContext` on Recipient TTA History, where filter changes still refetch using the stored page and offset. That component needs the reset behavior instead, or in addition.

src/routes/sessionReports/handlers.ts
```diff
@@ -377,6 +377,7 @@ export const getSessionReportsHandler = async (req: Request, res: Response) => {
       offset: offsetValue,
       limit: limitValue,
       format: formatValue as 'json' | 'csv',
+      userId,
```
kryswisnaskas: Codex said: **[P1] Request parameters can override the authenticated user ID.**

The fix adds authenticated `userId` before spreading `filteredFilterParams`. Because `userId` is not removed from `req.query`, a request such as `?userId=<another-user>&myReports.in[]=TR%20POC` overwrites the authenticated value. Put `userId` after the spread or explicitly remove it from accepted filter parameters.

kryswisnaskas (approved):
Looks good. Approving. Non-blocking: we might want to consider adding a focused TTA History regression test covering page 2 → apply filter → page 1/offset 0, since the current pagination test only covers Regional Dashboard.

AdamAdHocTeam (commented):
Looks good I will approve just wanted to note these more like nice to have improvements: 

Nice-to-haves (non-blocking): 

1) Consider guarding the participantCount ::integer casts in sessionReports.ts. The session form writes empty strings to these JSONB fields when deliveryMethod changes, and ''::integer throws "invalid input syntax for type integer" (COALESCE won't catch it since the cast errors first). A single malformed or legacy session in scope would fail the whole query and CSV export. The sibling [trSessionsForRecipient.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) already guards this exact case. Minimal fix: NULLIF(...->>'numberOfParticipants', '')::integer, and the same for the two hybrid fields. 

2) Consider deduping the reset-pagination useEffect + useSessionSort block shared by TrainingReportsSectionInContext (TTAHistory) and TrainingReportDashboard into a small hook to avoid drift.

AdamAdHocTeam (approved):

PR #3866, opened 2026-08-25
[TTAHUB-5736] load staging/prod GTM vars before frontend build

HaymondM:
## Description of change

Fixes staging and production Google Analytics tracking by making each environment's GTM build variables available before the frontend is built.

The deployed sites were not reliably including the Google Tag Manager script in the rendered HTML because the frontend build runs before Cloud.gov deploy-time environment variables are loaded. Vite therefore fell back to the committed frontend default where `REACT_APP_GTM_ENABLED=false`.

This change updates the CircleCI frontend build step for the normal `build_test_deploy` workflow so it:
- uses `deployment_config/staging_vars.yml` when building `main` for staging
- uses `deployment_config/prod_vars.yml` when building `production` for production
- filters to frontend-safe `REACT_APP_GTM*` and accessibility variables
- substitutes CircleCI environment values
- exports those values before `yarn --cwd frontend run build`

This enables the appropriate GTM configuration for staging and production deployment builds without changing feature-branch or local frontend defaults.

## How to test

1. Confirm CI passes for this branch.
2. Confirm `.circleci/config.yml` parses as valid YAML.
3. After merging to `main` and deploying to staging, view the staging page source and confirm the expected `googletagmanager.com` script appears.
4. After promoting the change to `production`, view the production page source at `https://ttahub.ohs.acf.hhs.gov` and confirm the expected GTM script appears.
5. Confirm Google Analytics receives the expected staging and production traffic/events.

Test in a dev environment
image
 

## Jira Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5736

### Every PR

- [x] Linked Jira issue
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -437,9 +437,10 @@ jobs:
             ./tools/parse-env-CLI.js "${frontend_env_config}" ./temp.vars
             grep -E '^REACT_APP_(GTM|INCLUDE_ACCESSIBILITY_CSS)' ./temp.vars > ./frontend-build.template.vars
             circleci env subst < ./frontend-build.template.vars > ./frontend-build.vars
-            set -a
-            . ./frontend-build.vars
-            set +a
+            while IFS= read -r line; do
+              [ -z "${line}" ] && continue
+              export "${line}"
+            done < ./frontend-build.vars
             rm -f ./temp.vars ./frontend-build.template.vars ./frontend-build.vars
           fi
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -420,8 +420,21 @@ jobs:
     - run:
         name: Build frontend
         command: |
-          if [ "${CIRCLE_BRANCH}" = "production" ] && [ "<< pipeline.parameters.action >>" = "build_test_deploy" ]; then
-            ./tools/parse-env-CLI.js ./deployment_config/prod_vars.yml ./temp.vars
+          frontend_env_config=""
+
+          if [ "<< pipeline.parameters.action >>" = "build_test_deploy" ]; then
+            case "${CIRCLE_BRANCH}" in
+              main)
+                frontend_env_config="./deployment_config/staging_vars.yml"
+                ;;
+              production)
+                frontend_env_config="./deployment_config/prod_vars.yml"
+                ;;
+            esac
+          fi
+
+          if [ -n "${frontend_env_config}" ]; then
+            ./tools/parse-env-CLI.js "${frontend_env_config}" ./temp.vars
             grep -E '^REACT_APP_(GTM|INCLUDE_ACCESSIBILITY_CSS)' ./temp.vars > ./frontend-build.template.vars
             circleci env subst < ./frontend-build.template.vars > ./frontend-build.vars
             set -a
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -189,7 +189,8 @@ workflows:
             only:
             - main
     jobs:
-    - build_and_lint
+    - build_and_lint:
+        load_deploy_frontend_env: false
     - deploy_job:
         name: reset (dev-blue)
         target_env: "dev-blue"
@@ -336,6 +337,10 @@ jobs:
   build_and_lint:
     executor: docker-executor
     resource_class: large
+    parameters:
+      load_deploy_frontend_env:
+        type: boolean
+        default: true
     steps:
     - checkout
     - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -419,7 +419,32 @@ jobs:
...
+            esac
+          fi
+
+          if [ -n "${frontend_env_config}" ]; then
+            ./tools/parse-env-CLI.js "${frontend_env_config}" ./temp.vars
+            grep -E '^REACT_APP_(GTM|INCLUDE_ACCESSIBILITY_CSS)' ./temp.vars > ./frontend-build.template.vars
+            circleci env subst < ./frontend-build.template.vars > ./frontend-build.vars
+            while IFS= read -r line; do
+              [ -z "${line}" ] && continue
+              export "${line}"
```
kryswisnaskas: Claude said: Quoted YAML values get literal quote characters baked into the exported env var (correctness bug)

`.circleci/config.yml:439` (the `export "${line}"` loop)

`parse-env-CLI.js` does a naive key:value split and does not strip surrounding quotes, so a YAML entry like:

`REACT_APP_INCLUDE_ACCESSIBILITY_CSS: "false"`

becomes the literal text `REACT_APP_INCLUDE_ACCESSIBILITY_CSS="false"` in `frontend-build.vars` (quote characters included as data). When that line is fed through `export "${line}"`, bash does not reinterpret the embedded quotes as syntax -- they become part of the value:

`$ line='FOO="bar"'; export "$line"; echo "$FOO" "bar"        # literal quotes preserved`

I verified this directly. Contrast with the project's existing `set_env_vars` command (`.circleci/config.yml:1731`), which writes `export ${line}` as text into `$BASH_ENV` and lets a real shell parse it via source -- that correctly strips the quotes (`FOO=true`, not `FOO="true"`), which I also verified.

Failure scenario: `vite.config.mjs` checks `env.REACT_APP_INCLUDE_ACCESSIBILITY_CSS === 'true'`. Today the yaml value is `"false"`, so the bug is silent (both `"false"` and `false` fail the `=== 'true'` check). But if anyone ever flips that flag to `"true"` in `staging_vars.yml`/`prod_vars.yml` to turn on the accessibility CSS, the built value will be the 6-character string `"true"` (quotes included), which still fails `=== 'true'` -- the feature will silently stay off with no error, and it'll be a confusing regression to debug later since the yaml clearly says `"true"`.

Fix: either strip quotes when writing `frontend-build.vars` (e.g. `sed -e 's/^"//' -e 's/"$//'` on the value, or fix `parse-env-CLI.js` to strip matching leading/trailing quotes), or reuse the proven pattern from `set_env_vars` -- write `export ${line}` lines to a script and source it (or `env -S`) instead of `export "${line}"` per line.

.circleci/config.yml
```diff
@@ -419,7 +419,32 @@ jobs:
           ./tools/run-yarn-audit.js;
     - run:
         name: Build frontend
-        command: yarn --cwd frontend run build
+        command: |
+          frontend_env_config=""
+
+          if [ "<< pipeline.parameters.action >>" = "build_test_deploy" ]; then
```
kryswisnaskas: Codex said: Manual staging deploys still miss the build-time GTM vars

In PR head `.circleci/config.yml:425`, the new export block only runs when `pipeline.parameters.action == "build_test_deploy"`. But the `deploy_only` workflow explicitly "will build the app and deploy to target env" and accepts `staging` at `.circleci/config.yml:146`.

Result: an API-triggered `deploy_only` to staging still builds the Vite bundle without staging GTM vars, and `deploy_job` cannot fix that later because the static `build/` artifact has already been created and persisted.

Suggested fix: make `build_and_lint` receive or derive the effective frontend target for `deploy_only`, then load `deployment_config/staging_vars.yml` before `yarn --cwd frontend run build`.

.circleci/config.yml
```diff
@@ -419,7 +419,32 @@ jobs:
...
+            case "${CIRCLE_BRANCH}" in
+              main)
+                frontend_env_config="./deployment_config/staging_vars.yml"
```
kryswisnaskas: Codex said: Main-branch scheduled reset builds can pick up staging GTM for dev deployments

The new selection is branch-based: `main` maps to `staging_vars.yml` at PR head `.circleci/config.yml:426-429`. The `reset_env_cron` workflow runs on `main`, builds once, then deploys that same artifact to `dev-blue`, `dev-green`, `dev-red`, `dev-pink`, and `dev-gold` at `.circleci/config.yml:182`.

Result: those dev reset deployments can receive a frontend artifact built with staging GTM settings.

Suggested fix: don't choose frontend build config from branch alone. Pass an explicit build target per workflow/job, or split per-environment builds where one workflow deploys the same frontend bundle to multiple environments.

kryswisnaskas (approved):

AdamAdHocTeam (approved):

PR #3867, opened 2026-08-25
[TTAHUB-5065] Add column for event report id

AdamAdHocTeam:
## Description of change

Because we rely on event id from the smart sheet import to be a unique identifier we want it to have its own db column.

## How to test

- Review the code
- Make sure we properly update legacy event columns.
- Importing a new event should populate the new column.
- All existing usages should pull the new value.

## Jira Issue(s)

* https://jira.acf.gov/browse/TTAHUB-5065

### Every PR

- [x] Linked Jira issue
- [ ] UI review complete
- [ ] QA review complete

### Before merge to main

- [ ] OHS demo complete

### Production Deploy

- [ ] PR created as **Draft**

### After merge/deploy


src/services/event.ts
```diff
@@ -376,12 +383,17 @@ export async function updateEvent(id: number, request: UpdateEventRequest): Prom
...
+  const { eventId } = data as { eventId?: string };
+
   await evt.update(
     {
       ownerId,
       pocIds,
       collaboratorIds,
       regionId,
+      ...(eventId ? { eventId } : {}),
       data: cast(JSON.stringify(data), 'jsonb'),
```
kryswisnaskas: Codex said: `updateEvent` can desync the canonical column from JSON

`src/services/event.ts` preserves the physical `eventId` column when `data.eventId` is omitted, but still overwrites the whole JSON `data` payload at line 397. Many live paths still read `event.data.eventId`, including alerts and session shaping.

Fix: make `updateEvent` always write a synced JSON value, e.g. merge `eventId: evt.eventId` into the stored `data`, and reject any request trying to change it to a different value.

src/models/eventReportPilot.js
```diff
@@ -40,6 +41,11 @@ export default (sequelize, DataTypes) => {
         type: DataTypes.INTEGER,
         allowNull: false,
       },
+      eventId: {
+        type: DataTypes.TEXT,
+        allowNull: false,
```
kryswisnaskas, line 45: Codex said: Several backend tests still create impossible EventReportPilot rows

The model now requires `eventId` here, but existing fixtures still create rows with no `data.eventId`, for example `src/scopes/trainingReports/startDate.test.js`, `src/scopes/trainingReports/endDate.test.js`, and an explicit missing-event fixture in `src/scopes/trainingReports/eventId.test.js`. These will fail validation under the new NOT NULL model constraint.

Fix: update the shared TR test fixture/helper to generate unique `data.eventId` values, and remove/update tests that expect missing event ids.

thewatermethod (approved):
Works! I don't love all the work we are doing here to keep the data object and eventId column in sync, I think a better approach would be to just fix all the call sites to refer to the actual column. If you feel that's out of scope, I think we should log a tech debt ticket to follow up and finish this work

src/models/hooks/eventReportPilot.js
```diff
@@ -44,6 +54,11 @@ const notifyNewOwner = async (_sequelize, instance) => {
   }
 };
 
+const beforeValidate = async (_sequelize, instance) => {
+  // Mirror before validation so the NOT NULL check on the column passes.
+  mirrorEventId(instance);
```
thewatermethod: Does it make sense to keep this mirrorEventId stuff in? It seems to exist only to persist data.eventId into the eventId column. It also doesn't validate the shape of the event ID, only that it's a string, so it might make sense to add a simple fn directly after "mirrorEventId" in the beforeValidate hook that throws an error if the shape of event ID doesn't match the expected "r01-PD-12312" or whatever it is supposed to bed

frontend/yarn-audit-known-issues
thewatermethod, line 1: we'll need to correct this on main rather than merge this in, I'll work on a PR to fix it

src/models/hooks/eventReportPilot.js
thewatermethod: actually, while I think adding that regex is probably smart, it will blow up a lot of tests I assume so maybe it would be tough to get done this sprint. I would still think about whether we need any hook here still

thewatermethod: And if we do need it, we should add an explicit test

thewatermethod:
```js
// src/models/hooks/eventReportPilot.test.js
import { beforeValidate } from './eventReportPilot';

describe('beforeValidate', () => {
  it('mirrors a string data.eventId onto the column', async () => {
    const instance = { data: { eventId: 'R01-TR-1234' } };
    await beforeValidate({}, instance);
    expect(instance.eventId).toBe('R01-TR-1234');
  });

  it('does not clobber when data is a sequelize literal / lacks eventId', async () => {
    const instance = { data: {}, eventId: 'R01-TR-9999' };
    await beforeValidate({}, instance);
    expect(instance.eventId).toBe('R01-TR-9999');
  });
});
```

thewatermethod:
Also worth noting that the import still populates data.eventId; we may want to change that to reduce confusion

frontend/yarn-audit-known-issues
AdamAdHocTeam: i agree but wanted to keep this PR inline with what it set out to do.

thewatermethod (approved):

